PR #65, opened 2022-08-05
Add running taipy-core as a service

toan-quach:


jrobinAV (commented):
We could benefit from testing the run method logic in a unit test. Even if the Core, Rest, and Gui services are mocked.

florian-vuillemot (approved):

Issue #66, opened 2022-08-09
Added unit tests for Taipy


toan-quach:
Currently taipy repo doesn't have any unit tests

Expectation:
- Set up a dev environment with pipenv in taipy repo
- Added unit tests for run function

PR #67, opened 2022-08-10
feature/#66 add pipenv env and unit tests

toan-quach:
#66

### DIFF
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Install Taipy without dependencies
         run: |
           pip install .
-          rm -r taipy
+          rm -rf src taipy
 
           python -c "import taipy as tp; tp.Scenario"
           python -c "import taipy as tp; tp.gui"
--- a/.github/workflows/packaging.yml
+++ b/.github/workflows/packaging.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Install Taipy without dependencies
         run: |
           pip install .
-          rm -rf src taipy
+          rm -rf src
 
           python -c "import taipy as tp; tp.Scenario"
           python -c "import taipy as tp; tp.gui"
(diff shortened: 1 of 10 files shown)


.github/workflows/packaging.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Install Taipy without dependencies
         run: |
           pip install .
-          rm -r taipy
+          rm -rf src taipy
```
FabienLelaquais: Do we still need 'taipy'' at this point?

.gitignore
```diff
@@ -1,3 +1,29 @@
 .idea
 .venv
 __pycache__
+
```
FabienLelaquais, line 4: I really don't think we need a tenth of these.

Pipfile
```diff
@@ -0,0 +1,17 @@
...
+
+[packages]
+taipy-gui = {git = "https://git@github.com/Avaiga/taipy-gui.git@develop"}
+taipy-rest = {git = "https://git@github.com/Avaiga/taipy-rest.git@develop"}
+
+[dev-packages]
+pytest = "*"
+
+[requires]
+python_version = "3.9"
```
FabienLelaquais: I though 3.8 would be the minimum...

FabienLelaquais (dismissed):
Minor things.
Looking good!

.github/workflows/packaging.yml
toan-quach: we dont!

Pipfile
toan-quach: true, I didn't pay much attention to it because I was trying to get it up and running quickly for testing. Will switch to 3.8 :D

.gitignore
toan-quach: I'll try and narrow it down to only the ones we need

jrobinAV (approved):

PR #73, opened 2022-10-18
Add Badges to repository Readme

joaoandre-avaiga:
This PR adds the following badges:

- Test status
- Supported Python versions
- Pypi package version

Coverage will be added in the future together with any other badge that it may be suggested.


jrobinAV (approved):
Good to me!

PR #85, opened 2023-02-02
fix typing for taipy.run

Dr-Irv:
The typing hints for the new `taipy.run()` were wrong.  It's not expecting a list.  Since the argument is `*apps`, you then specify the type of each element of `apps`.  So this updates the docs and the typing, and properly types the function `__typing_get()`

In addition, the pre-commit stuff didn't work.  Had to copy over some files from `taipy-gui` and also update the `pre-commit` file to get correct versions.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,2 @@
+[mypy]
+ignore_missing_imports = True
--- a/src/taipy/_run.py
+++ b/src/taipy/_run.py
@@ -18,13 +18,14 @@ from taipy.gui import Gui
 from taipy.rest import Rest
 from taipy.core import Core
 
-if sys.version_info < (3, 10):
-    from typing_extensions import TypeGuard
-else:
+if sys.version_info >= (3, 10):
     from typing import TypeGuard
 
+_AppType = t.Union[Gui, Rest, Core]
+_AppTypeT = t.TypeVar("_AppTypeT", Gui, Rest, Core)
 
-def _run(*apps: t.Union[Gui, Rest, Core], **kwargs) -> t.Optional[Flask]:
+
+def _run(*apps: t.List[_AppType], **kwargs) -> t.Optional[Flask]:
     """Run one or multiple Taipy services.
 
     A Taipy service is an instance of a class that runs code as a Web application.
@@ -36,9 +37,9 @@ def _run(*apps: t.Union[Gui, Rest, Core], **kwargs) -> t.Optional[Flask]:
         **kwargs: Other parameters to provide to the services.
     """
 
-    gui = __typing_get(apps, Gui)
-    rest = __typing_get(apps, Rest)
-    core = __typing_get(apps, Core)
+    gui = __get_app(apps, Gui)
+    rest = __get_app(apps, Rest)
+    core = __get_app(apps, Core)
 
     if gui and core:
         from taipy.core._version._version_cli import _VersioningCLI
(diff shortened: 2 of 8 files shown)


src/taipy/_run.py
```diff
@@ -49,8 +59,15 @@ def _run(*apps: t.List[t.Union[Gui, Rest, Core]], **kwargs) -> t.Optional[t.Unio
         return gui.run(**kwargs)
     else:
         app = rest or gui
+        assert app is not None  # for pyright typing
         return app.run(**kwargs)
 
 
-def __typing_get(l, type_):
-    return next(filter(lambda o: isinstance(o, type_), l), None)
+TObj = t.TypeVar("TObj", bound=t.Union[Gui, Core, Rest])
```
FabienLelaquais: This TObj must be hidden (using a _ prefix) so that the doc generation doesn't complain about an undocumented yet 'public' variable.
-> _t_obj

src/taipy/_run.py
```diff
@@ -9,22 +9,32 @@
...
 from taipy.core import Core
 
+if sys.version_info < (3, 10):
+    from typing_extensions import TypeGuard
+else:
+    from typing import TypeGuard
+
 
-def _run(*apps: t.List[t.Union[Gui, Rest, Core]], **kwargs) -> t.Optional[t.Union[Gui, Rest, Core]]:
+def _run(*apps: t.Union[Gui, Rest, Core], **kwargs) -> t.Optional[t.Union[Gui, Rest, Core, Flask]]:
```
FabienLelaquais: This signature was indeed pretty wrong. Thanks.
However, the return type should be `t.Optional[Flask]`.
There would be no use to return... well one of the parameters to `run()`.

src/taipy/_run.py
```diff
@@ -9,22 +9,32 @@
...
     """Run one or multiple Taipy services.
 
     A Taipy service is an instance of a class that runs code as a Web application.
 
     Parameters:
-        *args (List[Union[`Gui^`, `Rest^`, `Core^`]]): Services to run. If several services are provided, all the services run simultaneously. If this is empty or set to None, this method does nothing.
+        *args (Union[`Gui^`, `Rest^`, `Core^`]): Services to run, as separate arguments.
```
FabienLelaquais: Please add a &lt;/br&gt; before the end of the line. The doc will be easier to read.

mypy.ini
```diff
@@ -0,0 +1,2 @@
+[mypy]
```
FabienLelaquais, line 1: Why do we need this?
I'm not in favor of hiding missing imports, which is a problem we must spot.

FabienLelaquais (changes requested):
Hi Irv, thanks a lot for this useful input.
I've requested a few changes. But I certainly can do that on my side.
Please let me know if you want me to be in charge.

src/taipy/_run.py
```diff
@@ -49,8 +59,15 @@ def _run(*apps: t.List[t.Union[Gui, Rest, Core]], **kwargs) -> t.Optional[t.Unio
         return gui.run(**kwargs)
     else:
         app = rest or gui
+        assert app is not None  # for pyright typing
```
FabienLelaquais: And this would be due to a pyright bug/limitation, right?
No impact on the runtime, fortunately...

src/taipy/_run.py
Dr-Irv: fixed in next commit

src/taipy/_run.py
Dr-Irv: OK, fixed in next commit.

src/taipy/_run.py
Dr-Irv: Done in next commit.  Presume that you were looking for a line break not an HTML `</br>`

mypy.ini
Dr-Irv: I copied this from `taipy-gui/mypy.ini`.  I needed to do this because in order for the pre-commit to work, it runs `mypy`, and I had to copy over the contents of the other projects into the `taipy` project as follows:

-  `taipy-core/src/taipy/core` -> `taipy/src/taipy/core`
-  `taipy-gui/src/taipy/gui` -> `taipy/src/taipy/gui`
-  `taipy-rest/src/taipy/rest` -> `taipy/src/taipy/rest`

Without `mypy.ini` in there, then it complains about missing imports from other packages.

Dr-Irv:
> Hi Irv, thanks a lot for this useful input. I've requested a few changes. But I certainly can do that on my side. Please let me know if you want me to be in charge.

I addressed your comments in https://github.com/Avaiga/taipy/pull/85/commits/ba847f5029dc65b98a6ab62b09f077eaac96cde6

src/taipy/_run.py
FabienLelaquais: I really meant the HTML tag, because the Markdown parser will consume that line skip as it it were a white space.

Then all the doc appears in a single line, which is less readable. I will fix this ultimately. Thx

mypy.ini
FabienLelaquais: Too bad. But thanks.

jrobinAV (approved):

FabienLelaquais (approved):
Finally! 👌

src/taipy/_run.py
Dr-Irv: > And this would be due to a pyright bug/limitation, right? No impact on the runtime, fortunately...

It's more that when you have this logic:
```python
    if not rest and not gui:
        return None
    if gui and rest:
        gui._set_flask(rest._app)
        return gui.run(**kwargs)
    else:
        app = rest or gui
        assert app is not None  # Avoid pyright typing error
        return app.run(**kwargs)
```
when you get to the `else`, taipy can't infer that `rest` or `gui` must not be `None`.  Wouldn't say it's a bug.  It's actually a feature that `pyright` does quite well in general, but I reported the type narrowing issue here:  https://github.com/microsoft/pyright/issues/4643

`mypy` has the same issue, although I'm not sure why it didn't complain in your case, but if does have the same problem as what I put in the `pyright` issue.

Issue #93, opened 2023-02-15
Run GitHub action tests on Python 3.11


jrobinAV:
- taipy repository: Test GitHub action does not run the tests on Python 3.11
- taipy enterprise: https://github.com/Avaiga/taipy-enterprise/blob/develop/tools/publish.sh

PR #113, opened 2023-03-31
Feature/#93 - Run GitHub action tests on Python 3.11 and add linter workflow

trgiangdo:
https://github.com/Avaiga/taipy/issues/93

In this PR:
- Run GitHub action tests on Python 3.11
- Add linter workflow

### DIFF
--- a/src/taipy/_run.py
+++ b/src/taipy/_run.py
@@ -14,9 +14,9 @@ import typing as t
 
 from flask import Flask
 
+from taipy.core import Core
 from taipy.gui import Gui
 from taipy.rest import Rest
-from taipy.core import Core
 
 if sys.version_info >= (3, 10):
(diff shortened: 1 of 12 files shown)


src/taipy/_run.py
```diff
@@ -65,6 +65,8 @@ def _run(*apps: t.List[_AppType], **kwargs) -> t.Optional[Flask]:
         return app.run(**kwargs)
 
 
+# Avoid black adding too many empty lines
+# fmt: off
```
trgiangdo, line 69: This is to avoid conflict between black versions.

On python 3.10, black would only add 1 empty line after line 75.
But on Github action, black would add 2, which causes the conflict.

jrobinAV (dismissed):

jrobinAV (approved):

Issue #115, opened 2023-04-11
Feed the ScenarioSelector with actual data from the back-end


FabienLelaquais:
**What would that feature address**

So far the data is hard-coded (in the front-end).
We want some context variable to be holding the data we want to represent (the scenario tree).

***Description of the ideal solution***

One LOV would hold the list of items as tuples.

***Caveats***

This list has to be updated when the model is modified.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%

FredLL-Avaiga:
A first step would be to feed fake data from the Back-End

PR #123, opened 2023-04-20
#115 connect Gui to Core

FredLL-Avaiga:
This works with this demo 
Not finished (need to involve @JaufreyTpy )

```
import src.taipy

from taipy import Scope, Config, Frequency, Core, Scenario


def replace(input_msg, template, name):
    return input_msg.replace(template, name)

def on_action(state: State, id, action: str, scenario: Scenario):
    state.sc_id = scenario.id

sc_id = None

page = """
## Test Scenario Selector
<|toggle|theme|>

### Scenarios Selector
<|scenario_selector|>

Selected: <|{sc_id}|>
"""

# Gui.add_library(GuiCore())

input_msg_cfg = Config.configure_data_node(id="input_msg", scope=Scope.GLOBAL, default_data="Hello, today it's <TPL>!")
template_cfg = Config.configure_data_node(id="template", scope=Scope.GLOBAL, default_data="<TPL>")
date_cfg = Config.configure_data_node(id="date")
message_cfg = Config.configure_data_node(id="msg")

replace_template_cfg = Config.configure_task("replace", replace, [input_msg_cfg, template_cfg, date_cfg], message_cfg)
scenario_daily_cfg = Config.configure_scenario_from_tasks(
    id="scenario_daily_cfg", task_configs=[replace_template_cfg], frequency=Frequency.DAILY
)
scenario_cfg = Config.configure_scenario_from_tasks(id="scenario_cfg", task_configs=[replace_template_cfg])

Core().run()

Gui(page).run(debug=True)


```


src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -11,50 +11,108 @@
...
 
 
 class GuiCore(ElementLibrary):
+    __LIB_NAME = "taipy_gui_core"
+    __CTX_VAR_NAME = f"__{__LIB_NAME}_Ctx"
+
     __elts = {
         "scenario_selector": Element(
-            "scenario",
+            "scenario_id",
```
FabienLelaquais, line 143: Why use an id here? Scenario is a first-citizen class is it not?

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -11,50 +11,108 @@
...
+    def __init__(self, gui: Gui) -> None:
         self.gui = gui
-        self.core = core
-        self.scenarios: t.List[Scenario]
+        self.scenarios: t.Optional[
+            t.List[t.Tuple[str, str, int, bool, t.Optional[t.List[t.Tuple[str, str, int, bool, None]]]]]
+        ] = None
+        self.scenario_configs: t.Optional[t.List[t.Tuple[str, str]]] = None
+
+    def __add_scenarios(self, res, scenarios):
```
FabienLelaquais: I suspect that this could be defined within the get_scenarios() method...

FabienLelaquais (dismissed):
Looks really neat!

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: It could but why?

FredLL-Avaiga: It could also be static

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: It is but I thought that core users prefer to use IDs.
And I didn't want to bring adapters in this story...

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -11,50 +11,109 @@
...
+            date = parser.parse(date_str) if isinstance(date_str, str) else None
+        except Exception as e:
+            state.assign(GuiCoreContext._ERROR_VAR, f"Invalid date ({date_str}).{e}")
+            return
+        scenario = tp.create_scenario(scenario_config, date, data.get(GuiCoreContext.__PROP_SCENARIO_NAME))
+        if props := data.get("properties"):
+            for prop in props:
+                key = prop.get("key")
+                if key and key not in GuiCoreContext.__SCENARIO_PROPS:
+                    scenario._properties[key] = prop.get("value")
```
jrobinAV: That s an optimization. When updating an attribute or property of the scenario, the scenario is automatically saved. Using the `with scenario` context, the save is only performed at the exit.

```suggestion
        if props := data.get("properties"):
            with scenario as sc:
                for prop in props:
                    key = prop.get("key")
                    if key and key not in GuiCoreContext.__SCENARIO_PROPS:
                        sc._properties[key] = prop.get("value")
```

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -11,50 +11,109 @@
...
 
-from taipy.core import Core, Scenario
+from dateutil import parser
+
+import taipy as tp
 from taipy.gui import Gui, State
 from taipy.gui.extension import Element, ElementLibrary, ElementProperty, PropertyType
 
 
 class GuiCoreContext:
```
jrobinAV: Rename to CoreElementLibrary? CoreComponentLibrary?

jrobinAV (dismissed):

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: tx

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: I'll let you decide.
But as it is now automatically registered ...

src/taipy/gui_core/GuiCoreLib.py
FabienLelaquais: The why is for easier-to-read code and scope concerns.
A function used only by another one has all the reasons to be located in the caller's scope.

src/taipy/gui_core/GuiCoreLib.py
FabienLelaquais: "Component" is reserved for the front-end (React).
"Block" is for structuring elements, "Control" for the others.
The proper naming for the class would therefore be "CoreElementLibrary".

The "Gui" part of all the public items is certainly useless.
I would rename GuiCoreContext to CoreContext. Its access is protected by the modules.

jrobinAV (approved):

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -10,51 +10,157 @@
...
+            state.assign(GuiCoreContext._ERROR_VAR, f"Invalid date ({date_str}).{e}")
+            return
+        scenario = tp.create_scenario(scenario_config, date, data.get(GuiCoreContext.__PROP_SCENARIO_NAME))
+        if props := data.get("properties"):
+            with scenario as sc:
+                for prop in props:
+                    key = prop.get("key")
+                    if key and key not in GuiCoreContext.__SCENARIO_PROPS:
+                        sc._properties[key] = prop.get("value")
+        state.assign(GuiCoreContext._ERROR_VAR, "")
```
FabienLelaquais, line 97: Why that?

ghost:
@FredLL-Avaiga How can I help, here?

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -10,51 +10,157 @@
...
-    _VAR_NAME = "__CCCtx"
+    _ERROR_VAR = "gui_core_error"
 
-    def __init__(self, gui: Gui, core: Core) -> None:
+    def __init__(self, gui: Gui) -> None:
         self.gui = gui
-        self.core = core
-        self.scenarios: t.List[Scenario]
+        self.scenarios: t.Optional[
+            t.List[t.Tuple[str, str, int, bool, t.Optional[t.List[t.Tuple[str, str, int, bool, None]]]]]
```
FabienLelaquais, line 37: Whao. Nice type.

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -10,51 +10,157 @@
...
+    def get_scenario_configs(self):
+        if self.scenario_configs is None:
+            configs = tp.Config.scenarios
+            if isinstance(configs, dict):
+                self.scenario_configs = [(id, f"{c.id}") for id, c in configs.items()]
+        return self.scenario_configs
 
     def create_new_scenario(self, state: State, id: str, action: str, payload: t.Dict[str, str]):
-        pass
+        print(f"create_new_scenario(state, {id}, {action}, {payload}")
```
FabienLelaquais, line 74: I guess that should be removed.

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -10,51 +10,157 @@
...
+        if props := data.get("properties"):
+            with scenario as sc:
+                for prop in props:
+                    key = prop.get("key")
+                    if key and key not in GuiCoreContext.__SCENARIO_PROPS:
+                        sc._properties[key] = prop.get("value")
+        state.assign(GuiCoreContext._ERROR_VAR, "")
+
+    def select_scenario(self, state: State, id: str, action: str, payload: t.Dict[str, str]):
+        print(f"select_scenario(state, {id}, {action}, {payload}")
```
FabienLelaquais, line 100: I guess that should be removed.

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -10,51 +10,157 @@
...
+        args = payload.get("args")
+        if args is None or not isinstance(args, list) or len(args) == 0 or not isinstance(args[0], dict):
+            return
+        data = args[0]
+        action = data.get("user_action")
```
FabienLelaquais, line 105: So that will ultimately be a 'change', right?

FabienLelaquais (approved):
That looks excellent!

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: I've added an internal property error in the component ...
It was more a proof of concept than anything: it shows that we can create a variable in the user state.
The error reporting could also be done via notify ...

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: indeed it will in the end

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: yes, next iteration

Issue #119, opened 2023-04-17
[Core-GUI] Integration Core and GUI for scenario selector 


jrobinAV:
works with taipy-core 2.3.0.dev0 or later

## Basic core application
```
from datetime import datetime
from datetime import timedelta

import taipy as tp
from taipy import Scope, Config, Frequency, Core


def replace(input_msg, template, name):
    return input_msg.replace(template, name)


def test_core_gui():
    input_msg_cfg = Config.configure_data_node(id="input_msg", scope=Scope.GLOBAL, default_data="Hello, today it's <TPL>!")
    template_cfg = Config.configure_data_node(id="template", scope=Scope.GLOBAL, default_data="<TPL>")
    date_cfg = Config.configure_data_node(id="date")
    message_cfg = Config.configure_data_node(id="msg")

    replace_template_cfg = Config.configure_task("replace", replace, [input_msg_cfg, template_cfg, date_cfg], message_cfg)
    scenario_daily_cfg = Config.configure_scenario_from_tasks(id="scenario_daily_cfg",
                                                              task_configs=[replace_template_cfg],
                                                              frequency=Frequency.DAILY)
    scenario_cfg = Config.configure_scenario_from_tasks(id="scenario_cfg", task_configs=[replace_template_cfg])

    Core().run()
```

## Scenario creations
```
    sc_1 = tp.create_scenario(scenario_daily_cfg, creation_date=datetime(2023, 1, 2))
    sc_2 = tp.create_scenario(scenario_daily_cfg, creation_date=sc_1.date + timedelta(days=1))
    sc_2._properties["label"] = "THIS IS A SPECIAL LABEL"

    sc_2_bis = tp.create_scenario(scenario_daily_cfg,  creation_date=sc_1.date + timedelta(days=1))
    sc_3 = tp.create_scenario(scenario_daily_cfg, creation date=sc_1.date + timedelta(days=2))
    sc_none_1 = tp.create_scenario(scenario_cfg, creation_date=datetime(2023, 1, 10))
    sc_none_2 = tp.create_scenario(scenario_cfg, creation_date=datetime(2023, 1, 11))
```

## Retrieve scenario tree
```
tp.get_cycles_scenarios()
```

## Code that updates a scenario label:
```
import random

def random_scenario_label_update(scenario):
    scenario.name = ''.join(random.choice(string.ascii_lowercase) for i in range(8))
```

## Consumer example
```
import taipy as tp
[...]

PR #131, opened 2023-04-28
#119 aligning with taipy-gui

FredLL-Avaiga:


FabienLelaquais (approved):
The properties are far cleaner now. Thx!

PR #134, opened 2023-05-03
update install script (#115)

FredLL-Avaiga:
prepare scenario_visualizer backend

### DIFF
--- a/src/taipy/gui_core/GuiCoreLib.py
+++ b/src/taipy/gui_core/GuiCoreLib.py
@@ -52,10 +52,14 @@ class GuiCoreContext(CoreEventConsumerBase):
     __PROP_SCENARIO_CONFIG_ID = "config"
     __PROP_SCENARIO_DATE = "date"
     __PROP_SCENARIO_NAME = "name"
+    __PROP_SCENARIO_PRIMARY = "primary"
+    __PROP_SCENARIO_TAGS = "tags"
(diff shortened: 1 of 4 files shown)


src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -212,7 +253,15 @@ def on_init(self, gui: Gui) -> t.Optional[t.Tuple[str, t.Any]]:
...
-        state._add_attribute(GuiCoreContext._ERROR_VAR)
-        state._gui._bind_var_val(GuiCoreContext._ERROR_VAR, "")
-        state._add_attribute(GuiCoreContext._SCENARIO_SELECTOR_ID_VAR)
-        state._gui._bind_var_val(GuiCoreContext._SCENARIO_SELECTOR_ID_VAR, "")
+        for var in [
+            GuiCoreContext._SCENARIO_SELECTOR_ERROR_VAR,
+            GuiCoreContext._SCENARIO_SELECTOR_ID_VAR,
+            GuiCoreContext._SCENARIO_VIZ_ERROR_VAR,
+        ]:
+            state._add_attribute(var)
```
FabienLelaquais, line 261: Does it make sense to add_attribute() but not bind_var_val()?
In which case a state._add_attribute(gui, var) would certainly be simpler...

FabienLelaquais (approved):
Looks kind of familiar.

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: I've updated _add_attribute on the current State PR (feature/#737-state.refresh) 
I'll modify this once that PR is merged

PR #137, opened 2023-05-09
art: README changes, corrects aspect ratios for images, corrects code for demos

AlexandreSajus:
Previously, images on the README had their width and height defined which would change their original aspect ratio.
The values were defined in pixels which meant they would change size according to one's monitor resolution.

Here we define image dimensions by only using `width=XX%` which both conserves the image's original aspect ratio and makes the image have the same size regardless of the screen used.

Before:
image

After:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@
     
 | TAIPY GUI - the frond-end  | TAIPY Core - the back-end |
 | --------  | -------- |
-|<img src="readme_img/readme_gui_intro.gif" alt="Taipy GUI Gif"  width="100%"/> | <img src="readme_img/readme_core_intro.gif" alt="Taipy Core Gif"  width="100%"/>
+|<img src="readme_img/readme_gui_intro.gif" alt="Taipy GUI Animation"  width="100%"/> | <img src="readme_img/readme_core_intro.gif" alt="Taipy Core Animation"  width="100%"/>
 
     
 </div> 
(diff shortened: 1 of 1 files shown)


AlexandreSajus:
Added a commit to make the first example in the README runnable (taipy.Gui was not imported and the page had the wrong name)

Before:
image

After:
image

jrobinAV (dismissed):
Good catch

marisogo (dismissed):

README.md
```diff
@@ -40,7 +40,7 @@
     
 | TAIPY GUI - the frond-end  | TAIPY Core - the back-end |
 | --------  | -------- |
-|<img src="readme_img/readme_gui_intro.gif" alt="Taipy Logo"  width="350" height="300"/> | <img src="readme_img/readme_core_intro.gif" alt="Taipy Logo"  width="350" height="300"/>
+|<img src="readme_img/readme_gui_intro.gif" alt="Taipy GUI Gif"  width="100%"/> | <img src="readme_img/readme_core_intro.gif" alt="Taipy Core Gif"  width="100%"/>
```
FabienLelaquais: You should not specify the file type in the alt entry. That should be a description of the content, not how it is done.
=> "Animated image for introducing Taipy Core "

FabienLelaquais (dismissed):
Looks good!

README.md
AlexandreSajus: Fixed!

FabienLelaquais (approved):
👍

Issue #148, opened 2023-05-22
Scenario Graph: improve


FredLL-Avaiga:
[] rename graph to dag
[] add property to show dag dynamically and/or in the page

PR #149, opened 2023-05-22
#148 improve dag

FredLL-Avaiga:
- Rename graph to dag
- changed scenario implementation from dict to array
- properties  added:
  - show (dynamic bool)
  - with_button (default true)
  - width & height (if not defined and with button => dialog is full screen else default to 50vw & 50vh)


pavle-avaiga (approved):

FabienLelaquais (approved):
That looks really great! Well done!

Issue #150, opened 2023-05-22
Typing for `taipy.run()` still wrong


Dr-Irv:
In #85 I did a PR for fixing the typing of `taipy.run()`.  

In the commit https://github.com/Avaiga/taipy/pull/85/commits/d2360ef9f05414a92af3eb4935529419a832c598#diff-16c072932d406277262b9fbaae2f8a3e77493748417cc1714e17c61c0036b256

@FabienLelaquais changed
```python
def _run(*apps: t.Union[Gui, Rest, Core], **kwargs) -> t.Optional[Flask]:
```
to
```python
def _run(*apps: t.List[_AppType], **kwargs) -> t.Optional[Flask]:
```
But this change is incorrect.  Since `apps` is declared as `*apps`, the declaration should be `_AppType` to indicate that each item of `apps` is of type `_AppType`.  

I could do a PR, but would be easiest for you guys to just fix this.

FabienLelaquais:
Irv, you are absolutely right!
The next version will ultimately fix that.

Thanks a lot!

PR #153, opened 2023-05-23
Fix taipy._run() signature (#150)

FabienLelaquais:
Addresses #150.


jrobinAV (approved):

jrobinAV (changes requested):
lines 40, 41, and 42 apps should be renamed to services.

jrobinAV (approved):
Neat !

Issue #157, opened 2023-05-24
[Core-GUI] Create the Data Node Selector control


FabienLelaquais:
**Description**

Create a control that holds the data node entities from Core.

See overall features in [GUI controls for Core](https://docs.google.com/document/d/19R0SiykrlGALWU9ONDdtmni4PNBmJ-95s9i755FhWgg).

**Acceptance Criteria**
- [ ] DataNodes are listed, selectable, and updated on Core messages.
- [ ] Documentation is updated.

PR #179, opened 2023-06-09
#157 fix: node selector build issue

pavle-avaiga:


FredLL-Avaiga (commented):
Tx @pavle-avaiga but I've dealt with that already in my current PR
Can you drop this PR or shall I?

pavle-avaiga:
I will cancel this PR :)
Thanks @FredLL-Avaiga

PR #180, opened 2023-06-12
Update Dev Release Process

dinhlongviolin1:


.github/workflows/release-dev.yml
```diff
@@ -33,7 +65,7 @@ jobs:
         run: |
               echo """
               import json, os
-              with open(f\"src{os.sep}taipy{os.sep}version.json\") as version_file:
+              with open(f\"src{os.sep}taipy{os.sep}gui{os.sep}version.json\") as version_file:
```
FredLL-Avaiga: Copy paste?
I think Version.json is in Taipy

.github/workflows/release-dev.yml
```diff
@@ -79,9 +111,57 @@ jobs:
...
+        run: |
+              git reset --hard HEAD
+              git clean -fdx
+
+      - name: Increase dev version
+        id: new-version
+        run: |
+              echo """
+              import json, os
+              with open(f'src{os.sep}taipy{os.sep}version.json') as version_file:
```
FredLL-Avaiga, line 146: Idem

.github/workflows/publish-dev.yml
```diff
@@ -0,0 +1,159 @@
...
+              raise ValueError(f'Tag {sys.argv[1]} is not the latest dev version of branch {sys.argv[2]}')
+          """ > /tmp/check.py
+          python /tmp/check.py "${{ steps.vars.outputs.tag }}" "${{ steps.extract_branch.outputs.branch }}"
+
+      - name: Ensure package version has 'dev' suffix
+        run: |
+          echo """
+          import json, sys, os
+          SUFFIX = 'dev'
+          with open(f\"src{os.sep}taipy{os.sep}version.json\") as version_file:
```
FabienLelaquais, line 65: Do we really need these os.sep? I guess that open() supports '/' even on Windows...

.github/workflows/publish-dev.yml
```diff
@@ -0,0 +1,159 @@
...
+
+      - name: Download assets from github release tag
+        run: |
+          gh release download ${{ github.event.inputs.version }} --dir dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Verify there is a release asset
+        run: |
+          if [ ! -f dist/taipy-${{ github.event.inputs.version }}.tar.gz ]; then
```
FabienLelaquais, line 99: Not "dist/taipy." instead of "dist/taipy-"?
To me this '-' introduces another repo

FredLL-Avaiga (dismissed):

jrobinAV (dismissed):

FabienLelaquais (dismissed):
👍

.github/workflows/publish-dev.yml
dinhlongviolin1: pretty sure it is dist/taipy-

.github/workflows/publish-dev.yml
dinhlongviolin1: we have been using os.sep on other actions for a while. no need to change that. and os.sep ensures platform compatibility so it is always good to use it

PR #195, opened 2023-06-13
add title to dev releases

jrobinAV:
It s better to specify the title otherwise the latest commit is used as release title.


FabienLelaquais (approved):
👍

Issue #197, opened 2023-06-14
Gui: DAG: toolbar visible when no scenario


FredLL-Avaiga:
Hide toolbar when no scenario
Introduce the Core workaround for reading entities

PR #198, opened 2023-06-14
#197 dag toolbar core workaround

FredLL-Avaiga:
hide dag toolbar when no scenario
set read_entity_retry = 3 into config


FabienLelaquais (approved):
😑

PR #220, opened 2023-06-20
Hotfix2.3/Core635 - Wrong _CoreCLI import

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/635

### DIFF
--- a/src/taipy/_run.py
+++ b/src/taipy/_run.py
@@ -42,7 +42,7 @@ def _run(*services: _AppType, **kwargs) -> t.Optional[Flask]:
     core = __get_app(services, Core)
 
     if gui and core:
-        from taipy.core._version._cli._core_cli import _CoreCLI
+        from taipy.core._core_cli import _CoreCLI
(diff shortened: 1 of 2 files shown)


jrobinAV:
What about the unit test to avoid future regressions?

```
@mock.patch("taipy.gui.Gui.run")
@mock.patch("taipy.core.Core.run")
def test_run_pass_with_gui_and_core(core_run, gui_run):
    _run(Gui(), Core())
    gui_run.assert_called_once()
    core_run.assert_called_once()
```

jrobinAV (dismissed):

trgiangdo:
I thought that we don't need to add the unit test to the release branch. If needed, I will

jrobinAV:
> I thought that we don't need to add the unit test to the release branch. If needed, I will

I would add it just in case we are modifying the _run method again in a hotfix. There is little odds though.

jrobinAV (approved):

PR #228, opened 2023-06-23
use is_promotable and is_submittable

FredLL-Avaiga:


FabienLelaquais (dismissed):
👌

jrobinAV (approved):
LGTM I did not get all details of the tsx, but the python makes a lot of sense.

PR #231, opened 2023-06-23
Improve performances

### LINKED ISSUE #215: Improve performances on scenario and data selectors
- implement tp.exists(entity) on core
-

FredLL-Avaiga:
Resolves #215.

Manage scenario expandable/default_expanded

### DIFF
--- a/src/taipy/gui_core/viselements.json
+++ b/src/taipy/gui_core/viselements.json
@@ -85,10 +85,10 @@
             "doc": "If True, the scenario visualizer can be expanded.<br/>If False, the scenario visualizer is always collapsed and only its name and the <i>submit</i> button are visible."
           },
           {
-            "name": "expanded",
-            "type": "dynamic(bool)",
+            "name": "default_expanded",
+            "type": "bool",
             "default_value": "True",
-            "doc": "If True, the scenario visualizer is expanded, and its content is displayed.<br/>If False, the scenario visualizer is collapsed and only its name a <i>submit</i> button are visible."
+            "doc": "If True, when a valid scenario is selected, the scenario visualizer is expanded, and its content is displayed.<br/>If False, the scenario visualizer is collapsed and only its name a <i>submit</i> button are visible."
           },
           {
             "name": "show_submit",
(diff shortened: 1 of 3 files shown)


FredLL-Avaiga:
Is this worth a backport to 2.3 ?

jrobinAV (dismissed):
Looks good to me, but a few unit tests would have help understanding if it works well and if we don't miss some cases.

Thx for the expanded as well...

src/taipy/gui_core/viselements.json
```diff
@@ -85,10 +85,10 @@
             "doc": "If True, the scenario visualizer can be expanded.<br/>If False, the scenario visualizer is always collapsed and only its name and the <i>submit</i> button are visible."
           },
           {
-            "name": "expanded",
-            "type": "dynamic(bool)",
+            "name": "default_expanded",
```
FabienLelaquais: Why the property renaming? That creates what I think is a useless incompatibility...

FabienLelaquais (changes requested):
Can we discuss this?
I don't really agree...

FredLL-Avaiga:
We can :-)
I think the new name reflects the real use of the property.
And I think it was wrong to make it dynamic.
This property reflects the default state of the expand
That can then be changed by the user unless expandable is set to false

src/taipy/gui_core/viselements.json
FredLL-Avaiga: back to expanded

jrobinAV (approved):

FabienLelaquais:
> Looks good to me, but a few unit tests would have help understanding if it works well and if we don't miss some cases.
> 
> Thx for the expanded as well...

Fred an I have settled on a simplified set of properties:
- expandable (bool): whether or not the user can manually expand the viewer.
- expanded (bool): whether or not the viewer is expanded.

None of these properties are dynamic (that is, there is no programmatic way to set them).

- If expandable == False and expanded == False, the viewer will never be expanded.
- If expandable == False and expanded == True, the viewer will be expanded when a scenario is set, and only then.
- If expandable == True and expanded == False, the viewer will be collapsed, but the user can expand it, if a scenario is set.
- If expandable == True and expanded == True, the viewer will be expanded if a scenario is set and the user can collapse/expand it at will.

Would you agree that this covers all use cases?
Thx

FabienLelaquais (approved):
👌

PR #235, opened 2023-07-05
Feature/templates#14 - Auto-discover available templates

trgiangdo:
https://github.com/Avaiga/taipy-templates/issues/14

### DIFF
--- a/src/taipy/_cli/_scaffold_cli.py
+++ b/src/taipy/_cli/_scaffold_cli.py
@@ -14,13 +14,14 @@ import sys
 
 from cookiecutter.main import cookiecutter
 
+import taipy
 from taipy._cli._base_cli import _CLI
 from taipy.logger._taipy_logger import _TaipyLogger
 
 
 class _ScaffoldCLI:
     __logger = _TaipyLogger._get_logger()
-    __TAIPY_PATH = pathlib.Path(__file__).parent.parent.resolve() / "templates"
+    __TAIPY_PATH = pathlib.Path(taipy.__file__).parent.resolve() / "templates"
 
     _TEMPLATE_MAP = {str(x.name): str(x) for x in __TAIPY_PATH.iterdir() if x.is_dir()}
 
--- a/src/taipy/_cli/_scaffold_cli.py
+++ b/src/taipy/_cli/_scaffold_cli.py
@@ -39,5 +39,10 @@ class _ScaffoldCLI:
 
         if getattr(args, "which", None) == "create":
             application_path = cookiecutter(cls._TEMPLATE_MAP[args.template])
-            print(f"New Taipy application is created at {application_path}")
+            print(
+                f"New Taipy application has been created at {application_path}"
+                f"\n\nTo start the application, replace `main.py` with your"
+                f" application main Python file and run the following command:"
+                f"\n\tcd {application_path} && python main.py"
+            )
             sys.exit(0)
--- a/src/taipy/_cli/_scaffold_cli.py
+++ b/src/taipy/_cli/_scaffold_cli.py
@@ -16,11 +16,9 @@ from cookiecutter.main import cookiecutter
 
 import taipy
 from taipy._cli._base_cli import _CLI
-from taipy.logger._taipy_logger import _TaipyLogger
 
 
 class _ScaffoldCLI:
-    __logger = _TaipyLogger._get_logger()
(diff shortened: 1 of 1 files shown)


src/taipy/_cli/_scaffold_cli.py
```diff
@@ -15,23 +15,22 @@
...
 
     @classmethod
     def create_parser(cls):
         create_parser = _CLI._add_subparser("create", help="Create a new Taipy application.")
         create_parser.add_argument(
             "--template",
-            choices=["default", "multi-page-gui"],
-            default="default",
+            choices=list(cls._TEMPLATE_MAP.keys()),
+            default="taipy-default-template",
```
trgiangdo, line 32: The default choice is "taipy-default-template" for now.

We can rename the template in taipy-templates latter.

src/taipy/_cli/_scaffold_cli.py
```diff
@@ -40,5 +40,6 @@ def parse_arguments(cls):
         args = _CLI._parse()
 
         if getattr(args, "which", None) == "create":
-            cookiecutter(cls._TEMPLATE_MAP[args.template])
+            application_path = cookiecutter(cls._TEMPLATE_MAP[args.template])
+            cls.__logger.info(f"Taipy application created at {application_path}")
```
jrobinAV: I would not use the taipy logger here since it is designed for runtime when a Taipy application is running. 

Here we are at development time. It is configured to display the level, the timestamp, and maybe morethings that are useless for a command line.
Probably a print is enough.

What do you think?

Though it s not blocking to me.

src/taipy/_cli/_scaffold_cli.py
jrobinAV: I guess you meant to rename it to 'default'. Yes. In any case the default template will probably be changed in the next weeks to generate an empty application asking the user more questions like:
- How many pages and what are the page names?
- Do you want to run a Core service?
- a Rest service?

jrobinAV (dismissed):

src/taipy/_cli/_scaffold_cli.py
trgiangdo: In my defense, Taipy logger looks better 🤣

jrobinAV: 🤣

trgiangdo: Let's wait for @FabienLelaquais's review. I don't have strong opinion about this.

jrobinAV: Just to complete my comment: 
Logger (and so Taipy logger) is made for logs (big information 🤣 ). It is configurable to display more or less information on logs, so it is easier to automate a post-process, a search, a filter, or whatever treatment in the log base.

An output command line is not a log (or not in the same sense at least).
You don't have post-treatments on command line outputs. And usually, we don't want to make the command line output configurable. Making the command line output look like a log makes it strange to me.

Anyway, I won't fight on it. It's up to you.

FabienLelaquais: I do understand @jrobinAV's reluctance to use the internal Taipy logger for this "application". That indeed was not the purpose of this class.
And that would have been my first recommendation as well.

Now, I can see that no matter what, the 'taipy' script and the 'create' command, in particular, do depend on the Taipy library anyway (contrary to "light" scripts that would be more or less independent of deeply internal features of libraries they depend on).
So I'm good with using the Taipy logger. If @trgiangdo thinks the output is "better", and also more in line with other messages that Taipy would produce, in other contexts, let's stick with it.

src/taipy/_cli/_scaffold_cli.py
```diff
@@ -40,5 +38,6 @@ def parse_arguments(cls):
         args = _CLI._parse()
 
         if getattr(args, "which", None) == "create":
-            cookiecutter(cls._TEMPLATE_MAP[args.template])
+            application_path = cookiecutter(cls._TEMPLATE_MAP[args.template])
+            print(f"New Taipy application is created at {application_path}")
```
FabienLelaquais: "is" -> "has been"

May be a line on what to do next would be even better (like run the app).

FabienLelaquais (dismissed):
👍

jrobinAV (dismissed):

jrobinAV (approved):

PR #239, opened 2023-07-06
Dags and pins in selector

### LINKED ISSUE #236: Gui Core: DAG allow multiple dags in the same page

### LINKED ISSUE #238: Gui Core Selectors: Add pinning items
As discussed we want to be able to limit the number of visible elements in the selectors:
We can achieve this by adding a pin icon to every item in the tree and a "show pinned" toggle on top of the tree.
- When the toggle is on, only the pinned items are shown.
- Toggle is disabled when no pinned item
- The non leaf items can also be pinned, and this would select all the leaf of that branch.

FredLL-Avaiga:
Resolves #236 (multiple dags, paper around dag, size management).
Resolves #238 (pins in selector).

### DIFF
--- a/gui/src/CoreSelector.tsx
+++ b/gui/src/CoreSelector.tsx
@@ -1,0 +1,415 @@
+/*
+ * Copyright 2023 Avaiga Private Limited
+ *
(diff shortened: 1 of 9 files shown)


FabienLelaquais (approved):
Loved it!

PR #240, opened 2023-07-07
#825 avoid parasite updates

FredLL-Avaiga:
Resolves [taipy-gui#825](https://github.com/Avaiga/taipy-gui/issues/825)


FabienLelaquais (approved):
👍

PR #241, opened 2023-07-07
Scenario viewer adjustments

### LINKED ISSUE #230: Feedback on scenario core viz elmt
- [ ] Title (the label) should be bigger (from @vhfgosselin)
- [x] Should be expanded by default
- [x] Sub title `custom properties` is unnecessary
- [x] Tags should be True by default
- [ ] tag edition not working?
- [x] cycle should be displayed with `scenario.cycle.get_simple_label()`
- [x] expanded does not work
- [x] expandable does not work

FredLL-Avaiga:
Resolves #230.

-  Should be expanded by default
-  Sub title custom properties is unnecessary
-  Tags should be True by default
-  cycle should be displayed with scenario.cycle.get_simple_label()
-  expanded does not work
-  expandable does not work

### DIFF
--- a/gui/src/ScenarioViewer.tsx
+++ b/gui/src/ScenarioViewer.tsx
@@ -53,7 +53,6 @@ interface ScenarioViewerProps {
     id?: string;
     expandable?: boolean;
     expanded?: boolean;
-    defaultExpanded?: boolean;
     updateVarName?: string;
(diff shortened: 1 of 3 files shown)


FabienLelaquais (approved):
👌
Doc will get updated... when something happens there.

Issue #233, opened 2023-07-04
Scenario Selector - Preselect scenario configuration


FlorianJacta:
### Feature improvement

When adding a scenario through the Scenario Selector: The goal is to preselect the scenario configuration when there is only one scenario configuration in the application.

### Version
Taipy: 2.3.1

FredLL-Avaiga:
I wouldn't think this is worth a backport ...
@FabienLelaquais @jrobinAV ?

PR #244, opened 2023-07-17
Preselect scenario config if it is unique

FredLL-Avaiga:
Addresses #233.


FabienLelaquais (approved):
Makes a lot of sense to me...

PR #245, opened 2023-07-18
fix: read_entity_retry is set in configure_core()

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/666

### DIFF
--- a/src/taipy/gui_core/GuiCoreLib.py
+++ b/src/taipy/gui_core/GuiCoreLib.py
@@ -54,8 +54,6 @@ Pipeline.__bases__ += (_GCDoNotUpdate,)
 DataNode.__bases__ += (_GCDoNotUpdate,)
 Job.__bases__ += (_GCDoNotUpdate,)
 
-Config.configure_core(read_entity_retry=3)
-
 
 class _EntityType(Enum):
     CYCLE = 0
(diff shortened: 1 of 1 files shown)


src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -54,7 +54,7 @@ def __repr__(self):
 DataNode.__bases__ += (_GCDoNotUpdate,)
 Job.__bases__ += (_GCDoNotUpdate,)
 
-Config.configure_global_app(read_entity_retry=3)
+Config.configure_core(read_entity_retry=3)
```
jrobinAV: I believe this should be in taipy-core the default CoreSection value.
GuiCoreLib should not have to care about taipy core configuration. What do you think?

trgiangdo: The current default value of `read_entity_retry` for now is None 🤣

Since this is a special case only use in GuiCoreLib, I think it can be here.
But if the user always run the GuiCoreLib, I'm fine with moving it to Core.

jrobinAV: I would set the default value to 3 in taipy-core and remove this. It should have no impact on the performance. Plus that could cause the problem even if we don't use CoreGui visual elements.

trgiangdo: Sure. I'm on it

trgiangdo: This line has been removed. The responsibility has been moved to taipy-core on PR https://github.com/Avaiga/taipy-core/pull/689.

FabienLelaquais (approved):
Fair enough,

FredLL-Avaiga (approved):

Issue #247, opened 2023-07-19
BUG- Missing information on scenario in second client


FlorianJacta:
## Description of the issue

After creating a scenario through a Scenario selector, there is no information (in the edit button) about this scenario in the second client.

First client after creating my scenario:

![image](https://github.com/Avaiga/taipy/assets/98709993/fa56485b-0e5f-4f12-a49d-b5b8e0f9195f)

Second client:

![image](https://github.com/Avaiga/taipy/assets/98709993/b7f75352-f867-4012-831a-1b8ca2616322)

Warning in the console:
```
WARNING:root:
--- 9 warning(s) were found for page 'Prediction' in variable 'predictions' ---
 - Warning 1: Variable 'gui_core_sc_error' is not available in either the '__main__' or '__main__' modules.
 - Warning 2: Variable 'gui_core_sc_error' is not defined (in expression '{gui_core_sc_error}'):
'types.SimpleNamespace' object has no attribute 'gui_core_sc_error'
 - Warning 8: Variable 'gui_core_sv_error' is not defined (in expression '{gui_core_sv_error}'):
'types.SimpleNamespace' object has no attribute 'gui_core_sv_error'
 - Warning 9: Cannot evaluate expression 'gui_core_sv_error':
name 'gui_core_sv_error' is not defined
--------------------------------------------------------------------------------
```

## Expected behavior

No warnings should be raised and information about the scenario should be displayed.

## Version
Taipy 2.3.1

PR #249, opened 2023-07-19
#247 declare globally the core gui variables

FredLL-Avaiga:
What do you think @dinhlongviolin1 ?


dinhlongviolin1 (approved):
LGTM. I think it make sense for core gui elements since there is one instance of taipy core anyway!

FabienLelaquais (commented):
👌

Issue #232, opened 2023-06-27
Create on_create callbacks on Scenario Selector


FlorianJacta:
### Issue

Add a callback on Scenario Selector called 'on_create' that will be called after the creation of the scenario.
This will allow the user to populate the scenario and manage it.

FredLL-Avaiga:
or we could select it on creation => on_change would be enough ...
But I suppose you want to differentiate between creation and selection ...

FlorianJacta:
Right now, the workaround is on_change. This is what I have used. A user told me the same yesterday but this makes it really hard.  _on_change_ will not work (or hardly work) on many use cases.

FredLL-Avaiga:
on_creation with before or after

PR #254, opened 2023-07-21
backport on_create (#232)

FredLL-Avaiga:
Adresses #232. 

on scenario creation

can be used like:

```
def on_creation(state: State, id: str, action: str, payload: dict):
    #return "don't feel like creating a scenario now"
    #return
    scenario = create_scenario(payload.get("config"), payload.get("date"), payload.get("label"))
    with scenario as sc:
        sc.properties.update(payload.get("properties"))
    return scenario

```


src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -253,7 +257,42 @@ def crud_scenario(self, state: State, id: str, action: str, payload: t.Dict[str,
...
+                            on_creation_function,
+                            [
+                                id,
+                                on_creation,
+                                {
+                                    "config": scenario_config,
+                                    "date": date,
+                                    "name": name,
+                                    "properties": data.get("properties"),
+                                    "before": True,
```
FabienLelaquais: Do you still need this, with all the paths from on_creation?

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -253,7 +257,42 @@ def crud_scenario(self, state: State, id: str, action: str, payload: t.Dict[str,
...
+                                },
+                            ],
+                        )
+                        if isinstance(res, Scenario):
+                            # everything's fine
+                            state.assign(_GuiCoreContext._SCENARIO_SELECTOR_ERROR_VAR, "")
+                            return
+                        if res:
```
FabienLelaquais, line 284: Not even testing if it's a string? I guess that's fine...

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -253,7 +257,42 @@ def crud_scenario(self, state: State, id: str, action: str, payload: t.Dict[str,
...
+                            # do not create
+                            on_creation_function = None
+                            do_create = False
+                            state.assign(_GuiCoreContext._SCENARIO_SELECTOR_ERROR_VAR, f"{res}")
+                    except Exception as e:  # pragma: no cover
+                        if not gui._call_on_exception(on_creation, e):
+                            _warn(f"on_creation() before: Exception raised in '{on_creation}()':\n{e}")
```
FabienLelaquais: "before"?

src/taipy/gui_core/GuiCoreLib.py
```diff
@@ -253,7 +257,42 @@ def crud_scenario(self, state: State, id: str, action: str, payload: t.Dict[str,
...
+                else:
+                    on_creation_function = None
+                    _warn(f"on_creation(): '{on_creation}()' is not a function.")
```
FabienLelaquais: '{on_creation}()' -> '{on_creation}'

src/taipy/gui_core/viselements.json
```diff
@@ -55,6 +55,12 @@
             "type": "str",
             "default_value": "50vh",
             "doc": "The maximum height, in CSS units, of the tree."
+          },
+          {
+            "name": "on_creation",
+            "type": "Callback",
+            "doc": "The name of the function that is triggered when a scenario will be or has been created.<br/><br/>All the parameters of that function are optional:\n<ul>\n<li>state (<code>State^</code>): the state instance.</li>\n<li>id (str): the identifier of the button.</li>\n<li>action (str): the name of the action that provoked the change.</li>\n<li>payload (dict): the details on this callback's invocation.<br/>\nThis dictionary has the following keys:\n<ul>\n<li>before: if True, the action is invoked before the creation and the payload contains the parameters entered by the user (config, date, name, properties).</li>\n<li>after: if True, the action is invoked after the creation and the payload contains the scenario.</li>\n</ul>\n</li>\n<li>return: if before is True, the callback can return a scenario, an error message (the scenario will not be craeted) or None (the scenario will be created with the user parameters).</li>\n</ul>",
```
FabienLelaquais: You're talking about the button id? What button? What id?

No more before/after in the payload if I'm correct?

More globally, I guess this will be both simpler, and more complicated hu?

FabienLelaquais (changes requested):
Please let me see the final version

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: I don't understand your comment

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: yes it is with the f-string

src/taipy/gui_core/GuiCoreLib.py
FredLL-Avaiga: that's a console warning. It seems important to identify which on_creation has failed.
How would you differentiate the 2 calls ?

FabienLelaquais: ... I thought there would be no second call after all.
If the user wants to tweak the scenario, then he creates it (create_scenario) with the parameters you have provided, then do whatever he wants.

src/taipy/gui_core/GuiCoreLib.py
FabienLelaquais: See below

FredLL-Avaiga: do you mean that if there's no scenario it is before and if there's a scenario it is after ?
I think adding before(bool) makes things easier in the callback:
`if payload.get("before", False):`

I guess one could write 
`if payload.get("scenario"): # after`
`else: # before`

What do you prefer @FabienLelaquais ?

FabienLelaquais: Sorry, I was not clear. 

What I think is that on_creation() would be invoked right after the Create button is pressed.
That would send all that is needed to create the scenario, as I see in the code.
Now the user can add, check, modify these parameters, and decide what to do next:
- No changes whatsoever (or there is no on_creation callback): return None. The back end (of GUI) can now create the scenario as of today, and nothing else.
- There are things to set before the creation or to change afterward: the user creates the scenario himself, invoking the Core API.
   With the scenario in hand, it can perform whatever need to be done, then return it.
    The back end (of GUI) receives the scenario, and processes as if it had created it itself.
 - No scenario should be created at all. May be parameters are wrong, maybe there are already too many scenarios... nobody knows except the programmer.
    The programmer then fills in some explanatory text, and return it.
    The back end (of GUI) receives the string, and probably triggers a notification with the text (if not empty) and nothing else.
    
 Basically, as long as there is no changes to perform things work as "before" 2.4.
 If anything has to be setup before the scenario creation, or after, then the user is on his own.  He will certainly have to call `create_scenario(scenario_config, creation_date, scenario_label)` with the parameters received by the callback, and set the properties manually (they are in the callback parameters as well).
 This can be shown in a short example in the doc. We can even think of adding a scenario creation API with the CoreGui lib to simplify that creation if needed. But Core should have that of course.

FredLL-Avaiga: OK I get it

FabienLelaquais (approved):
👍

PR #258, opened 2023-08-02
Update release process

dinhlongviolin1:


FredLL-Avaiga (dismissed):

jrobinAV (dismissed):
LGTM. Even if it s a bit strange to have gui dependency required.

FabienLelaquais (dismissed):
That will work for this specific repo and image dir name, which is fine at this point.
I don't see your reluctance though...

jrobinAV (dismissed):

jrobinAV (dismissed):

tools/modify_readme.py
```diff
@@ -0,0 +1,15 @@
+import re
+import sys
+
+repo_name = sys.argv[1]
+branch_name = sys.argv[2]
+# Regex pattern <img\s+([^>]*?)(?<!['"])(?<!\/)src\s*=\s*(['"])(?!http|\/)(.*?)\2([^>]*?)>
```
FabienLelaquais, line 6: Using the r""  syntax would have prevented you to add this comment... and simplify the next line

FabienLelaquais (approved):
Looking really good!

tools/modify_readme.py
dinhlongviolin1: for double quotes you still need the \ so it would almost be the same

Issue #277, opened 2023-09-04
Data Node viewer - Layout/width issue with table view for large data


FlorianJacta:
# Issue

When the data is large, the table view will be bigger than its container and have graphical problems. 

# How to replicate

- Use the scenario-management template from this [PR](https://github.com/Avaiga/taipy-templates/pull/27). 
- Load a large data file in it.

In this code, the Data Node viewer is in a layout column but takes the entire page when a large data is loaded.

Before going to the table view:

![image](https://github.com/Avaiga/taipy/assets/98709993/83efb128-d4c9-4e8b-bb32-64b7d1b51c5b)

Table view:

![image](https://github.com/Avaiga/taipy/assets/98709993/bfba7e33-f816-4f2f-a1ac-459b208caef6)

# Expected behavior

The data view should stay in its container and not change the width of the visual element.

# Environment

Taipy Develop 3.0

FredLL-Avaiga:
added width property to datanode viewer
@jrobinAV Should the template be modified to specify a width ? width could be relative to the viewport `50vw` or calculated `calc(100% - 100px)`

FredLL-Avaiga:
@jrobinAV @trgiangdo Do you want to use this issue to adapt one or several templates ?
Shall we involve @JaufreyTpy ?

jrobinAV:
Hum... Yes. That makes sense.

Plus, it would be good to have a review of the scenario mgt template from you, @FredLL-Avaiga, since you implemented the core viz elmts.

PR #278, opened 2023-09-04
#277 width property

FredLL-Avaiga:
width could be relative to the viewport `50vw` or calculated `calc(100% - 100px)`


FabienLelaquais (approved):
👌

PR #284, opened 2023-09-06
#251 tabular views

FredLL-Avaiga:
style and refactor


FabienLelaquais (approved):
Things are far clearer as is.
I will need to see that in action to get it all though. 👍

Issue #270, opened 2023-08-30
Data Node Selector - Global Data Nodes not present in the selector


FlorianJacta:
# Issue

Global Data Nodes are not in the Data Node Selector. 

# How to replicate

Initial dataset is not in the data node selector.

```python
from taipy import Config, Scope, Frequency

import taipy as tp
import pandas as pd

def clean_data(initial_dataset):
    print("     Cleaning data")
    return initial_dataset

## Input Data Nodes
initial_dataset_cfg = Config.configure_data_node(id="initial_dataset", scope=Scope.GLOBAL, default_data=2)

## Remaining Data Node
cleaned_dataset_cfg = Config.configure_data_node(id="cleaned_dataset")

# Task config objects
clean_data_task_cfg = Config.configure_task(id="clean_data",
                                            function=clean_data,
                                            input=initial_dataset_cfg,
                                            output=cleaned_dataset_cfg,
                                            skippable=True)


# Configure our scenario config.
scenario_cfg = Config.configure_scenario(id="scenario", task_configs=[clean_data_task_cfg], frequency=Frequency.MONTHLY)

scenario = None
data_node = None

jobs = []

page = """
# Data Nodes
<|1 4|layout|
<|{data_node}|data_node_selector|>

<|{data_node}|data_node|scenario={scenario}|>
|>
"""

pages = {'/':'<|navbar|> <|toggle|theme|> <br/>',
         'Data-Node': page}


if __name__ == "__main__":
    tp.Core().run()

    tp.Gui(pages=pages).run(port=4999)
````

# Expected behavior

The "initial dataset" data node should be visible inside the data node selector but is not present.

# Environment

Taipy develop 3.0

PR #290, opened 2023-09-11
Global data nodes don't show in DataNodeSelector

FredLL-Avaiga:
Addresses #270.

add non owned datanodes


FabienLelaquais (approved):
Ha ha ha

Issue #291, opened 2023-09-12
Core Elements - Add lock_edit management to DataNode viewer


FabienLelaquais:
... with the lock reflecting the status in the header.

PR #306, opened 2023-09-27
lock management #291

FredLL-Avaiga:
for scalar values


FabienLelaquais (approved):
For scalar values... and more if I get this right.

PR #307, opened 2023-09-29
feat: Entity migration CLI

joaoandre-avaiga:


src/taipy/_cli/_migrate_cli.py
```diff
@@ -0,0 +1,35 @@
...
+from .migrate import _migrate_fs_entities
+
+
+class _MigrateCLI:
+    @classmethod
+    def create_parser(cls):
+        migrate_parser = _CLI._add_subparser(
+            "migrate", help="Migrate entities from old taipy versions to current taipy version."
+        )
+        migrate_parser.add_argument("--path", help="The path to the File System Repository folder")
```
trgiangdo: Instead of `--path`, how about `--repository-type` option with `nargs='*'` to parse all command-line arguments into a list.
- The first string is either "fs/filesystem, sql, mongodb"?
- After declaring the repo type:
  - if the repo type is fs, then the user can input a path (or default is ".data")
  - if the repo type is mongo/sql, then the user can input host, port, path, ...

The usage would be something like:
```
taipy migrate filesystem fspath/

taipy migrate sql localhost 8080 database_path.sqlite

taipy migrate mongodb localhost 8080 username password
```

src/taipy/_cli/_migrate_cli.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @classmethod
+    def parse_arguments(cls):
+        args = _CLI._parse()
+
+        if getattr(args, "which", None) == "migrate":
+            path = sys.argv[-1]  # last argument is path to FS folder
+            _migrate_fs_entities(path)
```
trgiangdo: We need to parse the argument to somehow call to `_migrate_mongo_entities()` and `_migrate_sql_entities()` as well

src/taipy/_cli/_migrate.py
```diff
@@ -0,0 +1,487 @@
...
+    entities = _migrate_entity("TASKS", entities)
+
+    __logger.info("Migrating DATANODES")
+    entities = _migrate_entity("DATANODE", entities)
+
+    __logger.info("Migrating JOBS")
+    entities = _migrate_entity("JOB", entities)
+
+    __logger.info("Migrating VERSION")
+    entities = _migrate_entity("VERSION", entities)
```
trgiangdo: This code block is duplicated on all `_migrate_..._entities()`. Probably should split it to a dedicated method.
```suggestion
    __migrate_entities(entities)
```

src/taipy/_cli/_migrate.py
```diff
@@ -0,0 +1,487 @@
...
+from typing import Dict, List, Tuple
+
+import pymongo
+
+from taipy.logger._taipy_logger import _TaipyLogger
+
+__logger = _TaipyLogger._get_logger()
+
+
+def __update_parent_ids(entity: Dict, data: Dict) -> Dict:
```
trgiangdo: Can we add a docstring clarify which Taipy version "parent_id" has been removed by "parent_ids"?

Further step may be splitting the methods that migrate from Taipy 1.0 -> Taipy 2.0 to a dedicated file, Taipy 2.0 -> Taipy 3.0 to a dedicated file, etc.

joaoandre-avaiga: I believe it was on 2.1. Theres no migration from taipy 1.0 from 2.0. Just 2.X to 3.0. If I recall correctly we had a migration from 1.0 to 2.0 in code. @jrobinAV What do you think about migrating entities from 1.0? It shouldnt be too much work and we could add it later as well, to not block the release.

src/taipy/_cli/_migrate_cli.py
joaoandre-avaiga: Great idea. I was thinking about a way to handle the different repositories. I will add your suggestion.

src/taipy/_cli/_migrate.py
jrobinAV: I believe we don t need to migrate from 1.0.

src/taipy/_cli/_migrate_cli.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class _MigrateCLI:
+    @classmethod
+    def create_parser(cls):
+        migrate_parser = _CLI._add_subparser(
+            "migrate", help="Migrate entities from old taipy versions to current taipy version."
+        )
+        migrate_parser.add_argument(
+            "--repository-type", nargs="*", help="The path to the File System Repository folder"
```
trgiangdo:
```suggestion
            "--repository-type", nargs="+", help="The repository type and additional repository properties"
```
My apology, the nargs should be "+" since we require at least 1 parameter. Also, we need to make the help string clearer for the user

src/taipy/_cli/_migrate_cli.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @classmethod
+    def parse_arguments(cls):
+        args = _CLI._parse()
+
+        if getattr(args, "which", None) == "migrate":
+            repository_type = args.repository_type[0]
+
+            if repository_type == "filesystem":
+                _migrate_fs_entities(args.repository_type[1])
```
trgiangdo: we should check if the `args.repository_type[1]` exists or not first, if not then assign to ".data"

src/taipy/_cli/_migrate_cli.py
```diff
@@ -0,0 +1,43 @@
...
+            elif repository_type == "sql":
+                _migrate_sql_entities(args.repository_type[1])
```
trgiangdo: we should check if the `args.repository_type[1]` exists or not first, if not then raise an `argparse.ArgumentError`

src/taipy/_cli/_migrate_cli.py
```diff
@@ -0,0 +1,43 @@
...
+            elif repository_type == "mongo":
+                _migrate_mongo_entities(*args.repository_type[1:])
```
trgiangdo: we should check if the `args.repository_type[1]` exists or not first, if not then we pass nothing to the method to use default value

tests/cli/test_cli.py
```diff
@@ -48,13 +48,14 @@ def clean_argparser():
...
-expected_help = """{run,manage-versions,create,help} ...
+expected_help = """{run,manage-versions,create,migrate,help} ...
 
 positional arguments:
-  {run,manage-versions,create,help}
+  {run,manage-versions,create,migrate,help}
     run                 Run a Taipy application.
     manage-versions     Taipy version control system.
     create              Create a new Taipy application.
+    migrate             Migrate entities from old taipy versions to current taipy version.
```
jrobinAV: Should we add a word saying it should be done after changing the code ?
Check if this is really the case.

jrobinAV (dismissed):

src/taipy/_cli/_migrate.py
```diff
@@ -0,0 +1,483 @@
...
+            if isinstance(timestamp, dict):
+                timestamp = timestamp.get("__value__")
+            new_edit = {"timestamp": timestamp}
+            if "job_id" in edit:
+                new_edit["job_id"] = edit["job_id"]
+            edits.append(new_edit)
+        datanode["edits"] = edits
+
+    # owner_id was not present in 2.0, need to be search for in tasks
+    datanode["owner_id"] = datanode.get("owner_id", None)
```
trgiangdo:
```suggestion
    datanode["owner_id"] = datanode.get("parent_id", None)
```
This line should migrate "parent_id" to "owner_id". Can you check again?
Also, I believe we need to pop the parent_id if it exists after migrating to owner_id

src/taipy/_cli/_migrate.py
```diff
@@ -0,0 +1,483 @@
...
+def _migrate_task(task: Dict) -> Dict:
+    # owner_id was not present in 2.0
+    if not task.get("owner_id"):
+        task["owner_id"] = task["parent_ids"][0] if task["parent_ids"] else None
+
+    # properties was not present in 2.0
+    task["properties"] = task.get("properties", {})
+
+    # skippable was not present in 2.0
+    task["skippable"] = task.get("skippable", False)
```
trgiangdo: This should also check if all output data nodes has `cacheable == True` to set it to True.

src/taipy/_cli/_migrate.py
```diff
@@ -0,0 +1,483 @@
...
+
+
+def _migrate_task_config(id: str, task: Dict, config: Dict) -> Dict:
+    task = __update_config_parent_ids(id, task, "TASK", config)
+    return _migrate_task(task)
+
+
+def _migrate_datanode(datanode: Dict) -> Dict:
+    # cacheable was removed in after 2.0
+    _ = datanode.pop("cacheable", False)
```
trgiangdo: We need to make sure to use the `cacheable` to determine the value of task's `skippable` attribute first before popping it.

src/taipy/_cli/_migrate.py
```diff
@@ -0,0 +1,483 @@
...
+                timestamp = timestamp.get("__value__")
+            new_edit = {"timestamp": timestamp}
+            if "job_id" in edit:
+                new_edit["job_id"] = edit["job_id"]
+            edits.append(new_edit)
+        datanode["edits"] = edits
+
+    # owner_id was not present in 2.0, need to be search for in tasks
+    datanode["owner_id"] = datanode.get("owner_id", None)
+
```
trgiangdo:
```suggestion
if last_edition_date := datanode.pop("last_edition_date", None):
    datanode["last_edit_date"] = last_edition_date
if edition_in_progress := datanode.pop("edition_in_progress", None):
    datanode["edit_in_progress"] = edition_in_progress

```
We also need to migrate the `last_edition_date` and `edition_in_progress` attributes.

src/taipy/_cli/_migrate.py
```diff
@@ -0,0 +1,483 @@
...
+    # additional_data_nodes attribute added
+    scenario["additional_data_nodes"] = None
+
+    return scenario
+
+
+def _migrate_task(task: Dict) -> Dict:
+    # owner_id was not present in 2.0
+    if not task.get("owner_id"):
+        task["owner_id"] = task["parent_ids"][0] if task["parent_ids"] else None
```
trgiangdo:
```suggestion
        task["owner_id"] = task["parent_id"][0] if task["parent_id"] else None
```
I believe we need to migrate `parent_id`, not `parent_ids`, to `owner_id`.
Also, if the `parent_id` exists, we need to delete it.

trgiangdo (changes requested):
There are quite a few migration method missing from the migration script:
- `_migrate_subscriber()` that rename scheduler to orchestrator on old jobs
- Migrate cacheable to skippable, mainly the part that if all outputs of a task is cacheable, then the task is skippable
- Migrate `last_edition_date` of a DataNode to `last_edit_date`
- Migrate `edition_in_progress` of a DataNode to `edit_in_progress`
- Migrate an entity that does not have the "version" field

src/taipy/_cli/_migrate.py
jrobinAV: After discussion, we proposed to add migration from 1.0 in another PR that should not block the 3.0.0 release.

src/taipy/_cli/_migrate.py
jrobinAV: well spotted.

src/taipy/_cli/_migrate.py
jrobinAV: Is not this for 1.0 migration?

joaoandre-avaiga:
The migrate cli was moved to core. This PR now contains just the entrypoint configuration for the CLI. It needs to be merged after merging the PR on taipy-core

jrobinAV (dismissed):

trgiangdo (dismissed):

jrobinAV (approved):

Issue #311, opened 2023-10-08
[BUG] - running a scenario on scenario-management template application raises error


jrobinAV:
## version: 
Latest taipy develop version
**pip freeze:** 
```
taipy @ git+https://git@github.com/Avaiga/taipy.git@e17b16ec73f202e6b4ac44cd4f962d3a3e6ffc17
taipy-config @ git+https://git@github.com/Avaiga/taipy-config.git@4c348ba2c2d0398e803f98d315950d9dd16f242c
taipy-core @ git+https://git@github.com/Avaiga/taipy-core.git@7aa0838ac02654b7fc400f81e6d798362b851458
taipy-gui @ git+https://git@github.com/Avaiga/taipy-gui.git@4236b5a08863d64a2fb90e606677d0471bab925e
taipy-rest @ git+https://git@github.com/Avaiga/taipy-rest.git@314aa2bef808f7108ec9b62d6e054297354a898c
taipy-templates @ git+https://git@github.com/Avaiga/taipy-templates.git@d48d90c445b9011d7c7d7124c9753c5a26b49444
```

## Path to reproduce:

**Scaffold the application:** 
```
jeanrobin@LAPTOP-RPTO2SFK MINGW64 ~/workspace/sand-box
$ taipy create --template scenario-management
Application root folder name [taipy_application]: taipy_application_2
Application main Python file [main.py]: 
Application title [Taipy Application]: 
Does the application use TOML Config? [No]: 
New Taipy application has been created at C:\Users\jeanrobin\workspace\sand-box\taipy_application_2

To start the application, change directory to the newly created folder:
        cd C:\Users\jeanrobin\workspace\sand-box\taipy_application_2
and run the application as follows:
        taipy run main.py
```

**Move to the folder:**
`cd taipy_application_2`

**Run the application:**
```
jeanrobin@LAPTOP-RPTO2SFK MINGW64 ~/workspace/sand-box/taipy_application_2
$ taipy run main.py
[2023-10-08 15:28:11][Taipy][INFO] Development mode: Clean all entities of version c00c5f30-e22c-4861-b414-25ea3c4f2f9c
[2023-10-08 15:28:11][Taipy][ERROR] DataNode not found: DATANODE_replacement_type_5b989245-f092-44a9-abce-38926bb1ee4a
[2023-10-08 15:28:11][Taipy][INFO]  * Server starting on http://127.0.0.1:5000
```

**Run a scenario in the UI:** 
- Select the default scenario.
[...]

PR #316, opened 2023-10-09
on_action parameter removed (#311)

FredLL-Avaiga:


FabienLelaquais (dismissed):
Far better don't you think?

jrobinAV (approved):
Without unit tests, I can only have faith in you...

FabienLelaquais (approved):
👌

Issue #319, opened 2023-10-10
Job Selector - No possiblity to delete or cancel multiple jobs


FlorianJacta:
## Issue

The button to delete or cancel multiple jobs is not clickable.

![image](https://github.com/Avaiga/taipy/assets/98709993/4e6cb9fd-7b02-4431-96b8-df248d8eed5a)

## How to replicate

Run this code, create a scenario, execute it, and try to use the button to delete jobs.

```python
from taipy import Config

import taipy as tp
import pandas as pd

def identity(input):
    print("     Identity")
    return input

## Input Data Nodes
input_cfg = Config.configure_data_node(id="input", default_data=2)

## Remaining Data Node
output_cfg = Config.configure_data_node(id="output")

# Task config objects
identity_task_cfg = Config.configure_task(id="identity",
                                            function=identity,
                                            input=input_cfg,
                                            output=output_cfg,
                                            skippable=True)


# Configure our scenario config.
scenario_cfg = Config.configure_scenario(id="my_scenario", task_configs=[identity_task_cfg])

jobs = []
scenario = None

page = """
<|{scenario}|scenario_selector|>
<|{scenario}|scenario|>

<|{jobs}|job_selector|>
"""

pages = {'/':'<|navbar|> <|toggle|theme|> <br/>',
         'Data-Node': page}


if __name__ == "__main__":
    tp.Core().run()

    tp.Gui(pages=pages).run()
```

## Expected behavior

These buttons should be clickable and should delete or cancel the selected jobs. If not possible, Taipy should better indicate why these buttons are not clickable.

## Environment 

Taipy 3.0.0.dev3

FredLL-Avaiga:
How many jobs did you select ?

FlorianJacta:
Here one but the same goes for 1, 2, ...

PR #320, opened 2023-10-10
Multiple delete Jobs (#319)

FredLL-Avaiga:


FabienLelaquais (approved):
So THAT was the issue.
Good job!

Issue #317, opened 2023-10-10
Data Node Viewer - History section not updated after an edit


FlorianJacta:
## Issue

When a data node is changed tby editing it hrough the Data section, the history section is not updated.

## Version

Taipy 3.0.0.dev3

PR #321, opened 2023-10-10
Refresh history when data is shown/edited (#317)

FredLL-Avaiga:
Would need a modification to Taipy-gui.table to get a more fine grained detection of possible history evolution.
That will do for now


FabienLelaquais (approved):
👍

PR #324, opened 2023-10-12
use the new optional feature of lov_value from gui

FredLL-Avaiga:
depends on https://github.com/Avaiga/taipy-gui/pull/970 to work
But won't break anything
<|{data_node}|data_node|>  used to show a warning about scenario not being bound


FabienLelaquais (approved):
👍

Issue #318, opened 2023-10-10
Job Selector - Jobs not appearing


FlorianJacta:
## Issue

After the first execution of a scenario, sometimes, no jobs are displayed in the job selector. 

## How to replicate

Run this code, create, and execute a scenario. Retry this same process until the job selector is not working properly.

```python
from taipy import Config

import taipy as tp
import pandas as pd

def identity(input):
    print("     Identity")
    return input

## Input Data Nodes
input_cfg = Config.configure_data_node(id="input", default_data=2)

## Remaining Data Node
output_cfg = Config.configure_data_node(id="output")

# Task config objects
identity_task_cfg = Config.configure_task(id="identity",
                                            function=identity,
                                            input=input_cfg,
                                            output=output_cfg,
                                            skippable=True)


# Configure our scenario config.
scenario_cfg = Config.configure_scenario(id="my_scenario", task_configs=[identity_task_cfg])

scenario = None

page = """
<|{scenario}|scenario_selector|>
<|{scenario}|scenario|>

<|job_selector|>
"""

pages = {'/':'<|navbar|> <|toggle|theme|> <br/>',
         'Data-Node': page}


if __name__ == "__main__":
    tp.Core().run()

    tp.Gui(pages=pages).run()

```

In this particular example, other warnings are displayed in the Python console. In another example, no warnings are displayed.

```console
[2023-10-10 12:27:00][Taipy][INFO] Python API for extension library 'taipy_gui_core' will not be available. To fix this, import 'taipy.gui.builder' before importing the extension library.
[2023-10-10 12:27:02][Taipy][INFO]  * Server starting on http://127.0.0.1:4999
TaipyGuiWarning: on_creation(): 'None' is not a function.
[2023-10-10 13:30:50][Taipy][ERROR] DataNode not found: DATANODE_input_24a83327-276e-4f79-a176-a2dc732d223a
TaipyGuiWarning: Decoding Message has failed: {'type': 'RU', 'name': '', 'payload': {'id': '', 'names': ['_TpL_tp_TpExPr_taipy_gui_core_Ctx_get_datanodes_tree_TPMDL_1_0'], 'refresh': True}, 'propagate': True, 'client_id': '20230302140156934952-0.38052708359870924', 'ack_id': '30c73300-c0b2-4d83-b966-d43351197eb2', 'module_context': '__main__'}:      
[...]

PR #328, opened 2023-10-13
refresh job_selector (#318)

FredLL-Avaiga:
Addresses #318.

Refresh jobs after submission

### DIFF
--- a/src/taipy/gui_core/_context.py
+++ b/src/taipy/gui_core/_context.py
@@ -119,18 +119,20 @@ class _GuiCoreContext(CoreEventConsumerBase):
         # locks
         self.lock = Lock()
         self.submissions_lock = Lock()
+        # super
         super().__init__(reg_id, reg_queue)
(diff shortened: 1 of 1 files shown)


src/taipy/gui_core/_context.py
```diff
@@ -119,18 +119,20 @@ def __init__(self, gui: Gui) -> None:
         # locks
         self.lock = Lock()
         self.submissions_lock = Lock()
+        # super
```
FabienLelaquais, line 122: ?

FabienLelaquais (approved):
I really don't know what this does...

src/taipy/gui_core/_context.py
FredLL-Avaiga: - sends a broadcast event each time a job is modified AND after submission
- does *not* refresh the scenario on submission
- a small bug (compare a enum value to an enum)

PR #344, opened 2023-10-23
Two small bugs on badly placed variables declaration

FabienLelaquais:


src/taipy/gui_core/_context.py
```diff
@@ -143,7 +144,9 @@ def process_event(self, event: Event):
                         _GuiCoreContext._CORE_CHANGED_NAME, {"scenario": [x for x in sequence.parent_ids]}
                     )
             except Exception as e:
-                _warn(f"Sequence ({sequence.id if hasattr(sequence, 'id') else 'No_id'}) access raised an issue", e)
```
FredLL-Avaiga, line 146: a better fix would be to use` event.entity_id` instead of `sequence.id if hasattr(sequence, 'id') else '<anonymous>'`
No need to declare sequence before then

FredLL-Avaiga (dismissed):

src/taipy/gui_core/_context.py
FabienLelaquais: Works for me

src/taipy/gui_core/_context.py
```diff
@@ -132,6 +132,7 @@ def process_event(self, event: Event):
                 event.entity_id if event.operation != EventOperation.DELETION and is_readable(event.entity_id) else None
             )
         elif event.entity_type == EventEntityType.SEQUENCE and event.entity_id:
+            sequence = None
```
FredLL-Avaiga, line 135: can be removed, no ?

FredLL-Avaiga (approved):

Issue #343, opened 2023-10-21
Scenario selector control - Issues with on_creation action


FabienLelaquais:
When requesting the creation of a scenario, the on_creation callback is invoked.

There are two bad problems that must be addressed:

This callback expects a 'payload' dictionary parameter.
- The payload has a 'label' parameter representing the user's 'label' setting. This is documented as 'name'.
- The payload has a 'properties' parameter that is documented as 'a dictionary' and nothing else.
   I was expecting that dictionary to hold the user-defined custom properties set at creation time.
   But the received dictionary is empty no matter what.
   What is the use of this empty dictionary? Should the user add its own properties in on_creation then return None to see these properties automatically added to the new scenario? Where are the user-defined user properties that the callback would certainly be willing to make use of?

FredLL-Avaiga:
do you prefer label or name @FabienLelaquais ? Label it is

Issue #342, opened 2023-10-21
Scenario control - Cannot modify a custom property


FabienLelaquais:
Create a new property on a scenario, then try to modify it.

The text fields (both key and value) will always bring you back to the original values whatever you type.

**Workaround**
Create a new property, and delete the first one.
(A refresh bug, though: when you delete the property, it temporarily disappears from the list, then reappears, then finally goes away).

Issue #341, opened 2023-10-21
UX improvements for the Scenario viewer


FabienLelaquais:
Tooltip on icons. The minimum would be:
- Submit buttons -> resp. "Submit scenario" and "Submit sequence"
      (or "Cannot submit [scenario|sequence]" - explanation would be even better but hey...)
- [✓] next to editable field -> "Apply"
- [×] next to editable field -> "Cancel"
- Trash icon next to edited custom property -> "Delete property"

Keyboard shortcuts:
- 'Enter' when editing the label, tags, custom properties or a sequence name should be equivalent to clicking the [✓] button.
- 'Esc' when editing the label, tags, custom properties or a sequence name should be equivalent to clicking the [×] button.

FredLL-Avaiga:
👍

FredLL-Avaiga:
tags: Enter is use to validate a tag
properties: There are 2 text fields when do we apply Enter ?

PR #346, opened 2023-10-23
- improvements (tooltip, keys) (#341#342 #343)

FredLL-Avaiga:
- custom properties #342
- on_creation #343


FabienLelaquais (approved):
Looks (very) good to me!

PR #347, opened 2023-10-24
tooltips

FredLL-Avaiga:
and limit unnecessary renders on scenario viewer

### DIFF
--- a/gui/src/ScenarioViewer.tsx
+++ b/gui/src/ScenarioViewer.tsx
@@ -11,16 +11,7 @@
  * specific language governing permissions and limitations under the License.
  */
 
-import React, {
-    useState,
-    useCallback,
-    useEffect,
-    useMemo,
-    ChangeEvent,
-    SyntheticEvent,
-    MouseEvent,
-    KeyboardEvent,
-} from "react";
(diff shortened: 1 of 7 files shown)


gui/src/ScenarioViewer.tsx
```diff
@@ -479,14 +476,9 @@ const ScenarioViewer = (props: ScenarioViewerProps) => {
                             </Grid>
                             <Grid item>
                                 {showSubmit ? (
-                                    <Tooltip title={disabled ? "Cannot submit Sequence" : "Submit Scenario"}>
+                                    <Tooltip title={disabled ? "Cannot submit Scenario" : "Submit Scenario"}>
```
FabienLelaquais, line 471: ? 
disabled is the discriminant between Scenarios and Sequences??

FabienLelaquais (dismissed):
👍

gui/src/ScenarioViewer.tsx
FredLL-Avaiga: fixed

FabienLelaquais (approved):
👌

PR #351, opened 2023-10-24
Improve Enter and Escape key handling

FredLL-Avaiga:
Tags : support for Esc
Properties: Esc in both Textfields, Enter in value Text Field
Elsewhere: support for Enter & Esc


FabienLelaquais (approved):
👍

Issue #355, opened 2023-10-25
BUG- Warning using on_submission_change property


FlorianJacta:
## Issue

A warning appears when using the on_submission_change callback:

```console
TaipyGuiWarning: Job (JOB_double_dda36d55-6e75-4ef6-9d51-2f24195df564) is not available:        
Working outside of application context.

This typically means that you attempted to use functionality that needed
the current application. To solve this, set up an application context
with app.app_context(). See the documentation for more information.
my_scenario has completed.
TaipyGuiWarning: Job (JOB_add_7472ec92-1859-4c8a-9b0a-a4fa438d9c76) is not available:
Working outside of application context.

This typically means that you attempted to use functionality that needed
the current application. To solve this, set up an application context
with app.app_context(). See the documentation for more information.
```

## How to replicate

Run this code

```python
from taipy.config import Config
import taipy as tp
from taipy.gui import Gui, notify


# Normal function used by Taipy
def double(nb):
    return nb * 2

def add(nb):
    return nb + 10


# Configuration of Data Nodes
input_cfg = Config.configure_data_node("input", default_data=21)
intermediate_cfg = Config.configure_data_node("intermediate")
output_cfg = Config.configure_data_node("output")


# Configuration of tasks
first_task_cfg = Config.configure_task("double",
                                       double,
                                       input_cfg,
                                       intermediate_cfg)

second_task_cfg = Config.configure_task("add",
                                        add,
                                        intermediate_cfg,
                                        output_cfg)



# Configuration of scenario
scenario_cfg = Config.configure_scenario(id="my_scenario",
                                         task_configs=[first_task_cfg, second_task_cfg],
                                         name="my_scenario")



def notify_from_submissions(state, submittable, details):
    submission_status = details.get('submission_status')

    if submission_status == 'COMPLETED':
        print(f"{[submittable.name](http://submittable.name/)} has completed.")
        notify(state, 'success', 'Completed!')
        # Add additional actions here, like updating the GUI or logging the completion.

    elif submission_status == 'FAILED':
        print(f"{[submittable.name](http://submittable.name/)} has failed.")
        notify(state, 'error', 'Completed!')
        # Handle failure, like sending notifications or logging the error.


if __name__=="__main__":
    tp.Core().run()
    scenario_1 = tp.create_scenario(scenario_cfg)

    scenario_1.submit(wait=True)

    scenario_md = """
<|{scenario_1}|scenario|on_submission_change=notify_from_submissions|>
[...]

PR #356, opened 2023-10-25
no state to get submission fn (#355)

FredLL-Avaiga:
#355 
and hide scenario selector button in data node viewer when owner is GLOBAL

### DIFF
--- a/gui/src/DataNodeViewer.tsx
+++ b/gui/src/DataNodeViewer.tsx
@@ -645,34 +645,38 @@ const DataNodeViewer = (props: DataNodeViewerProps) => {
                                             <Typography variant="subtitle2">{dnOwnerLabel}</Typography>
                                         </Grid>
                                         <Grid item xs={1}>
-                                            <Tooltip title="Show Scenarios">
-                                                <span>
-                                                    <IconButton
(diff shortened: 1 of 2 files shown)


gui/src/DataNodeViewer.tsx
```diff
@@ -645,34 +645,38 @@ const DataNodeViewer = (props: DataNodeViewerProps) => {
                                             <Typography variant="subtitle2">{dnOwnerLabel}</Typography>
                                         </Grid>
                                         <Grid item xs={1}>
-                                            <Tooltip title="Show Scenarios">
-                                                <span>
-                                                    <IconButton
-                                                        sx={tinySelPinIconButtonSx}
-                                                        onClick={showScenarios}
-                                                        disabled={!valid}
+                                            {dnOwnerId ? (
```
FabienLelaquais, line 648: 🙏

FabienLelaquais (approved):
👌

gui/src/DataNodeViewer.tsx
FredLL-Avaiga: BowingThankYouGIF

PR #362, opened 2023-10-27
Remove setup_requires

dinhlongviolin1:
Yesterday, we have a problem with the release process due to a problem when executing the setup_requires process in setup.py. The problem is due to a dependency PACKAGE NAME incompatibility that we have no control over (zope.interface != zope-interface).

If we remove the setup_requrires process on the develop branch similarly like what we are about to do on the release branch, you can no longer install taipy directly from develop branch. 

We can, however, in the next version of taipy-gui, moved to a previous version of `twisted` package which does not have the typo issue with zope.interface package. If we don't change the twisted version, the setup_requires process will not work on taipy package and we might as well remove it on the develop branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,14 +71,12 @@ jobs:
           python -m pip install --upgrade pip
           pip install build wheel
           # install taipy-gui from based on setup.py version
+          pip install "${{ steps.taipy_gui_version.outputs.VERSION }}"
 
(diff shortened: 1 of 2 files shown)


florian-vuillemot (approved):

FabienLelaquais (approved):
So that was radical

jrobinAV (approved):

PR #488, opened 2023-11-21
Merge taipy-templates

joaoandre-avaiga:


trgiangdo (commented):
The linter action is failing because templates use different linting rules due to some files are duplicated for different template.

We probably need to modify the linter action for templates.

trgiangdo (commented):
Again, if possible, I would like to have the test action also runs taipy-templates tests in this PR.

jrobinAV (dismissed):

src/__init__.py
jrobinAV, line 1: THis file should be removed, right?

jrobinAV (approved):

src/__init__.py
joaoandre-avaiga: This was removed on core's PR.

PR #511, opened 2023-11-30
fix deadlink on README

considine:
Use updated demo page


jrobinAV (approved):
Good catch. Thank you for your contribution!

PR #517, opened 2023-12-01
🙈 slightly improve .gitignore file

gmarabout:
# Goal
Ignore common or intermediary files (mostly the `node_modules` in the `gui` folder).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,17 @@
-.idea
+# Environment files
+.env
 .venv
-__pycache__
+.envrc
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
 
 # Distribution / packaging
 Pipfile.lock
 build/
 *.egg-info/
+
 # mypy
 .mypy_cache/
 dist/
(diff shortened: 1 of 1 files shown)


toan-quach (approved):
lgtm!

.gitignore
```diff
@@ -1,11 +1,17 @@
-.idea
```
jrobinAV, line 1: Why remove the .idea?

.gitignore
```diff
@@ -22,3 +28,6 @@ dist/
 # demo files
 demo-*
 demo_*/
+
+# Gui node_modules
+gui/node_modules
```
gmarabout, line 33: **note** : not useful after all. The `gui` folder is a reminiscence from the past and should no longer be there.

FredLL-Avaiga: Will you remove the line then ?

FredLL-Avaiga (approved):

.gitignore
gmarabout: Because it's duplicated.

.gitignore
gmarabout: Yes. Not sure this PR makes sense any longer. closing it.

PR #519, opened 2023-12-01
Event Creation: Add version and make sure we don't call autoreload attributes

gmarabout:
# Goal
* Make sure all events are created without accessing fields that would "auto reload" the entity (and thus could create performance issues)
* Add `version` in metadata of entities 

# Changed
* Data_node: add `_version`
* Job: Use internal fields to by-pass autoreload
* Scenario: Use internal fields to by-pass autoreload
* Submission: small refactoring
* Cycle: no change, there is no version for it?


toan-quach (approved):
lgtm!

jrobinAV (approved):

PR #523, opened 2023-12-01
Make _TaskFunctionWrapper an object class

gmarabout:
# Goals
We need to add authorisation and telemetry at certain specific spot of the code when a task function is executed.
The current implementation of `_TaskFunctionWrapper` is based on class methods and is not overridable.

# Changes
* We make `_TaskFunctionWrapper` a real class. All methods are now instance methods, no more "classmethods".
* The execution of the function is encapsulated into a `__execute_fct` method which can be overridden in Taipy-enterprise.
* We basically replaced inheritance by composition in `JobDispatcher` and subclasses.


jrobinAV (dismissed):
Looks ok to me.

jrobinAV (approved):

joaoandre-avaiga (approved):

PR #544, opened 2023-12-04
Split test and fix it.

jrobinAV:
- I stopped using InMemory data nodes that are not compatible with standalone mode
- I split a test into three to make it easier to maintain.
- I fixed an inconsistency regarding some scenario-scoped data nodes shared across multiple scenarios.


toan-quach (approved):
lgtm, just need to see if test pasts successfully

PR #569, opened 2023-12-05
Feature/add config id to submission entity

toan-quach:
# Changes:

- Added `config_id` to Submission entity
- If Sequence is submitted, `None` will be passed as `config_id` for Submission entity creation


src/taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -67,7 +68,9 @@ def submit(
             The created Jobs.
         """
         submission = _SubmissionManagerFactory._build_manager()._create(
-            submittable.id, submittable._ID_PREFIX  # type: ignore
+            submittable.id,  # type: ignore
+            submittable._ID_PREFIX,  # type: ignore
+            submittable.config_id if isinstance(submittable, Scenario) else None,  # type: ignore
```
jrobinAV:
```suggestion
            getattr(submittable, 'config_id ', None)  # type: ignore
```

jrobinAV (dismissed):

taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -67,7 +68,9 @@ def submit(
             The created Jobs.
         """
         submission = _SubmissionManagerFactory._build_manager()._create(
-            submittable.id, submittable._ID_PREFIX  # type: ignore
+            submittable.id,  # type: ignore
```
gmarabout, line 71: 🥜 the `type: ignore` is probably not uselful. Or is it?

taipy/core/submission/submission.py
```diff
@@ -48,6 +48,7 @@ def __init__(
         self,
         entity_id: str,
         entity_type: str,
+        entity_config_id: Optional[str] = None,
```
gmarabout, line 51: ❓ It is my understanding that the "config_id" will not exist in every cases. For instance, it does make sense for tasks and scenarios, but not at all for cycles nor sequences. Would it make more sense to use a `Properties` object to store this kind of additional/optional information? 
Like here https://github.com/Avaiga/taipy/blob/develop/taipy/core/task/task.py#L83.

By the way, I think that `Properties` might well become a field of the `_Entity` type.

gmarabout (approved):
I have just one comment on how we store the `config_id` information in the class, but beside that, it loks good to me! 👏🏻

taipy/core/_orchestrator/_orchestrator.py
toan-quach: it is since the type of the submittable here is the `Submittable` parent class. And this class doesn't have `id` defined

taipy/core/submission/submission.py
toan-quach: 1. **Would it make more sense to use a Properties object to store this kind of additional/optional information?**
-> hmm I don't think so as it's a system attribute instead of an additional/optional information. So putting it in the `_Properties` will put a different semantic meaning to the field.


2. **`Properties` might well become a field of the `_Entity` type?**
-> That's a valid point. But we don't have `_Properties` for all entity classes (`_Version`, `Submission`, `Job`)

jrobinAV (approved):

taipy/core/submission/submission.py
gmarabout: > So putting it in the _Properties will put a different semantic meaning to the field.

That was exactly my point. This field is optional and does not make sense in every cases. It could perfectly match the `Properties` purpose.

toan-quach: hmmm but then it would make `config_id` more of an additional property would it not? When to me it's a system property and not an additional one?

gmarabout: ok.

PR #604, opened 2023-12-13
minor cleaning

jrobinAV:
Fixing some minor type-checking stuff in tests.

### DIFF
--- a/taipy/core/taipy.py
+++ b/taipy/core/taipy.py
@@ -625,7 +625,7 @@ def get_jobs() -> List[Job]:
     return _JobManagerFactory._build_manager()._get_all()
 
 
-def delete_job(job: Job, force=False):
+def delete_job(job: Job, force: Optional[bool] = False):
     """Delete a job.
 
     This function deletes the specified job. If the job is not completed and
--- a/tests/core/test_taipy.py
+++ b/tests/core/test_taipy.py
@@ -16,6 +16,7 @@ import shutil
 from unittest import mock
 
 import pytest
+
 import taipy.core.taipy as tp
 from taipy.config.common.frequency import Frequency
 from taipy.config.common.scope import Scope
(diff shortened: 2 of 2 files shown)


tests/core/test_taipy.py
```diff
@@ -439,36 +431,48 @@ def test_compare_scenarios(self, scenario):
             mck.assert_called_once_with(scenario, scenario, data_node_config_id="dn")
 
     def test_subscribe_scenario(self, scenario):
+        def cb(s, j):
+            print()
```
trgiangdo, line 435: Should we put this function at the file scope?
So that we don't have to redefine it multiple times

taipy/core/taipy.py
```diff
@@ -625,7 +625,7 @@ def get_jobs() -> List[Job]:
     return _JobManagerFactory._build_manager()._get_all()
 
 
-def delete_job(job: Job, force=False):
+def delete_job(job: Job, force: Optional[bool] = False):
```
trgiangdo, line 628: This method returns None, but somehow we do have the
```
return _JobManagerFactory._build_manager()._delete(job, force)
```

The `return` is redundant, which is confusing but very minor

tests/core/test_taipy.py
```diff
@@ -544,40 +548,19 @@ def test_cancel_job(self):
...
-        with mock.patch("sys.argv", ["prog"]):
-            core = Core()
-            core.run()
-
-        scenario_1 = tp.create_scenario(scenario_cfg_1)
-        tp.submit(scenario_1)
-
-        with pytest.raises(ConfigurationUpdateBlocked):
-            Config.configure_scenario("block_scenario", set([task_cfg_1]))
-        core.stop()
```
trgiangdo, line 562: Why are we removing this test?

trgiangdo (approved):
I have some minor questions. Otherwise, LGTM!

tests/core/test_taipy.py
jrobinAV: We don't need to submit a scenario to test that the config is blocked. So whatever it is standalone or development, the config is blocked.
That means this test `test_block_config_when_core_is_running_in_development_mode` and the following one `test_block_config_when_core_is_running_in_standalone_mode` are doing the same.
So I kept only one and renamed it.

trgiangdo: That's reasonable. Thank you

PR #583, opened 2023-12-08
bug/ move template test files to general test folder

toan-quach:

PR #608, opened 2023-12-13
chore: refactor tests workflow

joaoandre-avaiga:


joaoandre-avaiga (commented):
Test

.github/workflows/linter.yml
trgiangdo, line 1: I can understand why this file is removed.

However, since we do not have another alternative linter action yet, is this ready to be removed?

.github/workflows/overall-tests.yml
```diff
@@ -0,0 +1,59 @@
+name: Overall Test Workflow
+
+on:
+  pull_request_review:
+    types: [submitted]
```
trgiangdo, line 5: For example, if JR approves a PR, and Fabien approves it right after, does it mean the action will be run twice?
Or it will only be run once because there is no change between the 2 approves?

.github/workflows/overall-tests.yml
```diff
@@ -0,0 +1,59 @@
...
+
+jobs:
+  tests:
+    if: github.event.review.state == 'approved'
+    timeout-minutes: 40
```
trgiangdo, line 10: I would increase the timeout, probably 60 minutes

.github/workflows/partial-tests.yml
```diff
@@ -0,0 +1,106 @@
...
+            core:
+              - 'taipy/core/**'
+            gui:
+              - 'taipy/gui/**'
+            gui-core:
+              - 'taipy/gui_core/**'
+            logger:
+              - 'taipy/logger/**'
+            rest:
+              - 'taipy/rest/**'
```
trgiangdo, line 53: I would add `templates` here as well

.github/workflows/partial-tests.yml
```diff
@@ -0,0 +1,106 @@
...
+        if: steps.changes.outputs.gui-core == 'true'
+        run: pipenv run pytest  tests/gui_core
+
+      - name: Pytest Logger
+        if: steps.changes.outputs.logger == 'true'
+        run: pipenv run pytest  tests/logger
+
+      - name: Pytest Rest
+        if: steps.changes.outputs.rest == 'true'
+        run: pipenv run pytest  tests/rest
```
trgiangdo, line 108: I would add `templates` here as well

Pipfile
```diff
@@ -57,7 +57,7 @@ pandas-stubs = "*"
 playwright = "*"
 pre-commit = "*"
 pyopenssl = "*"
-pytest = "*"
+pytest = "==7.4.3"
```
trgiangdo, line 60: Is this necessary?

.github/workflows/linter.yml
joaoandre-avaiga: We do have a linter action. It's part of the partial tests workflow. The linter is the first job of the workflow, if it fails the unit tests are not even executed.

joaoandre-avaiga: For now is the same linter we used before, but theres a comment explaining that will be replaced by `ruff` in the future.

.github/workflows/overall-tests.yml
joaoandre-avaiga: I believe the action will run again, which is not really desirable, but I don't see a way to prevent this at the moment. I still things it's better like this than run on every push, but we can discuss it.

.github/workflows/overall-tests.yml
joaoandre-avaiga: At this point, it's not needed. Tests are taking 14min to 30 min  depending on OS and quantity of jobs running. We can increase it in the future if this becomes a problem, but I think we should focus on optmizing the test run to take less time, since 14 ~ 30 min is already too much.

.github/workflows/partial-tests.yml
joaoandre-avaiga: I didn't add templates because the templates test is still inside the package, it needs to be moved to the root test folder. I can add but effectively will only work after the tests are moved to the correct folder.

Pipfile
joaoandre-avaiga: It's not really necessary, but I think having no fixed version, or at least no fixed major version can be bad. But I can move back to the `*` notation.

.github/workflows/partial-tests.yml
trgiangdo: PR https://github.com/Avaiga/taipy/pull/583 from @toan-quach has already did that, however we did not merge it because of the timeout.
We need to align on this topic

.github/workflows/partial-tests.yml
jrobinAV: I agree.

.github/workflows/partial-tests.yml
```diff
@@ -0,0 +1,112 @@
...
+      - uses: actions/checkout@v4
+      - uses: ricardochaves/python-lint@v1.4.0
+        with:
+          use-pylint: false
+          use-isort: false
+          use-mypy: false
+          extra-black-options: "--line-length=120 --diff"
+          extra-pycodestyle-options: "--max-line-length=120  --exclude=tests/gui --ignore=E121,E123,E126,E226,E24,E704,W503,W504,E203"
+          extra-mypy-options: "--ignore-missing-imports --implicit-optional --no-namespace-packages --exclude (taipy/templates/|generate_pyi.py|tools) --follow-imports skip"
+          extra-isort-options: "--line-length=120 --force-grid-wrap=10 --multi-line=VERTICAL_HANGING_INDENT --trailing-comma"
```
jrobinAV, line 23: Just by curiosity why did you moved linter from a separate workflow to the partial-tests workflow?

jrobinAV (dismissed):

.github/workflows/partial-tests.yml
joaoandre-avaiga: Linter is usually one of the first steps on a CI pipeline. I think it makes more sense to build it like a workflow together with unit tests and first execute the linter, if that step fails, unit tests aren't executed. For me it makes more manageable grouping related jobs into a single workflow instead of managing separate files.

joaoandre-avaiga (commented):
.

jrobinAV (approved):

Issue #379, opened 2023-10-05
BUG-Inputting the wrong column name in a chart reports empty data


AlexandreSajus:
**Description**
Inputting the wrong column name in a chart visual element will report that the dataset used is empty when it is not. This leads the user to think that the issue is with the dataset instead of the column name.

**How to reproduce**
Running this code where I made a mistake on the y column of the chart visual element:

```python
from taipy import Gui
import pandas as pd

data = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})

page = """
<|{data}|chart|type=bar|x=a|y=hello|>
"""

Gui(page).run()
```

Leads to this error:
```
--- 1 warning(s) were found for page '/' in variable 'page' ---
 - Warning 1: Error column "hello" is not present in the Dataframe "Empty DataFrame
Columns: [a, b]
Index: []".
```

**Expected behavior**
The error should not mention that the DataFrame is empty, only that the column is not found

**Runtime environment**
 - OS: Windows 11
 - Browser: Chrome

srinivaspavan9:
Hi  @AlexandreSajus  I have started working on this issue, I am trying to debug the code to find out where the warnings are being generated but it seems that I am not able to step into the Gui(page).run() line after hitting that breakpoint. what am i doing wrong ? and also i changed the first line to "from taipy.gui import Gui" as it was giving me import error. 
I need help

AlexandreSajus:
@srinivaspavan9 Yes you are right about the import error, this code is probably outdated. I'm not sure what you mean about the breakpoint issue. Maybe @jrobinAV could help?

srinivaspavan9:
> @srinivaspavan9 Yes you are right about the import error, this code is probably outdated. I'm not sure what you mean about the breakpoint issue. Maybe @jrobinAV could help?

I was trying to debug the code to find out where the warnings are being generated. But I am unable to debug because I am not able to get into the Gui(Page).run(), I mean right after hitting the break point the line is being skipped while debugging.

Can I know where the warnings are generated?

AlexandreSajus:
> > @srinivaspavan9 Yes you are right about the import error, this code is probably outdated. I'm not sure what you mean about the breakpoint issue. Maybe @jrobinAV could help?
> 
> I was trying to debug the code to find out where the warnings are being generated. But I am unable to debug because I am not able to get into the Gui(Page).run(), I mean right after hitting the break point the line is being skipped while debugging.
> 
[...]

PR #617, opened 2023-12-18
Updated utils.py for better warning messages

srinivaspavan9:
Updated Warning Message in utils.py for Issue #379 

This PR addresses BUG-Inputting the wrong column name in a chart reports empty data (#379). The issue was due to the misleading warning message in `utils.py`, which implied an empty DataFrame when a non-existent column name was used.

Changes Made:
- Updated the warning message to clearly indicate when a column is not found, and to list the available columns in the DataFrame.
- This change aims to provide clearer and more accurate information, avoiding confusion about the state of the DataFrame.

Screenshot after the change:

image


taipy/gui/_renderers/utils.py
```diff
@@ -37,7 +37,7 @@ def _get_columns_dict_from_list(
             idx += 1
         elif col:
             _warn(
-                f'Error column "{col}" is not present in the Dataframe "{value.head(0) if hasattr(value, "head") else value}".'  # noqa: E501
+                f'Error column "{col}" is not present. Available columns: {list(value.columns)}.'  # noqa: E501
```
FredLL-Avaiga: that would need a check on value

FredLL-Avaiga (changes requested):

taipy/gui/_renderers/utils.py
srinivaspavan9: I didn't get you

FredLL-Avaiga: Python is very permissive and if we use the value as a dataframe in the message we'd better test that value *is* a dataframe.

srinivaspavan9: Got it

taipy/gui/_renderers/utils.py
```diff
@@ -29,6 +30,11 @@ def _get_tuple_val(attr: tuple, index: int, default_val: t.Any) -> t.Any:
 def _get_columns_dict_from_list(
     col_list: t.Union[t.List[str], t.Tuple[str]], col_types_keys: t.List[str], value: t.Any
 ):
+    #Check if 'Value' is a Dataframe
```
FredLL-Avaiga: I meant the check should only be done at the warning message level

FredLL-Avaiga (changes requested):
check also the linter comments

FredLL-Avaiga (approved):
👌

taipy/gui/_renderers/utils.py
```diff
@@ -36,9 +37,13 @@ def _get_columns_dict_from_list(
             col_dict[col] = {"index": idx}
             idx += 1
         elif col:
-            _warn(
```
FredLL-Avaiga, line 39: I think it would be better if we calculate the list of columns depending on the type of value here
something like 
`cols = str(list(value.columns) if isinstance(value, dataframe) else list(value.keys()) if isinstance(value, (dict, _MapDict)) else value if isinstance(value, (list, tuple)) ... else value)`

FredLL-Avaiga (changes requested):

taipy/gui/_renderers/utils.py
srinivaspavan9: > I think it would be better if we calculate the list of columns depending on the type of value here
> 
> something like 
> 
> `cols = str(list(value.columns) if isinstance(value, dataframe) else list(value.keys()) if isinstance(value, (dict, _MapDict)) else value if isinstance(value, (list, tuple)) ... else value)`

Oh I can do that I thought we were just using data frames. What all data types does the function support?

FredLL-Avaiga: dataframe, dict, list ... I suppose this would be enough
`cols = str(list(value.columns) if isinstance(value, dataframe) else list(value.keys()) if isinstance(value, (dict, _MapDict)) else value if isinstance(value, (list, tuple)) else value)
`

srinivaspavan9: Added the required changes

srinivaspavan9:
how do i resolve these linter errors

taipy/gui/_renderers/utils.py
FredLL-Avaiga: I can't see them ?

trgiangdo:
> how do i resolve these linter errors

Hello @srinivaspavan9, sorry for the late response.

First you need to merge your branch with the latest Avaiga:develop branch. Then, in the linter error log, I saw that
```
/taipy/gui/_renderers/utils.py:35:121: E501 line too long (191 > 120 characters)
```
which means that the line you edited is too long.

Please address the issue before we can merge your change to Taipy. Thank you

FredLL-Avaiga:
Hi @srinivaspavan9,
Thanks for your work.
Taipy 3.1 is coming soon and it would be nice to include your contribution.
We're freezing the code very soon (likely tomorrow). Can you finish this PR before that deadline ?
Fred

PR #619, opened 2023-12-18
Feature/clean test orchetrator

jrobinAV:
Refactor test_orchetrator.py to shorten the computation time:
- Split it into multiple files. Each file is dedicated to testing a specific method
- Make tests more `unitary` testing dispatchers and orchestrators separately (using mocks)
- Add test of FunctionWrapper.
- I marked some remaining tests in test_orchestrator.py with `@pytest.mark.orchestrator_dispatcher`


taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -182,23 +170,22 @@ def _orchestrate_job_to_run_or_block(cls, jobs: List[Job]):
             cls.jobs_to_run.put(job)
 
     @classmethod
-    def __wait_until_job_finished(cls, jobs: Union[List[Job], Job], timeout: Optional[Union[float, int]] = None):
-        def __check_if_timeout(start, timeout):
-            if timeout:
-                return (datetime.now() - start).seconds < timeout
+    def _wait_until_job_finished(cls, jobs: Union[List[Job], Job], timeout: Optional[Union[float, int]] = None):
+        #  Note: this method should be prefixed by two underscores, but it has only one, so it can be mocked in tests.
```
trgiangdo, line 174: We can mock it by calling to `_Orchestrator __wait_until_job_finished`. However, I agree that it's not a good practice

tests/core/_orchestrator/_dispatcher/test_dispatcher__execute_job.py
```diff
@@ -0,0 +1,130 @@
...
+from tests.core.utils import assert_true_after_time
+
+
+def nothing(*args):
+    return
+
+
+def create_scenario():
+    dn_cfg = Config.configure_pickle_data_node("dn")
+    t1_cfg = Config.configure_task("t1", nothing, [], [dn_cfg])
```
trgiangdo, line 31: Why don't we config the task to be skippable here so we don't have to do that manually at each test?

tests/core/_orchestrator/_dispatcher/test_development_job_dispatcher.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_dispatch_executes_the_function_with_exceptions():
+    task = create_task()
+    job = Job(JobId("job"), task, "s_id", task.id)
+    dispatcher = _OrchestratorFactory._build_dispatcher()
+    e_1 = Exception("test")
+    e_2 = Exception("test")
+
+    with patch("taipy.core._orchestrator._dispatcher._task_function_wrapper._TaskFunctionWrapper.execute") as mck:
+        mck.return_value = [e_1, e_2]
```
trgiangdo, line 55: Instead of mocking the return value, how about executing a function that raise the exception?

tests/core/_orchestrator/_dispatcher/test_dispatcher__needs_to_run.py
```diff
@@ -0,0 +1,103 @@
...
+def test_need_to_run_task_not_skippable():
+    hello_cfg = Config.configure_data_node("hello", default_data="Hello ")
+    world_cfg = Config.configure_data_node("world", default_data="world !")
+    hello_world_cfg = Config.configure_data_node("hello_world")
+    task_cfg = Config.configure_task(
+        "name", input=[hello_cfg, world_cfg], function=nothing, output=[hello_world_cfg], skippable=False
+    )
+    task = _create_task_from_config(task_cfg)
+
+    assert _JobDispatcher(_OrchestratorFactory._build_orchestrator())._needs_to_run(task)
```
trgiangdo, line 47: Since the skippable by default is False, is this test necessary?

Or this is to test future change? In that case just ignore my comment haha

tests/core/_orchestrator/_dispatcher/test_dispatcher__needs_to_run.py
```diff
@@ -0,0 +1,103 @@
...
+
+def test_need_to_run_skippable_task_no_validity_period_on_output():
+    hello_cfg = Config.configure_data_node("hello", default_data="Hello ")
+    output_cfg = Config.configure_data_node("output")
+    task_cfg = Config.configure_task("name", input=[hello_cfg], function=nothing, output=[output_cfg], skippable=True)
+    task = _create_task_from_config(task_cfg)
+    dispatcher = _JobDispatcher(_OrchestratorFactory._build_orchestrator())
+    assert dispatcher._needs_to_run(task)  # output data is not written
+    task.output["output"].write("Hello world !")
+    assert not dispatcher._needs_to_run(task)  # output data is written
```
trgiangdo, line 68: Since the validity_period by default is None, is this test necessary?
Or this is to test future change? In that case just ignore my comment haha

taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -182,23 +170,22 @@ def _orchestrate_job_to_run_or_block(cls, jobs: List[Job]):
             cls.jobs_to_run.put(job)
 
     @classmethod
-    def __wait_until_job_finished(cls, jobs: Union[List[Job], Job], timeout: Optional[Union[float, int]] = None):
-        def __check_if_timeout(start, timeout):
-            if timeout:
-                return (datetime.now() - start).seconds < timeout
+    def _wait_until_job_finished(cls, jobs: Union[List[Job], Job], timeout: Optional[Union[float, int]] = None):
```
gmarabout, line 173: 🎨 Note related to the PR: `Union[float, int]` could actually be replaced by `float` because `int` is implied in the type hint `float`. See https://peps.python.org/pep-0484/#the-numeric-tower.

So I guess the signature could eventually be: 
`def _wait_until_job_finished(cls, *args: Job, timeout: Optional[float] = None)` 🎉

tests/core/_orchestrator/_dispatcher/mock_standalone_dispatcher.py
```diff
@@ -0,0 +1,79 @@
...
+        return "result"
+
+
+class MockProcessPoolExecutor(Executor):
+    submit_called = []
+    f = []
+
+    def submit(self, fn, *args, **kwargs):
+        self.submit_called.append((fn, args, kwargs))
+        f = Future()
```
gmarabout, line 25: ❓ Shouldn't we use the `FutureMock` here? Asking this because I see no use the `FutureMock` anywhere...

tests/core/_orchestrator/_dispatcher/test_standalone_job_dispatcher.py
```diff
@@ -0,0 +1,133 @@
...
+    orchestrator = _OrchestratorFactory._build_orchestrator()
+    orchestrator.jobs_to_run.put(job_1)
+    orchestrator.jobs_to_run.put(job_2)
+    orchestrator.jobs_to_run.put(job_3)
+    orchestrator.jobs_to_run.put(job_4)
+
+    with mock.patch("taipy.core._orchestrator._dispatcher._job_dispatcher._JobDispatcher._execute_job") as mck:
+        dispatcher = _StandaloneJobDispatcher(orchestrator)
+        dispatcher.start()
+        assert_true_after_time(lambda: mck.call_count == 4, msg="The 4 jobs were not dequeued.", time=5)
```
trgiangdo, line 132: Why is `time=5`?

tests/core/_orchestrator/_dispatcher/test_standalone_job_dispatcher.py
```diff
@@ -0,0 +1,133 @@
...
+        dispatcher.stop()
+        mck.assert_called_with(job_4)
```
trgiangdo:
```suggestion
        mck.assert_called_with(job_1)
        mck.assert_called_with(job_2)
        mck.assert_called_with(job_3)
        mck.assert_called_with(job_4)
```
Can this be done to make sure all jobs are executed?

tests/core/_orchestrator/_dispatcher/test_task_function_wrapper.py
```diff
@@ -0,0 +1,138 @@
...
+        output_dn_config_id,
+        {},
+        function=function,
+        input=input_dn,
+        output=output_dn,
+    )
+
+
+def multiply(nb1: float, nb2: float):
+    sleep(0.1)
```
trgiangdo: Why do we need the sleep here?

tests/core/_orchestrator/test_orchestrator__submit.py
```diff
@@ -0,0 +1,492 @@
...
+
+    # task output should be locked for edition
+    assert sc.dn_1.edit_in_progress
+    assert sc.dn_2.edit_in_progress
+    assert sc.dn_3.edit_in_progress
+
+    # jobs are created in a specific order and are correct
+    assert len(jobs) == 4
+    # t1
+    job_1 = jobs[0]
```
gmarabout, line 227: 💬 I wonder if the following code could be factorized to avoid the repetitions... 
That would make the test slightly more readable IMHO.

tests/core/_orchestrator/_dispatcher/test_task_function_wrapper.py
```diff
@@ -0,0 +1,138 @@
...
+def test_data_node_not_written_due_to_wrong_result_nb():
+    def fct_2_outputs():
+        return lambda nb1, nb2: (multiply(nb1, nb2), multiply(nb1, nb2) / 2)
+
+    task_expecting_3_outputs = _create_task(fct_2_outputs, 3)
+
+    res = _TaskFunctionWrapper("job_id", task_expecting_3_outputs).execute()
+
+    assert len(res) == 1
+    assert isinstance(res[0], Exception)
```
trgiangdo:
```suggestion
    exceptions = _TaskFunctionWrapper("job_id", task_expecting_3_outputs).execute()

    assert len(exceptions) == 1
    assert isinstance(exceptions[0], Exception)
```
I would name this `exceptions` for better readability of the test. It's not important though.

tests/core/_orchestrator/_dispatcher/test_task_function_wrapper.py
```diff
@@ -0,0 +1,138 @@
...
+
+        Config.core.storage_folder = ".new_storage_folder/"
+        return n * m
+
+    task_updating_cfg = _create_task(update_config_fct)
+    cfg_as_str = _TomlSerializer()._serialize(Config._applied_config)
+    res = _TaskFunctionWrapper("job_id", task_updating_cfg).execute(config_as_string=cfg_as_str)
+
+    assert len(res) == 1
+    assert isinstance(res[0], ConfigurationUpdateBlocked)
```
trgiangdo, line 116:
```suggestion
    exceptions = _TaskFunctionWrapper("job_id", task_updating_cfg).execute(config_as_string=cfg_as_str)

    assert len(exceptions) == 1
    assert isinstance(exceptions[0], ConfigurationUpdateBlocked)
```

tests/core/_orchestrator/_dispatcher/test_task_function_wrapper.py
```diff
@@ -0,0 +1,138 @@
...
+        assert Config.core.custom_property == "custom_property"
+        return n * m
+
+    Config.configure_core(storage_folder=".my_data/", custom_property="custom_property")
+
+    task_asserting_cfg_is_correct = _create_task(assert_config_is_correct_after_serialization)
+    cfg_as_str = _TomlSerializer()._serialize(Config._applied_config)
+    res = _TaskFunctionWrapper("job_id", task_asserting_cfg_is_correct).execute(config_as_string=cfg_as_str)
+
+    assert len(res) == 0  # no exception raised so the asserts in the fct passed
```
trgiangdo, line 133:
```suggestion
    exceptions = _TaskFunctionWrapper("job_id", task_asserting_cfg_is_correct).execute(config_as_string=cfg_as_str)

    assert len(exceptions) == 0  # no exception raised so the asserts in the fct passed
```

gmarabout (commented):
Huge work ! 👏🏻 Thank you @jrobinAV to tackle this big refactoring!
I added a couple of comments. Basically, beside the fact that the tests are failing in timeout and that needs to be investigated on its own, I think some tests are very repetitive and could be improved by more factorization.

tests/core/_orchestrator/_dispatcher/test_development_job_dispatcher.py
jrobinAV: This is possible, but you must create a function and the corresponding task. It s more code to understand.
In the end, we don't need to execute anything. We are just asserting that execute is called, and when exceptions are returned, the job is correctly updated. 
It does not bring any value to me to run a function.

tests/core/_orchestrator/_dispatcher/test_dispatcher__execute_job.py
jrobinAV: I did it on purpose to make it explicit in the test itself. So, we don't need to understand the `create_scenario` function to get the test's purpose.

tests/core/_orchestrator/_dispatcher/test_dispatcher__needs_to_run.py
jrobinAV: No, you are right. I believe it is not necessary.

tests/core/_orchestrator/_dispatcher/test_dispatcher__needs_to_run.py
jrobinAV: I am not sure the default validity period is tested elsewhere. If so, I am going to remove this test.

tests/core/_orchestrator/_dispatcher/mock_standalone_dispatcher.py
jrobinAV: I removed the FuturMock class.

tests/core/_orchestrator/_dispatcher/test_standalone_job_dispatcher.py
jrobinAV: The dispatcher thread must have time to dequeue 4 jobs. It takes a few hundred of milliseconds on my laptop, but who knows about GitHub actions? 
So 5 is a random small value. 🤣 
Any better value?

tests/core/_orchestrator/_dispatcher/test_standalone_job_dispatcher.py
jrobinAV: image

source: https://docs.python.org/3/library/unittest.mock.html

However I can use `assert_has_calls` instead

tests/core/_orchestrator/_dispatcher/test_task_function_wrapper.py
jrobinAV: Good catch. Thx!

tests/core/_orchestrator/test_orchestrator__submit.py
jrobinAV: Yes, probably. Let me think about it.

florian-vuillemot:
Should we wait for the decision on https://github.com/Avaiga/taipy/issues/631?

jrobinAV:
> Should we wait for the decision on #631?

I would say no. I just deactivated some tests with side effects. Whatever the outcome of the decision, we don't want these tests to run in this action. 

What do you think?

trgiangdo (dismissed):

trgiangdo (dismissed):

trgiangdo (dismissed):

trgiangdo (approved):

Issue #411, opened 2023-10-05
Improve checker for configuration ids


FlorianJacta:
**What would that feature address**
Improve the checker of configuration names to deal with restricted names. For example, when the user calls a data node "name" or "default"

The restricted IDs correspond to all the attributes and properties of the following classes: Scenario, Sequence, Task, and DataNode (and children classes).

trgiangdo:
I'm not sure to understand the requirement here @jrobinAV @FlorianJacta 
Does it mean the following configuration:
```python
dn_1_config = Config.configure_data_node(
    id="dn_1_id",
    storage_type="pickle",
    scope=Scope.SCENARIO,
    name="name"
    )
```
will raise an error in the ConfigChecker? Similar to `name="default"`.

Or `id="name"` will raise an error?

jrobinAV:
![image](https://github.com/Avaiga/taipy/assets/88036007/d395560a-66f1-47ba-a89e-69300f24956a)

jrobinAV:
Yes indeed. Scenario has an attribute called `name`. Thus `name` is restricted as a data_node_config_id because at run time, `scenario.name` would not be possible to retrieve the data node.

So `Config.configure_data_node(id="name", ...)` should raise a configchecker error.

PR #620, opened 2023-12-19
Feature/#411 - Raise error if config_id is an entity attribute

trgiangdo:
https://github.com/Avaiga/taipy/issues/411

### DIFF
--- a/taipy/core/data/data_node.py
+++ b/taipy/core/data/data_node.py
@@ -16,6 +16,7 @@ from datetime import datetime, timedelta
 from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
 import networkx as nx
+
 from taipy.config.common._validate_id import _validate_id
 from taipy.config.common.scope import Scope
 from taipy.logger._taipy_logger import _TaipyLogger
@@ -103,9 +104,9 @@ class DataNode(_Entity, _Labeled):
         editor_expiration_date: Optional[datetime] = None,
         **kwargs,
     ):
-        self.config_id = _validate_id(config_id)
+        self._config_id = _validate_id(config_id)
         self.id = id or DataNodeId(self.__ID_SEPARATOR.join([self._ID_PREFIX, self.config_id, str(uuid.uuid4())]))
-        self.owner_id = owner_id
+        self._owner_id = owner_id
         self._parent_ids = parent_ids or set()
         self._scope = scope
         self._last_edit_date = last_edit_date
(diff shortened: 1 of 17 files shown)


taipy/core/data/data_node.py
```diff
@@ -120,6 +121,22 @@ def __init__(
...
+        self._config_id = val
+
+    @property
+    def owner_id(self):
+        return self._owner_id
+
+    @owner_id.setter
+    def owner_id(self, val):
+        self._owner_id = val
+
```
jrobinAV: I believe it was on purpose not to have the properties.

Both `config_id` and owner_id must not be editable. You cannot change the config_id of a data node at runtime.

taipy/core/scenario/scenario.py
```diff
@@ -156,6 +155,14 @@ def __getattr__(self, attribute_name):
...
         raise AttributeError(f"{attribute_name} is not an attribute of scenario {self.id}")
 
+    @property
+    def config_id(self):
+        return self._config_id
+
+    @config_id.setter
+    def config_id(self, val):
+        self._config_id = val
+
```
jrobinAV: It must not be editable.

taipy/core/task/task.py
```diff
@@ -109,6 +107,22 @@ def properties(self):
...
+        self._config_id = val
+
+    @property
+    def owner_id(self):
+        return self._owner_id
+
+    @owner_id.setter
+    def owner_id(self, val):
+        self._owner_id = val
+
```
jrobinAV: Same here. They must not be editable.

jrobinAV (changes requested):
I believe this PR is promising but does not cover the use cases. It prevents a `data_node_config` from being named like a data node attribute, a `scenario_config` from being named like a scenario attribute, etc. 
But actually we want to prevent a `data_node_config` from having the name of a scenario attribute.

1. For instance, a data_node_config cannot be named "tags" because we want to be able to do 'scenario.<data_node_config_name>.read()' and if <data_node_config_name> is "tags", there is a conflict with the 'tags' property.
2. Another question, why is it a pb to name a scenario "tags", or a data_node "skippable". I cannot see any conflict.

taipy/core/data/data_node.py
trgiangdo: It was editable before since the `config_id` was public. To avoid it to be editable, we can remove the setter

jrobinAV: yes, indeed.

trgiangdo:
In the latest commits:
- The id of a `DataNodeConfig` should not be named as a Task or Scenario attribute
- The id of a `TaskConfig` should not be named as a Scenario attribute

jrobinAV (approved):

PR #624, opened 2023-12-20
Add markers mode to decimators

FredLL-Avaiga:
so that @vhfgosselin can use them with markers only


FabienLelaquais (approved):
Looks very reasonable.

PR #634, opened 2023-12-25
Indexed properties in the Page Builder API

### LINKED ISSUE #616: Support indexed properties in Page Builder
Some properties are indexed. The syntax is *property_name[index]*. 
This of course cannot be a parameter name, from the Page Builder API.

We need to invent a way to express this anyway. `property_(index, ...)`?

dinhlongviolin1:
Resolves #616

The updated regular expression will handle double underscore followed by numbers and characters: `example__abc123`

These will have the same result
```
page = """
<|{data}|chart|mode=lines|x=Date|y[1]=Temp°C|y[2]=Min|y[3]=Max|line[1]=dash|color[2]=blue|color[3]=red|>
"""
```
```
with tgb.Page() as tgb_page:
    tgb.chart("{data}", mode="lines", x="Date", y__1="Temp°C", y__2="Min", y__3="Max", line__1="dash",color__2="blue", color__3="red")
```


taipy/gui/builder/_element.py
```diff
@@ -28,6 +29,7 @@ class _Element(ABC):
 
     _ELEMENT_NAME = ""
     _DEFAULT_PROPERTY = ""
+    __RE_INDEXED_PROPERTY = re.compile(r"^(.*?)_([\d]+)$")
```
FredLL-Avaiga: No need for the square brackets

FredLL-Avaiga (dismissed):
Joyeux Noël ☃️

taipy/gui/builder/_element.py
dinhlongviolin1: The square brackets is used for the declaration of the allowed chars, in this case numbers and is not included in the allowed chars

taipy/gui/builder/_element.py
```diff
@@ -28,6 +29,7 @@ class _Element(ABC):
 
     _ELEMENT_NAME = ""
     _DEFAULT_PROPERTY = ""
+    __RE_INDEXED_PROPERTY = re.compile(r"^(.*?)(__([\w\d]+)|_([\d]+))$")
```
FredLL-Avaiga: Should we need the 2 ?
ie have only 2 underscores as separator...

FredLL-Avaiga (dismissed):

FabienLelaquais (approved):
As easy as your ABC.
Well done!

PR #635, opened 2023-12-26
Use variable as kwarg value

### LINKED ISSUE #628: Consider a simpler syntax for tgb property values
The botton line is that the "{variable}" syntax is too confusing for some users.

For tgb: users have to constantly put their properties within quotes and brackets which is very confusing. For example, 
```python
tgb.table("{table_dict}", properties=table_properties)
```
does not work but:
```python
tgb.table("{table_dict}", properties="{table_properties}")
```
does

This is way too confusing and is not the expected syntax when coding in pure Python with the tgb. The user should be able to input his variable directly even if this means getting the variable name and adding "{...}" on our end

dinhlongviolin1:
Resolves #628.

This will work 
```
from taipy.gui import Gui

a = 10
b = 12


with tgb.Page() as page:
    tgb.text(a)
    tgb.slider(a)
    tgb.text(b)
    tgb.slider(b)
    tgb.text("{a + b}")

Gui(page=page).run()

```


taipy/gui/builder/_element.py
```diff
@@ -37,30 +40,63 @@ def __new__(cls, *args, **kwargs):
         return obj
 
     def __init__(self, *args, **kwargs):
+        source_frame = inspect.stack()[3]
+        self.__locals = {
+            key: value
+            for key, value in t.cast(FrameType, source_frame.frame).f_locals.items()
+            if not key.startswith("__")
```
FabienLelaquais: Do we need to care about variables starting with a single '_'?

FabienLelaquais (dismissed):
If this works, it's amazing!
This introduces a slight confusion compared to the MD syntax, where MD forces to reference variables in a f-string format, and not TGB anylonger. But I think we can deal with that.

dinhlongviolin1:
> If this works, it's amazing!
> 
> This introduces a slight confusion compared to the MD syntax, where MD forces to reference variables in a f-string format, and not TGB anylonger. But I think we can deal with that.

It is working already so you can try it out if you want

taipy/gui/builder/_element.py
dinhlongviolin1: Just want to filter out the double underscore / private variable so we can go through less variables when processing

FabienLelaquais: Yes I understand that.
My question can be shorten to: why not '`if not key.startswith("_")`'

dinhlongviolin1:
Due to special cases, such as expression within a string, the PR will be closed as it is not a fool proof solution. 
```
# This take on the string instead of the expression that is defined within the string
s = "{a + b}"
with tgb.Page() as page:
    # these are not the same
    tgb.text(s)
    tgb.text("{a+b}")
```

PR #637, opened 2023-12-27
feature/#612 exposed submission api to taipy core apis

### LINKED ISSUE #612: Expose Submission and submission ids to the relevant taipy core APIs
- Make the all the relevent APISs accept submission objects or submission ids as parameters so users can easily manage the submissions.
- Ensure user manual is up-to-date
- Ensure rf manual is up-to-date, in particular this page has to be fixed: https://docs.taipy.io/en/develop/manuals/reference/pkg_taipy.core.submission.submission
- Remove todo in Core visual elements: `taipy/gui_core/_context.py` line 847

toan-quach:
Resolves #612.

# Changes
- Added exception `SubmissionNotDeletedException` to be raised when deleting an un-finished submission entity
- added `is_deletable` for submission (false by default for submission and job) and `is_readable` (true by default)
- check deletable before deleting a submission
- added submission entity to various exposed taipy core api
- added unit tests


taipy/core/submission/_submission_manager.py
```diff
@@ -53,3 +54,26 @@ def _get_latest(cls, entity: Union[Scenario, Sequence, Task]) -> Optional[Submis
...
+
+    @classmethod
+    def _delete(cls, submission: Union[Submission, SubmissionId]):
+        if isinstance(submission, str):
+            submission = cls._get(submission)
+        if cls._is_deletable(submission):
+            super()._delete(submission.id)
+        else:
+            err = SubmissionNotDeletedException(submission.id)
+            cls._logger.warning(err)
```
florian-vuillemot: Warning and not exception?

taipy/core/submission/_submission_manager.py
```diff
@@ -53,3 +54,26 @@ def _get_latest(cls, entity: Union[Scenario, Sequence, Task]) -> Optional[Submis
...
+            raise err
+
+    @classmethod
+    def _is_deletable(cls, submission: Union[Submission, SubmissionId]) -> bool:
+        if isinstance(submission, str):
+            submission = cls._get(submission)
+        if submission.is_finished():
+            return True
+        return False
```
florian-vuillemot:
```suggestion
        return submission.is_finished()
```

taipy/core/taipy.py
```diff
@@ -141,6 +169,8 @@ def is_readable(
         return _JobManagerFactory._build_manager()._is_readable(entity)  # type: ignore
     if isinstance(entity, DataNode) or (isinstance(entity, str) and entity.startswith(DataNode._ID_PREFIX)):
         return _DataManagerFactory._build_manager()._is_readable(entity)  # type: ignore
+    if isinstance(entity, Submission) or (isinstance(entity, str) and entity.startswith(Submission._ID_PREFIX)):
+        return _SubmissionManagerFactory._build_manager()._is_readable(entity)  # type: ignore
```
florian-vuillemot: This condition is duplicate (lines 130 and 401). Can you create a dedicated function that better explains what this condition does?

florian-vuillemot (dismissed):
Refactoring is possible

taipy/core/submission/_submission_manager.py
toan-quach: Does it make more sense to be an error instead? @jrobinAV what do you think?

taipy/core/taipy.py
toan-quach: done :D

taipy/core/submission/submission.py
```diff
@@ -201,6 +201,30 @@ def _update_submission_status(self, job: Job):
...
+
+        Returns:
+            True if the submission is finished.
+        """
+        submission_status = self.submission_status
+        return (
+            submission_status == SubmissionStatus.COMPLETED
+            or submission_status == SubmissionStatus.FAILED
+            or submission_status == SubmissionStatus.CANCELED
+            or submission_status == SubmissionStatus.UNDEFINED
```
florian-vuillemot:
```suggestion
            submission_status in [SubmissionStatus.COMPLETED, SubmissionStatus.FAILED, SubmissionStatus.CANCELED, SubmissionStatus.UNDEFINED]
```

taipy/core/taipy.py
```diff
@@ -102,24 +125,39 @@ def is_editable(
...
-        return _TaskManagerFactory._build_manager()._is_editable(entity)  # type: ignore
-    if isinstance(entity, Job) or (isinstance(entity, str) and entity.startswith(Job._ID_PREFIX)):
-        return _JobManagerFactory._build_manager()._is_editable(entity)  # type: ignore
-    if isinstance(entity, DataNode) or (isinstance(entity, str) and entity.startswith(DataNode._ID_PREFIX)):
-        return _DataManagerFactory._build_manager()._is_editable(entity)  # type: ignore
+    if _is_cycle(entity):
+        return _get_manager(Cycle._MANAGER_NAME)._is_editable(entity)
+    if _is_scenario(entity):
+        return _get_manager(Scenario._MANAGER_NAME)._is_editable(entity)
+    if _is_sequence(entity):
```
florian-vuillemot: These conditions/return are made several times, can we factor?

taipy/core/_entity/_reload.py
```diff
@@ -106,4 +107,6 @@ def _get_manager(manager: str):
         "job": _JobManagerFactory._build_manager(),
         "task": _TaskManagerFactory._build_manager(),
         "submission": _SubmissionManagerFactory._build_manager(),
-    }[manager]
+    }[
+        manager
+    ]  # type: ignore
```
jrobinAV, line 112: I am not sure this is standard formating. Is it?
(I hope it is not 🤣)

taipy/core/exceptions/exceptions.py
```diff
@@ -151,6 +151,16 @@ def __init__(self, job_id: str):
         self.message = f"Job: {job_id} does not exist."
 
 
+class SubmissionNotDeletedException(RuntimeError):
+    """Raised if there is an attempt to delete a submission that cannot be deleted.
+
+    This exception can be raised by `taipy.delete_job()^`.
```
jrobinAV: Wrong copy paste?

taipy/core/submission/_submission_manager.py
jrobinAV: I am OK with a warning, but I don't have a strong opinion on it.

taipy/core/submission/submission.py
jrobinAV: For me, an UNDEFINED submission status, does not mean the submission is finished.

taipy/core/submission/_submission_manager.py
```diff
@@ -53,3 +54,24 @@ def _get_latest(cls, entity: Union[Scenario, Sequence, Task]) -> Optional[Submis
...
+        else:
+            err = SubmissionNotDeletedException(submission.id)
+            cls._logger.warning(err)
+            raise err
+
+    @classmethod
+    def _is_deletable(cls, submission: Union[Submission, SubmissionId]) -> bool:
+        if isinstance(submission, str):
+            submission = cls._get(submission)
+        return submission.is_finished()
```
jrobinAV: is_finished or undefined. No?

taipy/core/taipy.py
```diff
@@ -913,10 +980,10 @@ def get_entities_by_config_id(
 
     entities: List = []
 
-    if entities := _ScenarioManagerFactory._build_manager()._get_by_config_id(config_id):
+    if entities := _get_manager(Scenario._MANAGER_NAME)._get_by_config_id(config_id):  # type: ignore
```
jrobinAV: WHy do we need a type ignore here?

taipy/core/job/_job_manager.py
```diff
@@ -92,3 +92,7 @@ def _is_deletable(cls, job: Union[Job, JobId]) -> bool:
         if job.is_finished():
             return True
         return False
+
+    @classmethod
+    def _is_editable(cls, entity: Union[Job, str]) -> bool:
+        return False
```
jrobinAV, line 98: So, a job is never editable?

jrobinAV (approved):

jrobinAV (changes requested):
UNDEFINED status should not be considered as finished.

taipy/core/submission/_submission_manager.py
toan-quach: I'll keep it as error log then

taipy/core/_entity/_reload.py
toan-quach: ah hahahaha idk my linter just auto fixed that, lemme fix it again

taipy/core/taipy.py
toan-quach: the `_get_by_config_id` is not available in _Manager subclass which is the return type of `_get_manager`

taipy/core/job/_job_manager.py
toan-quach: not by the user no? sounds weird that a job can be edited to me, no?

taipy/core/taipy.py
toan-quach: Done, can you help review again?

taipy/core/job/_job_manager.py
jrobinAV: Yep, ok. It is fair.

taipy/core/taipy.py
jrobinAV, line 1: The factorization gives better code, but it looks strange to me.
Indeed, when we get a manager or another, we are not necessarily calling the same method (interface). We are calling two different methods that happened to have the same name and the attribute names. It is by chance and not by design. I guess that's why we have a type ignore.

The only method where it make sense to me are the methods that are declared in the Manager abstract class. That's it.
What do you think?

taipy/core/taipy.py
```diff
@@ -102,24 +104,27 @@ def is_editable(
...
-    if isinstance(entity, Job) or (isinstance(entity, str) and entity.startswith(Job._ID_PREFIX)):
-        return _JobManagerFactory._build_manager()._is_editable(entity)  # type: ignore
-    if isinstance(entity, DataNode) or (isinstance(entity, str) and entity.startswith(DataNode._ID_PREFIX)):
-        return _DataManagerFactory._build_manager()._is_editable(entity)  # type: ignore
+    if manager := _check_entity_and_get_manager(entity, [Cycle, Scenario, Sequence, Task, Job, DataNode, Submission]):
+        return manager._is_editable(entity)
     return False
 
 
 def is_readable(
```
jrobinAV, line 164: We want to overload the method for all the types. Like e did for the `exist` method for example. It could be part of another PR though. Up to you.

taipy/core/taipy.py
```diff
@@ -368,6 +354,8 @@ def delete(entity_id: Union[TaskId, DataNodeId, SequenceId, ScenarioId, JobId, C
       the scenario can be deleted.
```
jrobinAV, line 466: When we delete a scenario, we want to delete the associated submission. (same for the sequence.)

taipy/core/taipy.py
```diff
@@ -368,6 +354,8 @@ def delete(entity_id: Union[TaskId, DataNodeId, SequenceId, ScenarioId, JobId, C
...
     - If a `SequenceId` is provided, the related jobs are deleted.
     - If a `TaskId` is provided, the related data nodes, and jobs are deleted.
+    - If a `SubmissionId^` or a `JobId^` is provided, the submission or job entity can only be deleted if
```
jrobinAV: When I delete a submission are the related jobs deleted as well? It must be to be consistent.

taipy/core/taipy.py
```diff
@@ -368,6 +354,8 @@ def delete(entity_id: Union[TaskId, DataNodeId, SequenceId, ScenarioId, JobId, C
       the scenario can be deleted.
     - If a `SequenceId` is provided, the related jobs are deleted.
```
jrobinAV, line 467: There is no doc for data_nodes

taipy/core/taipy.py
jrobinAV: We are reaching the factorization limits.

taipy/core/common/_check_entity_and_get_manager.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2023 Avaiga Private Limited
```
jrobinAV:
```suggestion
# Copyright 2021-2024 Avaiga Private Limited
```

taipy/core/taipy.py
toan-quach: not sure what you meant by overloading for all types?

taipy/core/taipy.py
toan-quach: well what you said is true, but then it does show the abstract class doesn't represent well the child class. for example, _hard_delete() can be added to _Manager also

taipy/core/taipy.py
toan-quach: yep it does, `_get_children_entity_ids` in `taipy/core/scenario/_scenario_manager.py` also contains finding the `Submission` under the scenario so it can be deleted along with it as well

taipy/core/taipy.py
toan-quach: added

taipy/core/taipy.py
jrobinAV: The thing is that we don't want to have an hard_delete method for jobs or data nodes for instance. 

I believe we are Overengineering here. We don't need to add a method to get_manager from the manager name. As a result, we get an AbstractManager instead of a proper instance that does not have all the methods we want. This also leads to some unnecessary type ignore comments.

taipy/core/_entity/_reload.py
```diff
@@ -89,7 +90,7 @@ def _do_set_entity(self, *args, **kwargs):
 
 
 @functools.lru_cache
-def _get_manager(manager: str):
+def _get_manager(manager: str) -> _Manager:
```
jrobinAV, line 93: I believe we should remove `_get_manager` method. It is overengineering to me.
It does not make the code easier to read or maintain.
It is just an indirection to me.

toan-quach: It's still useful in the `_reload.py` context. We can consider removing it from `taipy.py` due to overengineering. But I think it's still needed in the _Reload class

jrobinAV: Yes, in the _reload.py, it makes sense to keep it. I agree.

taipy/core/taipy.py
```diff
@@ -769,11 +835,12 @@ def clean_all_entities_by_version(version_number=None) -> bool:
         __logger.warning(f"{e.message} Abort cleaning the entities of version '{version_number}'.")
         return False
 
-    _JobManagerFactory._build_manager()._delete_by_version(version_number)
+    _SubmissionManagerFactory._build_manager()._delete_by_version(version_number)
     _ScenarioManagerFactory._build_manager()._delete_by_version(version_number)
     _SequenceManagerFactory._build_manager()._delete_by_version(version_number)
     _TaskManagerFactory._build_manager()._delete_by_version(version_number)
     _DataManagerFactory._build_manager()._delete_by_version(version_number)
+    _SubmissionManagerFactory._build_manager()._delete_by_version(version_number)
```
jrobinAV: SubmissionManagerFactory used twice.

jrobinAV:
The code factorization we want to perform does not apply to `taipy.py`. It breaks the linter.
My feeling is that:
- We don't want to use abstract managers in these functions because we want to manipulate the implementation and call specific methods that are not defined in the Abstract parent Manager class.
- We don't want to use abstract entity or ids because it breaks the type checking.
- The factorization here does not bring more readability or maintainability. It reduces verbosity though.

So, I propose to revert the factorization. I pushed a commit to revert the factorization on `_set` and `is_submittable` functions. @toan-quach Please tell me what you think. If you agree, I or you can perform the rest of the revert.

toan-quach:
@jrobinAV Yep I agree with you. I wont be able to do it within today but can tomorrow 😀 let me know what you think?

jrobinAV:
> @jrobinAV Yep I agree with you. I wont be able to do it within today but can tomorrow 😀 let me know what you think?

OK, let me propose something. I'll try to push a proposal today.

toan-quach:
Awesome thanks!!! @jrobinAV

jrobinAV (approved):
Looks good to me. 

Integration-testing must be fixed asap in a separate PR.

PR #644, opened 2024-01-03
Add colors to marker  properties that can hold a column name

### LINKED ISSUE #622: Add the possibility in treemap to set a color according to a specific value and a size depending on a weight set on the sector
Enhancement proposition : Add the possibility in treemap to set a color according to a specific value and a size depending on a weight set on the sector 
the weight being a third element in addition to the value and to the name
it could look like this see the two files attached
testmap
and
[map.csv](https://github.com/Avaiga/taipy/files/13716357/map.csv)

FredLL-Avaiga:
Resolves #622.

`marker = dict(colors="colors_col")`


dinhlongviolin1 (approved):
Great great one liner PR

FabienLelaquais (approved):
You must be kidding me

Issue #357, opened 2023-10-25
BUG- on_submission_change called twice with same status


FlorianJacta:
## Issue

The callback is called twice with the same status. 

## How to replicate

Run this code and see that two notifications are emitted. 

```python
from taipy.config import Config
import taipy as tp
from taipy.gui import Gui, notify


# Normal function used by Taipy
def double(nb):
    return nb * 2

def add(nb):
    return nb + 10


# Configuration of Data Nodes
input_cfg = Config.configure_data_node("input", default_data=21)
intermediate_cfg = Config.configure_data_node("intermediate")
output_cfg = Config.configure_data_node("output")


# Configuration of tasks
first_task_cfg = Config.configure_task("double",
                                       double,
                                       input_cfg,
                                       intermediate_cfg)

second_task_cfg = Config.configure_task("add",
                                        add,
                                        intermediate_cfg,
                                        output_cfg)



# Configuration of scenario
scenario_cfg = Config.configure_scenario(id="my_scenario",
                                         task_configs=[first_task_cfg, second_task_cfg],
                                         name="my_scenario")



def notify_from_submissions(state, submittable, details):
    submission_status = details.get('submission_status')

    if submission_status == 'COMPLETED':
        print(f"{[submittable.name](http://submittable.name/)} has completed.")
        notify(state, 'success', 'Completed!')
        # Add additional actions here, like updating the GUI or logging the completion.

    elif submission_status == 'FAILED':
        print(f"{[submittable.name](http://submittable.name/)} has failed.")
        notify(state, 'error', 'Completed!')
        # Handle failure, like sending notifications or logging the error.


if __name__=="__main__":
    tp.Core().run()
    scenario_1 = tp.create_scenario(scenario_cfg)

    scenario_1.submit(wait=True)

    scenario_md = """
<|{scenario_1}|scenario|on_submission_change=notify_from_submissions|>
<|{scenario_1.output if scenario_1 else None}|data_node|>
"""
    Gui(scenario_md).run()
```

## Environment

**Taipy 3.0**

FredLL-Avaiga:
Needs a blocking event queue to make event management cleaner

jrobinAV:
IS it a duplicate or the ticket #475 ?

FlorianJacta:
Yes, it is a duplicate. I guess I wanted it to be solved, or my memory is fading. Do I close the other issue?

jrobinAV:
[...]

PR #647, opened 2024-01-04
feature/#357 replace submission entity with submission status

toan-quach:
#357

# Changes

I made some changes to store submission status instead of submission entity to avoid unnecessary reloading that will cause some conditions to be wrong


FredLL-Avaiga (approved):
LGTM

jrobinAV (approved):

PR #649, opened 2024-01-04
Fix multi user error with dict (#566)

### LINKED ISSUE #566: BUG-Multi-user fails when state.data["X"] = ...
**Description**
A user on Discord reported this. Multi-user and state-bound variables do not work properly when we modify data using `state.data["X"] = ...`

**How to reproduce**
Using the following code:
```python
from taipy.gui import Gui

data = {"X": range(1, 6), "Y": [x * x for x in range(1, 6)]}

page = """
<|{data}|table|show_all|rebuild|>
<|button|label=Double X and Y|on_action=double_x_and_y|>
"""


def double_x_and_y(state):
    state.data["X"] = [2 * x for x in state.data["X"]]
    state.data["Y"] = [2 * x for x in state.data["Y"]]


Gui(page=page).run()
```

1. Run the app in browser A
2. Connect to the app in an incognito page B
3. Press the button in browser A
4. Refresh incognito page B
5. Incognito page B will show the data modified while no button has been pressed in B

**Expected behavior**
Modifying data using slices should modify the state-bound variable, not the original one. Incognito page B should show the original data.

**Runtime environment**
 - OS: Windows 11
 - Browser: Chrome
 - Taipy: 2.4.0

dinhlongviolin1:
Resolves #566 .


FabienLelaquais (approved):
👌

taipy/gui/utils/_map_dict.py
```diff
@@ -23,7 +24,7 @@ class _MapDict(object):
     __local_vars = ("_dict", "_update_var")
 
     def __init__(self, dict_import: dict, app_update_var=None):
-        self._dict = dict_import
+        self._dict = copy.deepcopy(dict_import)
```
FredLL-Avaiga, line 27: I'm afraid that will generate a *lot* of copying.
I guess we would need to add a static method to _MapDict for internal use that would *not* do the deepcopy and would prevent some undesirable side effects 
ex: _MapDict.__getitem__ (_MapDict.py:L36)

FredLL-Avaiga (changes requested):
I'm not even sure I want us to handle this
if we handle this, should we handle lists, class instances ...
I think we can't 
Maybe we should even deprecate _MapDict (if possible)
the main point of _MapDict was to allow dictionnary access through a dotted notation which (I guess) nobody uses and is very non pythonic ...

One of the first answer to this kind of issue was: "use on_init to deep_copy your data"

dinhlongviolin1:
Will be closed as this should be resolve as a doc issue. This might be too specific for only a single data type

PR #651, opened 2024-01-05
Fix - Freezing tests

trgiangdo:


toan-quach (approved):

jrobinAV (approved):
Oh good catch. I missed them.

That's good to isolate them, but are you sure these are the root cause of the freezing stuff?

trgiangdo:
Fixed by PR https://github.com/Avaiga/taipy/pull/654

PR #655, opened 2024-01-08
feature/#490 submit() functions return a Submission object

### LINKED ISSUE #490: Change submit APIs
The `submit` methods must evolve.

- [x] Accept custom properties parameters (!kwargs). The purpose is to pass some user information. The custom parameters should be stored within the Submission so the submitter can retrieve them.
- [x] Return the Submission object instead of the list of jobs. This is a BREAKING CHANGE. 
Since the breaking change is really minor we implement it and deliver a 3.1 release with the breaking change.
- [ ] Update corresponding enterprise APIs
- [ ] Updates the documentation set.
- [ ] Update release notes explicating the breaking change
- [ ] Update the migration documentation page

toan-quach:
Resolves #490.

# Changes

- Added custom properties parameters (**properties) to `tp.submit` so user can pass user information to the submission
- Added `properties` attribute to `Submission` to store user information on the submission
- **Change return type of `tp.submit` from `Job` or `List[Job` to `Submission` (BREAKING CHANGE)**
- **Made `Job` store callbacks for `on_status_change` upon creation (CRUCIAL CHANGES in `taipy/core/job/_job_manager.py`)**
- Fixed failed tests (alot!!!)


taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -134,7 +145,11 @@ def submit_task(
         else:
             if wait:
                 cls._wait_until_job_finished(job, timeout=timeout)
-        return job
+        return submission
+
+    @classmethod
+    def _update_submission_status(cls, job: Job):
+        cls._submission_entities[job.submit_id]._update_submission_status(job)
```
toan-quach, line 152: I decided to store `Submission`s in `cls._submission_entities` instead of getting it from the repo is because `Submission` has some in-memory state that are not stored in the repo (`__blocked_jobs`, `__pending_jobs`, `__running_jobs`). We can make changes to store them in the repo and replace `cls._submission_entities` with `SubmissionManagerFactory._build_manager()._get(job.submit_id)` but I think that's outside the scope of this ticket so I decided not to go forward with it.

taipy/core/job/_job_manager.py
```diff
@@ -50,9 +50,9 @@ def _create(
             force=force,
             version=version,
         )
-        cls._set(job)
         Notifier.publish(_make_event(job, EventOperation.CREATION))
         job._on_status_change(*callbacks)
+        cls._set(job)
```
toan-quach, line 55: Due to failed tests that are checking for `Job`'s `_subscribers` in `tests/core/_orchestrator/test_orchestrator__submit.py`. Previously these tests passed because the jobs are taken from memory which contains the callbacks. But because the submit function now returns a Submission, the jobs are taken from `Submission.jobs` -> jobs are taken from repos -> tests failed because these callbacks are stored in memory for the moment (and yet in repos). So I make the changes to make sure the callbacks are available both in memory and repos of all time

taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -38,6 +39,7 @@ class _Orchestrator(_AbstractOrchestrator):
     blocked_jobs: List = []
     lock = Lock()
     __logger = _TaipyLogger._get_logger()
+    _submission_entities: Dict[str, Submission] = {}
```
toan-quach, line 42: Please refer to my comment here https://github.com/Avaiga/taipy/pull/655/files#r1444309709 for why this attributes were created

taipy/core/job/_job_manager.py
gmarabout: What you say makes a lot of sense. Removing the "states" to use the repository instead should go into a dedicated work/PR.

FredLL-Avaiga (dismissed):
LGTM

FredLL-Avaiga:
I suppose I should I modify Core vis elements to conform with the new API.
In this PR or in another one ?

toan-quach:
@FredLL-Avaiga ah let me handle it together in this PR, since it's a breaking change, would want to solve all issues before closing :D

taipy/core/_orchestrator/_orchestrator.py
jrobinAV: This isn't good. 😢
It is proof that the in-memory state is dangerous and really error-prone. This is going to be a mess to maintain over time.  We may want to absorb this technical debt sooner than expected by storing the state.

Can you create a ticket about that with a high priority?

Thx.

taipy/core/job/_job_manager.py
jrobinAV: I wonder if a better implementation could be to pass `callbacks` variable to the `Job` constructor without calling `job._on_status_change(*callbacks)`. 
Something like that:
``` python
job = Job(
    id=JobId(f"{Job._ID_PREFIX}_{task.config_id}_{uuid.uuid4()}"),
    task=task,
    submit_id=submit_id,
    subscribers=callbacks
    submit_entity_id=submit_entity_id,
    force=force,
    version=version
)
cls._set(job)
Notifier.publish(_make_event(job, EventOperation.CREATION))
```

Does it make sense to you?

taipy/core/taipy.py
```diff
@@ -93,7 +95,7 @@ def is_submittable(entity: Union[Scenario, ScenarioId, Sequence, SequenceId, Tas
 def is_editable(
     entity: Union[
         DataNode, Task, Job, Sequence, Scenario, Cycle, DataNodeId, TaskId, JobId, SequenceId, ScenarioId, CycleId
-    ]
+    ],
```
jrobinAV: I believe that is not our convention. Can you check your local code formatting to be sure it matches our linter/pre commit tool?

taipy/core/taipy.py
```diff
@@ -71,6 +71,8 @@ def set(entity: Union[DataNode, Task, Sequence, Scenario, Cycle]):
         return _TaskManagerFactory._build_manager()._set(entity)
     if isinstance(entity, DataNode):
         return _DataManagerFactory._build_manager()._set(entity)
+    if isinstance(entity, Submission):
```
jrobinAV, line 83: The type Submission is not in the signature nor the doc.

taipy/core/taipy.py
```diff
@@ -120,7 +122,7 @@ def is_editable(
 def is_readable(
     entity: Union[
         DataNode, Task, Job, Sequence, Scenario, Cycle, DataNodeId, TaskId, JobId, SequenceId, ScenarioId, CycleId
-    ]
+    ],
```
jrobinAV: Same here.

taipy/core/taipy.py
```diff
@@ -165,19 +168,22 @@ def submit(
             in asynchronous mode.
         timeout (Union[float, int]): The optional maximum number of seconds to wait
             for the jobs to be finished before returning.
-
+        **properties (dict[str, any]): A keyworded variable length list of additional arguments.
```
jrobinAV: A blank line should be added before Returns keyword.

taipy/core/taipy.py
```diff
@@ -289,7 +295,7 @@ def get(entity_id: str) -> Union[Task, DataNode, Sequence, Scenario, Job, Cycle]
 
 
 def get(
-    entity_id: Union[TaskId, DataNodeId, SequenceId, ScenarioId, JobId, CycleId, str]
+    entity_id: Union[TaskId, DataNodeId, SequenceId, ScenarioId, JobId, CycleId, str],
```
jrobinAV: Same here. Please don't add comas before a parenthesis. I believe it is not standard formatting.

taipy/core/task/task.py
```diff
@@ -191,13 +190,13 @@ def submit(
                 mode.
             timeout (Union[float, int]): The maximum number of seconds to wait for the job to be finished before
                 returning.
-
+            **properties (dict[str, any]): A keyworded variable length list of additional arguments.
```
jrobinAV, line 193: A blank line before the Returns keyword.

tests/core/_orchestrator/test_orchestrator__submit.py
```diff
@@ -184,22 +189,22 @@ def test_submit_scenario_development_mode_blocked_jobs():
...
     # submission is created and correct
     assert len(_SubmissionManagerFactory._build_manager()._get_all()) == 1
-    submission = _SubmissionManagerFactory._build_manager()._get(job_1.submit_id)
     assert submission.submission_status == SubmissionStatus.BLOCKED
     assert submission.jobs == jobs
     assert submission.creation_date == s_time
     assert submission.entity_id == scenario.id
     assert submission.entity_type == "SCENARIO"
     assert submission.entity_config_id == "scenario_cfg"
+    assert submission.properties == {"no_of_retry": 10, "log": True, "log_file": "file_path"}
```
jrobinAV, line 210: Since you get the submission directly from the submit method, you don't get it from the Manager anymore. That means you don't test it is stored correctly.

```suggestion
# submission is stored
    submission = _SubmissionManagerFactory._build_manager()._get(job_1.submit_id)
    assert submission.submission_status == SubmissionStatus.BLOCKED
    assert submission.jobs == jobs
    assert submission.creation_date == s_time
    assert submission.entity_id == scenario.id
    assert submission.entity_type == "SCENARIO"
    assert submission.entity_config_id == "scenario_cfg"
    assert submission.properties == {"no_of_retry": 10, "log": True, "log_file": "file_path"}
```

What do you think? Maybe it could be a different test function.

jrobinAV (dismissed):

taipy/core/_orchestrator/_orchestrator.py
toan-quach: Agree, I think it's about time as well since we're implementing cluster mode. Can't delay it any further. Though I can't remember all the places that have this issues :/

taipy/core/job/_job_manager.py
toan-quach: it does, though I'd avoid doing that to prevent further complicating the PR. Would suggest to do it in a separate PR since we're not sure of the side effect of doing that.

tests/core/_orchestrator/test_orchestrator__submit.py
toan-quach: make perfect sense to me 😃

taipy/core/taipy.py
toan-quach: ah yep let me add that. I will also make Submission's `is_editable` to return True (after we have merged the expose submission api PR) as well since it now contains some user information. And we're adding set Submission to tp.set

tests/core/_orchestrator/test_orchestrator__submit.py
toan-quach: added retrieving submission from repo before asserting

taipy/core/_orchestrator/_orchestrator.py
toan-quach: @jrobinAV 
I have created the ticket here. Let me know what you think? https://github.com/Avaiga/taipy/issues/660

taipy/core/taipy.py
jrobinAV: Must be in the function signature as well. ;)

jrobinAV (dismissed):

taipy/core/taipy.py
```diff
@@ -238,19 +241,23 @@ def submit(
             in asynchronous mode.
         timeout (Union[float, int]): The optional maximum number of seconds to wait
             for the jobs to be finished before returning.
+        **properties (dict[str, any]): A keyworded variable length list of additional arguments.
```
FabienLelaquais: This information is not really useful. What happens with these 'additional arguments'?

taipy/core/task/task.py
FabienLelaquais: ... and explanation on properties would help. This doc doesn't help much.

FabienLelaquais (dismissed):
Looking great!

taipy/core/taipy.py
toan-quach: hm you're right, I don't have much with me atm. Can you suggest something? Would something like this work?
```suggestion
        **properties (dict[str, any]): A keyworded variable length list of additional arguments that will be stored within the `Submission^`.
```

FabienLelaquais: Yeah why not, if that's what's going on. May be a link to the API to retrieve that from the Submission object would be even better...

FredLL-Avaiga (dismissed):

taipy/core/taipy.py
toan-quach: you can retrieve it from the submission object :D

FabienLelaquais: Sure. That's my point.
Replacing the \`Submission^\` xref by \`Submission.get_properties()^\` or whatever.

taipy/core/job/job.py
```diff
@@ -58,7 +60,7 @@ class Job(_Entity, _Labeled):
     _MANAGER_NAME = "job"
     _ID_PREFIX = "JOB"
 
-    def __init__(self, id: JobId, task: Task, submit_id: str, submit_entity_id: str, force=False, version=None):
+    def __init__(self, id: JobId, task: "Task", submit_id: str, submit_entity_id: str, force=False, version=None):
```
gmarabout, line 63: I think we could eventually avoid that using `from __future__ import annotations`.

gmarabout (approved):

jrobinAV (approved):

taipy/core/job/job.py
toan-quach: I didn't know about that! Certainly sounds really useful! Maybe we can have a ticket to approach this later @jrobinAV

PR #665, opened 2024-01-10
✨ Add API to manually set job's stacktrace

gmarabout:
# Goals

The development of a "Taipy enterprise" specific feature requires to be able to manually set a job's stack trace.

# Changes
* Add a `stacktrace` setter on `Job` class
* Add unit test

### DIFF
--- a/taipy/core/job/job.py
+++ b/taipy/core/job/job.py
@@ -132,7 +132,7 @@ class Job(_Entity, _Labeled):
     def creation_date(self, val):
         self._creation_date = val
 
-    @property
+    @property  # type: ignore
     @_self_reload(_MANAGER_NAME)
     def stacktrace(self) -> List[str]:
         return self._stacktrace
--- a/taipy/core/job/job.py
+++ b/taipy/core/job/job.py
@@ -136,6 +136,11 @@ class Job(_Entity, _Labeled):
     def stacktrace(self) -> List[str]:
         return self._stacktrace
 
+    @stacktrace.setter  # type: ignore
+    @_self_setter(_MANAGER_NAME)
+    def stacktrace(self, val):
+        self._stacktrace = val
+
     @property
     def version(self):
         return self._version
--- a/taipy/core/job/job.py
+++ b/taipy/core/job/job.py
@@ -133,6 +133,7 @@ class Job(_Entity, _Labeled):
         self._creation_date = val
 
     @property
+    @_self_reload(_MANAGER_NAME)
     def stacktrace(self) -> List[str]:
         return self._stacktrace
 
--- a/tests/core/job/test_job.py
+++ b/tests/core/job/test_job.py
@@ -147,6 +147,19 @@ def test_status_job(task):
     assert job.is_skipped()
 
 
+def test_stacktrace_job(task):
+    submission = _SubmissionManagerFactory._build_manager()._create(task.id, task._ID_PREFIX, task.config_id)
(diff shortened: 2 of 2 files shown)


taipy/core/job/job.py
```diff
@@ -136,6 +136,11 @@ def creation_date(self, val):
     def stacktrace(self) -> List[str]:
```
toan-quach, line 137: I'm not sure of the reason why but if you want to make `stacktrace` auto set, you might want to add auto reload to its attribute also by adding `@_self_reload(_MANAGER_NAME)` after `@property` here

tests/core/job/test_job.py
```diff
@@ -147,6 +147,19 @@ def test_status_job(task):
...
+    submission = _SubmissionManagerFactory._build_manager()._create(task.id, task._ID_PREFIX, task.config_id)
+    job = Job("job_id", task, submission.id, "SCENARIO_scenario_config")
+
+    fake_stacktrace = """Traceback (most recent call last):
+File "<stdin>", line 1, in <module>
+ZeroDivisionError: division by zero""".splitlines()
+
+    job.stacktrace = fake_stacktrace
+    assert len(job.stacktrace) == 3
+    assert all(st == fake_stacktrace[idx] for idx, st in enumerate(job.stacktrace))
```
florian-vuillemot: Looks like evidence. Why don't you run a function?

toan-quach (dismissed):
lgtm but do consider adding sefl reload to it if it makes sense

tests/core/job/test_job.py
gmarabout: I don't understand... What is an evidence?

taipy/core/job/job.py
gmarabout: Can you explain what you mean by "auto set"? Why is it needed now? (notice I have not added this property which was already there).

toan-quach: the `@_self_setter(_MANAGER_NAME)` that you used will automatically save the current entity (here is a job) into the repo along with the recent changes. By doing that, the repo is always up-to-date. But if someone make an update to the repo, with the `_self_reload` decorator, you will just read from your memory the value of the attribute (stacktrace in this case), which has already been outdated. So the `self_reload` decorator ensures to reload the entity and extract the value in the repo instead of returning the current in-memory value. It's a feature we called "auto setter and auto reloader" -> auto set/save the changes to the repo and auto reload the value from the repo instead of the one in memory. Let me know if that doesn't make sense to you :D

tests/core/job/test_job.py
gmarabout: Ok, let's just simplify this code ;-)

taipy/core/job/job.py
gmarabout: Thanks for the explanation. Done!

toan-quach (approved):
lgtm!

tests/core/job/test_job.py
florian-vuillemot: What I mean is that you define a variable and check whether the variable is correctly set. Why not create a job that raised this error and then check if the stacktrace is correctly set?

gmarabout: That is the whole point of this PR: we need to set a stack trace that does not directly come from the execution of the job. In my case, the execution can/will be remote.

florian-vuillemot: oh my bad!

florian-vuillemot (approved):

PR #696, opened 2024-01-16
Display selector as list of radio buttons or check boxes

FredLL-Avaiga:
resolves #689

resolves #690

```py3
from taipy.gui import Gui, State

val = None

def on_change(sdtate: State, var: str, val: str):
  print(f"on_change(sdtate: State, var: {var}, val: {val})")


Gui("""
#test Selector Radio
<|{val}|selector|lov=Item 1;Item 2;Item 3|label=Selector|class_name=row fullwidth|hover_text=This is on the way|mode=radio|>
#test Selector Check
<|{val}|selector|lov=Item 1;Item 2;Item 3|label=Selector|class_name=row fullwidth|hover_text=This is on the way|mode=check|>
""").run()

```

PR #697, opened 2024-01-16
Make job and submission editable

jrobinAV:
A behavior change was introduced previously to make jobs and submissions not editable.
We want to keep these entities editable.

related #696 
The PR 696 should be merged first to fix some linter issues.

### DIFF
--- a/taipy/core/job/_job_manager.py
+++ b/taipy/core/job/_job_manager.py
@@ -89,10 +89,6 @@ class _JobManager(_Manager[Job], _VersionMixin):
     def _is_deletable(cls, job: Union[Job, JobId]) -> bool:
         if isinstance(job, str):
             job = cls._get(job)
-        if job.is_finished():
-            return True
-        return False
-
-    @classmethod
-    def _is_editable(cls, entity: Union[Job, str]) -> bool:
-        return False
+        if not job.is_finished():
+            return False
+        return True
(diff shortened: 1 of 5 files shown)


taipy/core/job/_job_manager.py
```diff
@@ -86,13 +86,9 @@ def _get_latest(cls, task: Task) -> Optional[Job]:
             return max(jobs_of_task)
 
     @classmethod
-    def _is_deletable(cls, job: Union[Job, JobId]) -> bool:
+    def _is_deletable(cls, job: Union[Job, JobId]) -> bool:  # type: ignore
```
toan-quach: why do we need `# type: ignore` here?

toan-quach (dismissed):
lgtm, just minor question regarding type hint

gmarabout (dismissed):
LGTM

taipy/core/job/_job_manager.py
jrobinAV: We don't.

jrobinAV:
```suggestion
    def _is_deletable(cls, job: Union[Job, JobId]) -> bool:
```

jrobinAV: I removed it.

toan-quach (approved):

PR #711, opened 2024-01-17
🐛 Fix orchestrator being initialised at each call

gmarabout:
# Context

In Taipy Enterprise, the orchestrator implementation is using state to try to recover from a crash.
This works only if the orchestrator is a singleton and is not initialised every time we call `_build_orchestrator()`.

# Changes
* Add a check in the "orchestrator factory" to return a previously built orchestrator

# Important note
The `_OrchestratorFactory._build_orchestrator()` is not thread safe and should be called only from the main thread.

### DIFF
--- a/taipy/core/_orchestrator/_orchestrator_factory.py
+++ b/taipy/core/_orchestrator/_orchestrator_factory.py
@@ -32,6 +32,8 @@ class _OrchestratorFactory:
 
     @classmethod
     def _build_orchestrator(cls) -> Type[_AbstractOrchestrator]:
+        if cls._orchestrator:
+            return cls._orchestrator
         if util.find_spec(cls._TAIPY_ENTERPRISE_MODULE) is not None:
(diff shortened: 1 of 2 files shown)


jrobinAV (dismissed):

taipy/core/_orchestrator/_orchestrator_factory.py
```diff
@@ -32,6 +32,8 @@ class _OrchestratorFactory:
 
     @classmethod
     def _build_orchestrator(cls) -> Type[_AbstractOrchestrator]:
+        if cls._orchestrator:
```
toan-quach, line 35: somehow it feels like we had this before, dejavu I guess haha

toan-quach (dismissed):
lgtm!!! thanks for fixing this issue!

joaoandre-avaiga (dismissed):

toan-quach (approved):

PR #715, opened 2024-01-18
prevent non changed layout to generate unnecessary renders

### LINKED ISSUE #710: Need to get several indexes when doing a box/ lasso selection from a plotly graph
**Description**
When using the new feature to integrate plotly charts, we do get the index of the selected point on the chart.
However, when performing a box or lasso selection, the selected_state is empty instead of returning a list of indexes corresponding
to the selected points.

**How to reproduce**
You can check using the following code below

- A code fragment
    ```
   from taipy.gui import Gui 
  import plotly.express as px

  df = px.data.gapminder()
  df_first = df[df['year']==2007].reset_index(drop=True)
  bubble_fig = px.scatter(df_first, x="gdpPercap", y="lifeExp",size="pop", hover_name="country", log_x=True, size_max=60)

  selected_state = []

  def on_change(state,var,val):
     print("val = ", val)
     if var == 'selected_state':
        print(state.selected_state)
        print("selected state", val, "row:", state.df_first.iloc[val])
 
   md = """
   <|chart|figure={bubble_fig}|selected={selected_state}|>
      ```
  Gui(md).run(port=4563)

**Expected behavior**
selected_state variable shouldn't be empty, it should contain a list of indexes corresponding to the selected points.

**Runtime environment**
Taipy develop branch
Windows 11

FredLL-Avaiga:
resolves #710

### DIFF
--- a/frontend/taipy-gui/src/components/Taipy/Chart.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Chart.tsx
@@ -530,10 +530,10 @@ const Chart = (props: ChartProp) => {
         [data, dataKey, props.figure]
     );
 
-    const selectHandler = useCallback(
-        (points?: PlotDatum[]) => {
+    const onSelect = useCallback(
+        (evt?: PlotSelectionEvent) => {
             if (updateVars) {
-                const traces = (points || []).reduce((tr, pt) => {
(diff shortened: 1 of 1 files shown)


dinhlongviolin1 (dismissed):
Lgtm

FabienLelaquais (approved):
👍👍

PR #725, opened 2024-01-18
Property to not show the creation dialog

### LINKED ISSUE #214: Data Node Selector - Filter on selected scenario
## New Feature / Improvement

Filtering the Data Node selector by scenario would allow the user to see Data Nodes of the selected scenario without all the others. It can be really useful in an application. It would also create a "bridge" between the Scenario selector and Data Node selector.

Proposed syntax, for example:

`<|{data_node}|data_node_selector|scenario={scenario}|>`

This would show only Data nodes corresponding to the selected scenario ('scenario').

Taipy version:
Taipy 2.3.0

### LINKED ISSUE #280: Scenario Selector - optionally do not show form after clicking 'Add Scenario'
# Feature Improvement

Possibility to remove the form after clicking 'Add Scenario'. The scenario creation would be done with default values or through the `on_creation` callback.

# Environment

Taipy Develop 3.0

FredLL-Avaiga:
Scenario to filter list in datanode_selector.

Resolves #280
Partly resolves #214.

### DIFF
--- a/frontend/taipy/src/ScenarioSelector.tsx
+++ b/frontend/taipy/src/ScenarioSelector.tsx
@@ -84,6 +84,7 @@ interface ScenarioSelectorProps {
     className?: string;
     dynamicClassName?: string;
     showPins?: boolean;
+    showDialog?: boolean;
 }
 
 interface ScenarioEditDialogProps {
@@ -277,7 +278,7 @@ const ScenarioEditDialog = ({ scenario, submit, open, actionEdit, configs, close
                                         <DatePicker
                                             label="Date"
                                             value={new Date(form.values.date)}
-                                            onChange={(date) => form.setFieldValue("date", date?.toISOString())}
+                                            onChange={(date?:Date|null) => form.setFieldValue("date", date?.toISOString())}
                                             disabled={actionEdit}
                                         />
                                     </LocalizationProvider>
@@ -410,7 +411,7 @@ const ScenarioEditDialog = ({ scenario, submit, open, actionEdit, configs, close
 };
 
 const ScenarioSelector = (props: ScenarioSelectorProps) => {
-    const { showAddButton = true, propagate = true, showPins = false } = props;
+    const { showAddButton = true, propagate = true, showPins = false, showDialog = true } = props;
     const [open, setOpen] = useState(false);
     const [actionEdit, setActionEdit] = useState<boolean>(false);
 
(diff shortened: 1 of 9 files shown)


frontend/taipy/src/ScenarioSelector.tsx
```diff
@@ -277,7 +278,7 @@ const ScenarioEditDialog = ({ scenario, submit, open, actionEdit, configs, close
                                         <DatePicker
                                             label="Date"
                                             value={new Date(form.values.date)}
-                                            onChange={(date) => form.setFieldValue("date", date?.toISOString())}
+                                            onChange={(date?:Date|null) => form.setFieldValue("date", date?.toISOString())}
```
FabienLelaquais, line 281: Do we need the |null when ? is used? just asking... seems strange

taipy/gui_core/_context.py
```diff
@@ -361,6 +361,16 @@ def crud_scenario(self, state: State, id: str, payload: t.Dict[str, str]):  # no
                         return
                 elif on_creation is not None:
                     _warn(f"on_creation(): '{on_creation}' is not a function.")
+                elif not with_dialog:
+                    if len(Config.scenarios) == 2:
+                        scenario_config = [sc for k, sc in Config.scenarios.items() if k != "default"][0]
+                    else:
+                        state.assign(
+                            _GuiCoreContext._SCENARIO_SELECTOR_ERROR_VAR,
+                            f"Error creating Scenario: need only one scenario config ({len(Config.scenarios)}).",
```
FabienLelaquais:
```suggestion
                            f"Error creating Scenario: only one scenario config needed ({len(Config.scenarios)} found).",
```

frontend/taipy/src/ScenarioSelector.tsx
```diff
@@ -466,6 +444,33 @@ const ScenarioSelector = (props: ScenarioSelectorProps) => {
...
+    const onDialogOpen = useCallback(() => {
+        if (showDialog) {
+            setOpen(true);
+            setActionEdit(false);
+        } else {
+            onSubmit(false, false, {}, false);
+        }
+    }, [showDialog, onSubmit]);
+
+    const onDialogClose = useCallback(() => {
```
FabienLelaquais, line 456: Does this need to be a hook?

FabienLelaquais (dismissed):
👌

frontend/taipy/src/ScenarioSelector.tsx
FredLL-Avaiga: ? adds undefined
but the signature needs null too (when you do not select a date)

frontend/taipy/src/ScenarioSelector.tsx
FredLL-Avaiga: yes because it depends on setState actions that are only available in the component

FabienLelaquais (approved):
👍

PR #767, opened 2024-01-26
Add GUI specific tests

dinhlongviolin1:


FabienLelaquais (dismissed):
So far, so good...

.github/actions/gui-test/prefix/action.yml
```diff
@@ -0,0 +1,14 @@
...
+    - name: Build the Docker image
+      shell: bash
+      working-directory: ./tools/gui/docker/nginx-prefix
+      run: docker build . --file Dockerfile --tag taipy-nginx-proxy
+    - name: run nginx reverse proxy docker container
+      shell: bash
+      run: docker run -d -p 4000:4000 taipy-nginx-proxy
+    - name: run pytest with prefix configuration
+      shell: bash
+      run: pipenv run pytest -m "teste2e" tests/gui/e2e --e2e-base-url="/prefix/" --e2e-port="4000"
```
florian-vuillemot, line 14: It may be clearer to create a `docker-compose` file running the `pipenv` command and `nginx`. 
You can use the `--abort-on-container-exit` and `--exit-code-from` arguments to let the `docker-compose` exit correctly.

florian-vuillemot (approved):
I don't see the current reason to create several GitHub `actions` when you're not reusing them. Putting everything in the `gui-test` action is simpler IMHO. But that's a matter of taste.

jrobinAV (approved):

.github/actions/gui-test/prefix/action.yml
dinhlongviolin1: Thanks for the suggestion. If I am not mistaken, docker compose can only run in an isolated container. Meaning that if I want to run pipenv I would need to setup another pipenv environment. Docker compose would work for me if there is no previous environment setup. But since we already have a pipenv environment, setting up NGINX for the proxy would be my only usage for containerization.

florian-vuillemot: That's a good reason!

FredLL-Avaiga (approved):

FabienLelaquais (approved):
👍

PR #779, opened 2024-01-30
Split .data folder

joaoandre-avaiga:


taipy/core/_entity/_migrate_cli.py
trgiangdo, line 1: Do we need to add a new entry in the migrated Config?
There was only "storage_folder" before, do we need to add the "tp_storage_folder"

taipy/core/config/core_section.py
```diff
@@ -28,9 +28,12 @@ class CoreSection(UniqueSection):
 
     Attributes:
         root_folder (str): Path of the base folder for the taipy application. The default value is "./taipy/"
-        storage_folder (str): Folder name used to store Taipy data. The default value is ".data/". It is used in
+        storage_folder (str): Folder name used to store user data. The default value is ".user_data/". It is used in
```
trgiangdo: How about "data_folder"?

taipy/core/config/core_section.py
```diff
@@ -28,9 +28,12 @@ class CoreSection(UniqueSection):
...
             conjunction with the *root_folder* attribute. That means the storage path is <root_folder><storage_folder>
-            (The default path is "./taipy/.data/").
+            (The default path is "./taipy/.user_data/").
+        tp_storage_folder (str): Folder name used to store Taipy data. The default value is ".taipy_data/". It is used
```
trgiangdo: How about "entities_folder"?

taipy/core/config/core_section.py
```diff
@@ -28,9 +28,12 @@ class CoreSection(UniqueSection):
...
+            in conjunction with the *root_folder* attribute. That means the storage path is
+            <root_folder><storage_folder> (The default path is "./taipy/.taipy_data/").
```
trgiangdo: Are we sure about this? I don't think the `.taipy` folder ever exists

taipy/core/data/csv.py
```diff
@@ -122,12 +122,13 @@ def __init__(
             **properties,
         )
         self._path = properties.get(self.__PATH_KEY, properties.get(self.__DEFAULT_PATH_KEY))
+        if self._path and ".data" in self._path:
```
florian-vuillemot, line 127: `.data`?
Can we at least put it as a const?

joaoandre-avaiga: This is only to migrate old taipy projects to the new structure. If .data is on the datanode current path, we move the file to the new folder. I can put it in a constant.

taipy/core/config/core_section.py
joaoandre-avaiga: Im not sure. I kept the current text, just replaced .data with the new folder name. Maybe @jrobinAV can verify if this docstring is correct.

taipy/core/config/core_section.py
jrobinAV: Whatever the name of the config attribute, I would prefix it with taipy so it is explicit that it is for internal usage. Actually I would use `taipy_folder` directly, but I am ok with both `taipy_storage_folder` and `taipy_entities_folder`.
What do you think?

For the default value, I would go for `.taipy` instead of `.taipy_data`

taipy/core/config/core_section.py
jrobinAV: This folder is the entrypoint folder for a user to get and drop data. I prefer `data_folder` instead of `storage_folder`, to be honest, but then it will be a breaking change. Am I correct?
What do you think?

For the default value, I have two remarks:
1. Why not make the default value not start with a `.`. In the end, it is made to be accessed by the user. It is not a hidden folder.
2. Why prefix it by user? Why not just `data`?

taipy/core/config/core_section.py
jrobinAV: `<storage_folder>` should be renamed by `<taipy_storage_folder>` (or whatever the name of the attribute we chose).
```suggestion
            <root_folder><taipy_storage_folder> (The default path is "./taipy/.taipy/").
```

taipy/core/config/core_section.py
```diff
@@ -324,9 +349,12 @@ def _configure(
...
                 The default value is "./taipy/"
-            storage_folder (Optional[str]): Folder name used to store Taipy data. The default value is ".data/".
-                It is used in conjunction with the `root_folder` field. That means the storage path is
-                <root_folder><storage_folder> (The default path is "./taipy/.data/").
+            storage_folder (str): Folder name used to store user data. The default value is ".user_data/". It is used in
+                conjunction with the *root_folder* attribute. That means the storage path is
+                <root_folder><storage_folder> (The default path is "./taipy/.user_data/").
+            tp_storage_folder (str): Folder name used to store Taipy data. The default value is ".taipy_data/". It is
+                used in conjunction with the *root_folder* attribute. That means the storage path is
+                <root_folder><storage_folder> (The default path is "./taipy/.taipy_data/").
```
jrobinAV: You mean <tp_storage_folder>.

jrobinAV (commented):
We need to converge on the config attribute names and their default values.

Otherwise, I believe it is good.

taipy/core/config/core_section.py
joaoandre-avaiga: > This folder is the entrypoint folder for a user to get and drop data. I prefer data_folder instead of storage_folder, to be honest, but then it will be a breaking change. Am I correct?

Yes, I kept the config name the same to avoid a breaking change. My propose is to keep storage folder for the user data, so it doesn't have impact right now. We could change the name in 4.0, in the future.

> Why not make the default value not start with a .. In the end, it is made to be accessed by the user. It is not a hidden folder.

I guess theres no reason to keep it hidden.

>Why prefix it by user? Why not just data?

I think data may be a little to generic and could conflict with other things the user have. We do delete stuff from the data folder in some cases, so it may be more secure if it's not a common name.

jrobinAV (approved):

PR #785, opened 2024-01-31
Shutdown ProcesspoolExecutor when calling Core().stop()

jrobinAV:
After stoping the standalone job dispatcher run loop, we now shutdown all the ProcessPoolExecutor.

### DIFF
--- a/taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
@@ -52,3 +52,7 @@ class _StandaloneJobDispatcher(_JobDispatcher):
     def _update_job_status_from_future(self, job: Job, ft):
         self._pop_dispatched_process(job.id)  # type: ignore
         self._update_job_status(job, ft.result())
+
+    def stop(self):
+        self._executor.shutdown(wait=True, cancel_futures=True)
+        super().stop()
--- a/taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
@@ -8,7 +8,7 @@
 # Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
-
+import sys
 from concurrent.futures import Executor, ProcessPoolExecutor
 from functools import partial
 from typing import Callable, Optional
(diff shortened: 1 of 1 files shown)


gmarabout (dismissed):
Looks good to me.

trgiangdo (dismissed):
Oh the trick is to shutdown after stop the thread. Great catch!

@gmarabout Does this help resolve the problem on your side?
Because the test is passing consistently on my side with this fix now

taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
```diff
@@ -52,3 +52,7 @@ def _release_worker(self, _):
     def _update_job_status_from_future(self, job: Job, ft):
         self._pop_dispatched_process(job.id)  # type: ignore
         self._update_job_status(job, ft.result())
+
+    def stop(self):
+        super().stop()
+        self._executor.shutdown(wait=True, cancel_futures=True)
```
trgiangdo:
```suggestion
        self._executor.shutdown(wait=True)
```

Python 3.8 does not have the `cancel_futures` parameter. I tried to remove it and it's still working consistently.

florian-vuillemot: Did a test catch it?

gmarabout:
> Oh the trick is to shutdown after stop the thread. Great catch!
> 
> @gmarabout Does this help resolve the problem on your side? Because the test is passing consistently on my side with this fix now

Unfortunately no. Unit tests are still failing on my side. 
But I guess this modification should be added anyway?

gmarabout (approved):

taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
jrobinAV: yep. Many, actually.

gmarabout:
> > Oh the trick is to shutdown after stop the thread. Great catch!
> > @gmarabout Does this help resolve the problem on your side? Because the test is passing consistently on my side with this fix now
> 
> Unfortunately no. Unit tests are still failing on my side. But I guess this modification should be added anyway?

My bad! It IS fixing the unit tests!

taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
jrobinAV: Yes a lot of tests are failing on Python 3.8. I just pushed a fix.

florian-vuillemot (approved):

florian-vuillemot:
Should we create a dedicated function that checks the python version? I'm afraid this kind of "if" will be forgotten when we drop support for python 3.8.

toan-quach (approved):

trgiangdo (approved):

PR #800, opened 2024-02-06
Added Step Property to the Slider component

### LINKED ISSUE #639: Make a slider in float type with a particular step value
I'm trying to make a slider in float type with a particular step value. 
How to make it?

I think it will be easier to use if there's a "step" property: 

```
pressure = 0.05
with tgb.Page() as page:
    tgb.slider("{pressure}", min=0.01, max=0.1, step=0.01)
```

Satoshi-Sh:
## Description 
Added a step property to the Slider component so that users can 

I need some help here
- How can I use the same setting of formatting? I just turned off formatting for the moment. 
- I created the test function, but I got an assertion error. Not sure how to fix it.

## Related Issue 
resolves #639


frontend/taipy-gui/src/components/Taipy/Slider.tsx
```diff
@@ -68,6 +69,7 @@ const Slider = (props: SliderProps) => {
 
     const min = lovList.length ? 0 : props.min;
     const max = lovList.length ? lovList.length - 1 : props.max;
+    const step = props.step === undefined ? 1 : props.step;
```
FredLL-Avaiga: The default value can be set in the expansion of props (L<55)

frontend/taipy-gui/src/components/Taipy/Slider.tsx
```diff
@@ -30,6 +30,7 @@ interface SliderProps extends LovProps<number | string | number[] | string[], nu
     height?: string;
     min?: number;
     max?: number;
+    step?: number;
```
FredLL-Avaiga, line 33: the property needs to be declared on the back end too
See https://vscode.dev/github/Avaiga/taipy/blob/develop/taipy/gui/_renderers/factory.py#L428
Something like `("step", PropertyType.number, 1)` should do

tests/gui/builder/control/test_slider.py
```diff
@@ -36,6 +36,20 @@ def test_slider_with_min_max_builder(gui: Gui, test_client, helpers):
...
 
+def test_slider_with_step_builder(gui: Gui, test_client, helpers):
+    gui._bind_var_val("x", 0)
+    with tgb.Page(frame=None) as page:
+        tgb.slider(value="{x}", min=-10, max=10, step=2)
+    expected_list = [
+        "<Slider",
+        "min={-10.0}",
+        "max={10.0}",
+        "step={2.0}",
```
FredLL-Avaiga, line 47: the update to factory.py should fix this
PS: You also need to add jsx test ;-)

FredLL-Avaiga (changes requested):
need backend updates too
(and modification to viselement.json)

tests/gui/builder/control/test_slider.py
Satoshi-Sh: Thanks for guiding, @FredLL-Avaiga . 

I fixed what you commented on. I'm not sure about jsx test. Do I just need to add a similar function to the `tests/gui/control/test_slider.py` or something else?

FredLL-Avaiga: you can add a test in Slider.spec.tsx if it makes sense
PS: I've updated your branch

Satoshi-Sh: Yeah, it makes sense. I'm unfamiliar with testing but I will have a look at it later and let you know how it goes.

Satoshi-Sh: I'm not sure if the test is good for the step. I wanted to trigger one step increment/decrement in a different way like (keydown ">"), but I couldn't find the way to do so. 

@FredLL-Avaiga

FredLL-Avaiga: testing the behavior of the component needs to be done in the Slider.spec.tsx 
That's where you can create user interaction (userEvent.type() or userEvent.click())

Satoshi-Sh: I added a test to `Slider.spec.tsx`,  and used fireEvent for that. 

I tried to trigger slider one increment/decrement by function but couldn't find the way so. I just did like
```
const step = parseFloat(elt.step);
fireEvent.change(elt, {target:{value: 33 + step }});
```

FredLL-Avaiga: Can you try `await user.keyboard('[ArrowRight]')` after `userEvent.click(<the right element>)` 
And then check the value ?

FabienLelaquais:
This PR is looking better and better. Thanks @Satoshi-Sh and @FredLL-Avaiga.
Note that the missing piece before finishing this will include the property documentation.

tests/gui/builder/control/test_slider.py
Satoshi-Sh:
```
it("changes by set step", async()=>{
        const dispatch = jest.fn();
        const state: TaipyState = INITIAL_STATE;
        const { getByDisplayValue } = render(
            <TaipyContext.Provider value={{ state, dispatch }}>
                <Slider value={33} min={1} max={50} step={0.1} />
            </TaipyContext.Provider>
        );
        const elt = getByDisplayValue("33") as HTMLInputElement;
        await userEvent.click(elt);
        await userEvent.keyboard('[ArrowRight]')
        expect(elt.value).toBe("33.1");
        });
```
I tried like this. The click itself changes the value to 1 (min value) from 33, but the keyboard doesn't change the value at all.

Could you also tell me where is the property documentation for the slider?

FredLL-Avaiga: You don't need the TaipyContext.Provider for this test

You'll have to find the right element to focus before using keyboard :-(

But I suppose the backend test (.py) is enough as we don't need to test MUI ...
So you may as well forget this frontend test alltogether

Documentation is in taipy/gui/viselements.json

Satoshi-Sh: Sounds good to me. I just removed the change on the Slider.spec.tsx and added a property explanation to the viselements.json.

Satoshi-Sh: @FredLL-Avaiga  Is it fine to make this PR ready for review? 

Still, it fails some tests though.

FredLL-Avaiga: Yes I think so

FredLL-Avaiga (dismissed):
🥇

taipy/gui/viselements.json
```diff
@@ -150,6 +150,12 @@
             "default_value": "100",
             "doc": "The maximum value.<br/>This is ignored when <i>lov</i> is defined."
           },
+          {
+            "name": "step",
+            "type": "int|float",
+            "default_value": "1",
+            "doc": "The step value.<br/>This is ignored when <i>lov</i> is defined."
```
```suggestion
            "doc": "The step value: the gap between two consecutive values the slider set. It is a good practice to have (<i>max</i>-<i>min</i>) being divisible by <i>step</i>.<br/>This property is ignored when <i>lov</i> is defined."
```

FabienLelaquais (dismissed):

FabienLelaquais (approved):
👍

Satoshi-Sh:
Thanks for the guidance, @FredLL-Avaiga :tada:

FredLL-Avaiga:
> Thanks for the guidance, @FredLL-Avaiga :tada:  

Thanks for your contribution

PR #801, opened 2024-02-07
Fix - Add wait when stop dispatcher

trgiangdo:
In some Core tests that use the job dispatcher, sometimes when the job dispatcher thread of the previous test is not fully terminated yet, and the next test is already started, the result of the second tests can be nondeterministic.

This PR adds the wait mechanism to the `Dispatcher.stop()` method to address this issue.

In this PR:
- Add `wait` and `timeout` parameters to dispatcher.stop() method. When the `wait` parameter is True then the method will wait for the thread to stop. The `timeout` is None to make sure that the method waits until the thread is fully terminated when the dispatcher is stopped.
- Remove some duplicated lines in tests that are no longer necessary


jrobinAV (dismissed):

jrobinAV (dismissed):

gmarabout (dismissed):

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
```diff
@@ -47,9 +47,16 @@ def is_running(self) -> bool:
...
+    def stop(self, wait: bool = True, timeout: Optional[float] = None):
+        """Stop the dispatcher.
+
+        Parameters:
+            wait (bool): If True, the method will wait for the dispatcher to stop.
+            timeout (Optional[float]): The maximum time to wait. If None, the method will wait indefinitely.
+        """
         self._STOP_FLAG = True
+        if wait and self.is_alive():
+            self.join(timeout=timeout)
```
florian-vuillemot: Should we add a debug log? If `timeout is None`, it may be difficult to debug.

taipy/core/_orchestrator/_orchestrator_factory.py
```diff
@@ -61,9 +61,9 @@ def _build_dispatcher(cls, force_restart=False) -> Optional[_JobDispatcher]:
         return cls._dispatcher
 
     @classmethod
-    def _remove_dispatcher(cls) -> Optional[_JobDispatcher]:
+    def _remove_dispatcher(cls, wait: bool = True, timeout: Optional[float] = None) -> Optional[_JobDispatcher]:
```
florian-vuillemot: This function always returns None.
```suggestion
    def _remove_dispatcher(cls, wait: bool = True, timeout: Optional[float] = None) -> None:
```

florian-vuillemot (commented):
I like it when a PR removes more code than it adds 👌

taipy/core/_orchestrator/_orchestrator_factory.py
jrobinAV: Good catch

toan-quach (dismissed):
lgtm

jrobinAV (dismissed):

jrobinAV (approved):

PR #806, opened 2024-02-08
feat: support for pandas 1.3

joaoandre-avaiga:


taipy/core/data/_abstract_sql.py
```diff
@@ -226,9 +226,14 @@ def _read_as_pandas_dataframe(
         join_operator=JoinOperator.AND,
     ):
         with self._get_engine().connect() as conn:
+            result = conn.execute(text(self._get_read_query(operators, join_operator)))
+
+            # On pandas 1.3.5 there's a bug that makes that the dataframe from sqlalchemy query is
+            # created without headers
+            keys = [col for col in result.keys()]
```
florian-vuillemot, line 233: Isn't that enough?
```suggestion
            keys = result.keys()
```
Or:
```suggestion
            keys = list(result.keys())
```

florian-vuillemot (commented):
Wonderful ! But do we test in the CI with pandas == 1.3.5? Should we add a dedicated test or Pipefile?

Pipfile
```diff
@@ -21,7 +21,7 @@ markdown = "==3.4.4"
 marshmallow = "==3.20.1"
 networkx = "==2.6"
 openpyxl = "==3.1.2"
-pandas = ">=2.0.2,<2.3.0"
+pandas = ">=1.3.5,<2.3.0"
```
jrobinAV, line 24: I thought the convention was to use Pipfile as the min version supported by Taipy for each dependency. So that means we should not have any range here, just the min version.
That is the role of the setup.py to define a range.

taipy/core/Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-pandas = ">=2.0.2,<2.3.0"
+pandas = ">=1.3.5,<2.3.0"
```
jrobinAV, line 7: Same remark on the min version vs range.

taipy/gui/Pipfile
```diff
@@ -12,7 +12,7 @@ gevent = "==23.7.0"
 gevent-websocket = "==0.10.1"
 kthread = "==0.2.3"
 markdown = "==3.4.4"
-pandas = ">=2.0.2,<2.3.0"
+pandas = ">=1.3.5,<2.3.0"
```
jrobinAV, line 15: Same remark on the min version versus a range of versions.

jrobinAV (commented):
That's great.

1. @florian-vuillemot Why do we have a range of versions in the pipfile for pandas?

2. I would also implement the append method so we are clean in terms of compatibility and documentation is easier. I would do it in another PR.

jrobinAV (approved):

taipy/core/data/_abstract_sql.py
joaoandre-avaiga: Unfortunately result.keys returns an sqlalchemy object that does support list casting, thats why I've used a list comprehension.

PR #828, opened 2024-02-13
input update change on enter only when change_delay=-1

### LINKED ISSUE #610: Input visual element has a frustrating delay
The input visual element will register the inputted text at a specific time delay (defined by the property change_delay). When this register happens, the app will have a quick lag that will cause typed text not to be registered. This is very frustrating: The user is typing something, stops for half a second, and then continues typing, and some of his new characters are missing.

You can test this issue on [TalkToTaipy](https://talk-to-taipy.taipy.cloud/) or [TaipyGPT](https://demo-llm-chat.taipy.cloud/). Just type at different speeds while taking quick pauses; some characters won't be registered at some point, which will be frustrating. You can also try it on mobile; the change_delay seems to change between local, Cloud, and Mobile.

**Discussion**

An example such the Chat situation obviously raises the problem of the continuous update of the text value. Only the final value makes sense.

We may want to force the user to press Enter when he's done. Setting change_delay to -1 may be the solution.

Another track is to figure out, with a ping on the connection initialization, the bandwidth and adapt the delays accordingly.

**Expected change**
The text should only be registered regularly if the user defines an on_change function. In my two use cases, I only need the user's text when he presses "Enter" (on_action). I do not need it registered regularly, so I did not define on_change.

FredLL-Avaiga:
resolves #610

`<|{val}|input|change_delay=-1|>`


dinhlongviolin1 (dismissed):
Lgtm

taipy/gui/viselements.json
```diff
@@ -1448,7 +1448,7 @@
             "name": "change_delay",
             "type": "int",
             "default_value": "<i>App config</i>",
-            "doc": "Minimum time between triggering two calls to the <i>on_change</i> callback.<br/>The default value is defined at the application configuration level by the <strong>change_delay</strong> configuration option. if None, the delay is set to 300 ms."
+            "doc": "Minimum time between triggering two calls to the <i>on_change</i> callback.<br/>The default value is defined at the application configuration level by the <strong>change_delay</strong> configuration option. if None, the delay is set to 300 ms. if -1, the input change is triggered only on Enter"
```
```suggestion
            "doc": "Minimum time between triggering two calls to the <i>on_change</i> callback.<br/>The default value is defined at the application configuration level by the <strong>change_delay</strong> configuration option. if None, the delay is set to 300 ms.<br/>If set to -1, the input change is triggered only when the user presses the Enter key."
```

FabienLelaquais (dismissed):
👍

FabienLelaquais (approved):
👌

PR #836, opened 2024-02-15
💩 Add flush/fsync to force file synchronization

gmarabout:
EXPERIMENTAL
DON'T MERGE!

### DIFF
--- a/.github/workflows/overall-tests.yml
+++ b/.github/workflows/overall-tests.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.11']
+        python-version: ['3.10', '3.11']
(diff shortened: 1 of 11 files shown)


.github/workflows/overall-tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     defaults:
       run:
-        working-directory: ./app/${{ matrix.os }}/${{ matrix.python-version }}
+        working-directory: ${{ matrix.os }}/${{ matrix.python-version }}
```
jrobinAV: Nice!

gmarabout: Yes... but that does not work 🤣

PR #861, opened 2024-02-21
release worker first in standalone mode

jrobinAV:
This small change fixes the performance issue.

### DIFF
--- a/taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
@@ -10,6 +10,7 @@
 # specific language governing permissions and limitations under the License.
 
 import threading
+import time
 from abc import abstractmethod
 from queue import Empty
 from typing import Dict, Optional
@@ -69,6 +70,8 @@ class _JobDispatcher(threading.Thread):
                             break
                         job = self.orchestrator.jobs_to_run.get(block=True, timeout=0.1)
                     self._execute_job(job)
+                else:
+                    time.sleep(0.1)  # We need to sleep to avoid busy waiting.
             except Empty:  # In case the last job of the queue has been removed.
                 pass
             except Exception as e:
--- a/taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
@@ -51,8 +51,9 @@ class _StandaloneJobDispatcher(_JobDispatcher):
         future = self._executor.submit(_TaskFunctionWrapper(job.id, job.task), config_as_string=config_as_string)
 
         self._set_dispatched_processes(job.id, future)  # type: ignore
+        future.add_done_callback(self._release_worker)  # We must release the worker before updating the job status
+        # so that the worker is available for another job as soon as possible.
         future.add_done_callback(partial(self._update_job_status_from_future, job))
-        future.add_done_callback(self._release_worker)
 
     def _release_worker(self, _):
         self._nb_available_workers += 1


FredLL-Avaiga (dismissed):
Well done!

FabienLelaquais (dismissed):
👏👏👏

florian-vuillemot (dismissed):

joaoandre-avaiga (dismissed):

toan-quach (dismissed):

gmarabout (dismissed):
Well done!!!

jrobinAV:
OK, after some more investigations, I better understand the issue.
The first fix (run release_worker callback before update_status) is not a correct solution; it is just a side effect that makes it work. The thing is that the real problem only appears when all the workers are occupied, and releasing the worker first hides the problem.

### Here is an explanation:

We have two callbacks to run after a job is completed.
1. `release_worker`: Releases the worker so it can be used to run another job.
2. `update_job_status`: Change the current job status to fail or completed, and update the subsequent job statuses to unblock them. This results in populating the queue of jobs to run with the next job.

Basically, the two callbacks must be completed to start the next job, as you can see in the following piece of code.

``` python
    def run(self):
        from time import sleep
        _TaipyLogger._get_logger().info("Start job dispatcher...")
        while not self._STOP_FLAG:
            try:
                if self._can_execute():
                    with self.lock:
                        job = self.orchestrator.jobs_to_run.get(block=True, timeout=0.1)
                    self._execute_job(job)
            except Exception:  # In case the last job of the queue has been removed.
                pass

    def _can_execute(self) -> bool:
        """Returns True if the dispatcher have resources to execute a new job."""
        return self._nb_available_workers > 0
```
The first callback, `release_worker,` makes the `_can_execute` method return `True` when the worker is released. The second callback, `update_job_status`, populates the `self.orchestrator.jobs_to_run`.

In the code above, when all the workers are occupied, the `if` condition returns `False` and directly retries without sleeping at all. This is bad practice and consumes a lot of CPU, so other threads are slower. This is what we observed when the `update_job_status` is executed first. The callback thread is slower. It is not the case when it is executed after the `release_worker`, because the previous code does not directly loop. It is sleeping thanks to the `self.orchestrator.jobs_to_run.get()` before looping again.

### The solution is easy:
We just need to sleep to go faster!

``` python
    def run(self):
        from time import sleep
        _TaipyLogger._get_logger().info("Start job dispatcher...")
        while not self._STOP_FLAG:
            try:
                if self._can_execute():
                    with self.lock:
                        job = self.orchestrator.jobs_to_run.get(block=True, timeout=0.1)
                    self._execute_job(job)
                else:
                    sleep(0.1)
            except Exception:  # In case the last job of the queue has been removed.
                pass

    def _can_execute(self) -> bool:
        """Returns True if the dispatcher have resources to execute a new job."""
        return self._nb_available_workers > 0
```

gmarabout (approved):
Awesome!! Well done @jrobinAV 👏🏻

FredLL-Avaiga:
does this work with sleep(0) ?

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
```diff
@@ -69,6 +70,8 @@ def run(self):
                             break
                         job = self.orchestrator.jobs_to_run.get(block=True, timeout=0.1)
                     self._execute_job(job)
+                else:
+                    time.sleep(0.1)  # We need to sleep to avoid busy waiting.
```
gmarabout, line 74: 💭 Maybe this value should be configurable via env variable/configuration.

gmarabout (approved):

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
FredLL-Avaiga: Not sure that would bring any value

FredLL-Avaiga (approved):

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
gmarabout: I think it might. Here we arbitrarily wait for 0.1 second. Why not 0.01 ? Or 1 second?

gmarabout: The number of workers is configurable. I see no reason for wait time in dispatcher loop not to be configurable as well. 
It should just come with reasonable defaults (and `0.1` looks a reasonable value).

gmarabout (approved):

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
FredLL-Avaiga: I think this `sleep()` is really technical ie it allows to give back control to the system and the value is not really important here: it should be the minimum possible.
That's why I asked if sleep(0) was working ...

trgiangdo (approved):

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
toan-quach: I agree with Gregoire, though then it makes the issue a bit more complicated as too much option isn't necessarily a good thing. If we decide to work on it, maybe a different PR would be better too

toan-quach (approved):

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
gmarabout: I totally agree ! This is not blocking for me and could eventually be addressed later if there is a need.

PR #863, opened 2024-02-22
feature/ added retry to save and load entity in sql repo

toan-quach:
# Changes
- Added retry mechanism for load and save entities in sql repository


taipy/core/_repository/_sql_repository.py
```diff
@@ -11,18 +11,23 @@
...
 
 from .._repository._abstract_repository import _AbstractRepository
+from ..common._utils import _retry_read_entity
 from ..common.typing import Converter, Entity, ModelType
 from ..exceptions import ModelNotFound
 from .db._sql_connection import _SQLConnection
 
 
 class _SQLRepository(_AbstractRepository[ModelType, Entity]):
+    __EXCEPTIONS_TO_RETRY = (ModelNotFound,)
```
jrobinAV: The modelNotFound exception meaning is not necessary an error happened. 
It could be that a user is trying to get an entity that has just been deleted by another user.

We don't want to rely on this exception to retry.

What we want is to have an exception that indicates that the query cannot be performed correctly. Not that the entity does not exist. Am I correct?

toan-quach: Yep that's true. I will create a new Exception for it later.

taipy/core/_repository/_sql_repository.py
```diff
@@ -205,10 +211,20 @@ def __insert_model(self, model: ModelType):
...
-    def _update_entry(self, model):
+    def __update_entry(self, model):
         query = self.table.update().filter_by(id=model.id)
-        self.db.execute(str(query.compile(dialect=sqlite.dialect())), model.to_list() + [model.id])
+        cursor = self.db.execute(str(query.compile(dialect=sqlite.dialect())), model.to_list() + [model.id])
         self.db.commit()
+        cursor.close()
+
+    def _update_entry(self, model):
+        start = datetime.now()
```
toan-quach: @jrobinAV My main question for this PR is should we allow user to determine the value for how long to wait or how many time to retry setting the entity?

trgiangdo: Should we use the same mechanism as we are having on FS repo?
Probably the same decorator

toan-quach: i dont see why not :D

taipy/core/exceptions/exceptions.py
```diff
@@ -186,7 +186,7 @@ def __init__(self, sequence_id: str):
 class NonExistingSequence(Exception):
     """Raised if a requested Sequence is not known by the Sequence Manager."""
 
-    def __init__(self, sequence_id: str, scenario_id: str=None):
+    def __init__(self, sequence_id: str, scenario_id: str = None):
```
jrobinAV: That's not the convention. In the default values definition, in method signatures, there is no space around the '='.

taipy/core/_repository/db/_sql_connection.py
```diff
@@ -87,4 +77,4 @@ def _build_connection() -> Connection:
 
 @lru_cache
 def __build_connection(db_location: str):
-    return sqlite3.connect(db_location, check_same_thread=False)
+    return sqlite3.connect(db_location, check_same_thread=False, timeout=200)
```
jrobinAV: I believe 200 seconds is unreasonable.

taipy/core/_repository/_sql_repository.py
jrobinAV: I would not let the user define how long to wait. 
Why not for the retry nb, but it has to be for good reasons.

With `filesystem`, we can know if a file exists before trying to read it. So if the file exists and we cannot read it, we can raise an exception that is catch to retry.
With `sql`, we must read the database to know if the entity exists. So if reading the entity fails, we cannot know if it is because it does not exist (which is normal) or because of a technical issue like the locked database. We need to understand how to differentiate the 2 cases to implement the retry.

jrobinAV (commented):
I believe the retry should not be at the `load` method level, but deeper. I guess the right place should be at the `self.db.execute` level. That means we need to slightly refactor the SQLRepository class.

By the way, I am concerned about the `_load_all` and the `_search` methods that are executing queries in a loop. I believe we can build a single query so we do just one query.

What do you think?

taipy/core/_repository/db/_sql_connection.py
toan-quach: I agree, I'd say 15 or lower is good enough

taipy/core/exceptions/exceptions.py
toan-quach: oh? not sure why this was changed!

taipy/core/_repository/_sql_repository.py
toan-quach: I think for now, I'd prefer to have something simple to see if it fixes the issue on integration tests. the function for retry we have in FS is for read only and the discussion we're having is for write in sql. So to have something for it, we should introduce new term like the one using in read function `Config.core.read_entity_retry`, we should have something like `Config.core.write_entity_retry` for write also. What do you guys think?

jrobinAV (changes requested):
I am requesting changes on this PR just to be sure we don't merge it by mistake until it is further discussed.

taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
```diff
@@ -61,6 +61,7 @@ def _dispatch(self, job: Job):
         config_as_string = _TomlSerializer()._serialize(Config._applied_config)  # type: ignore[attr-defined]
         future = self._executor.submit(_TaskFunctionWrapper(job.id, job.task), config_as_string=config_as_string)
         future.add_done_callback(partial(self._update_job_status_from_future, job))
+        future.add_done_callback(self._release_worker)  # We must release the worker before updating the job status
```
jrobinAV:
```suggestion
```

jrobinAV (dismissed):
I believe it is ready to be reviewed for merging.

jrobinAV (approved):

jrobinAV (approved):

PR #864, opened 2024-02-22
GUI: Json Resolver (#826)

### LINKED ISSUE #826: Provide JSONEncoder renderes for Chalk'it
Chalk'it widgets need that all viewable Python objects are previoulsy converted to JSON before sending them to Chalk'it. Currently, these include objects of type:
- plotly.graph_objs._figure.Figure
- pandas.core.frame.DataFrame
- folium.folium.Map
- geopandas.geodataframe.GeoDataFrame
- PIL.Image.Image

in addition to legacy Taipy JSONEncoders.

This JSON encoding needs to be handled without compromising legacy encoding for Taipy markdown or html pages.
See [json.py](https://github.com/bengaid/taipy/blob/experimental-chalkit/taipy/gui/_renderers/json.py) and associated [chalkit_helper.py](https://github.com/bengaid/taipy/blob/experimental-chalkit/taipy/gui/_renderers/json.py)

dinhlongviolin1:
Resolve #826 

Demo usage

```python
from taipy.gui import JsonAdapter

class FunctionJsonAdapter(JsonAdapter):
    def parse(self, o):
        if isinstance(o, FunctionType):
            return o.__name__
```

### DIFF
--- a/taipy/gui/_renderers/json.py
+++ b/taipy/gui/_renderers/json.py
@@ -10,6 +10,8 @@
 # specific language governing permissions and limitations under the License.
 from __future__ import annotations
 
+import typing as t
+from abc import ABC, abstractmethod
 from datetime import date, datetime, time
 from json import JSONEncoder
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


taipy/gui/_renderers/json.py
```diff
@@ -20,35 +22,60 @@
...
+            return o._to_dict()
+        if isinstance(o, _MapDict):
+            return o._dict
+        if isinstance(o, _TaipyBase):
+            return o.get()
+        if isinstance(o, (datetime, date, time)):
+            return _date_to_string(o)
+        if isinstance(o, Path):
+            return str(o)
+        if isinstance(o, FunctionType):
```
FredLL-Avaiga: Are the 2 (FunctionType and lambdaType) new ?
Could this have an impact on taipy-gui : like no error when you get a function when your not supposed to ?

FredLL-Avaiga (dismissed):
Performance if there are many resolver ?

taipy/gui/_renderers/json.py
dinhlongviolin1: It was new. I added it on a recent PR for the frontend decopule. This is just json encoding so I don't think it will affect anything. Previously you will get an error and the result is None. Now you get the function name with no error which is fine tbh

dinhlongviolin1:
> Performance if there are many resolver ?

We should be good performance wise. the isinstance check is not resource intensive at all and we will only resolve if it matches the if condition.

taipy/gui/_renderers/json.py
FredLL-Avaiga: less error reporting but I guess it's OK
Or should those be handled by the chalk'it Json resolver ?

dinhlongviolin1: I like that proposal!. For normal taipy app you will never encounter this issue. Only for custom frontend will you encounter this when getting the function list. 
Let's wait for what @FabienLelaquais says.

taipy/gui/_renderers/json.py
```diff
@@ -20,35 +22,60 @@
...
+            return o.__name__
+        if isinstance(o, LambdaType):
+            return str(o)
+
+
+class _TaipyJsonResolver(object, metaclass=_Singleton):
+    def __init__(self):
+        self._resolver: t.List[JsonResolver] = []
```
FabienLelaquais: I would have called that '_resolvers'...

taipy/gui/_renderers/json.py
```diff
@@ -20,35 +22,60 @@
...
+        self.register(_DefaultJsonResolver())
+
+    def register(self, resolver: JsonResolver):
+        self._resolver.append(resolver)
+
+    def resolve(self, o):
+        try:
+            for resolver in self._resolver:
```
FabienLelaquais: Don' t we want to have the user's resolver take precedence?
Then read the list backwards??

FabienLelaquais (dismissed):
I like that. Easy..!

FabienLelaquais:
> > Performance if there are many resolver ?
> 
> We should be good performance wise. the isinstance check is not resource intensive at all and we will only resolve if it matches the if condition.

Agreed. And there will certainly never be more that a single custom adapter.

FabienLelaquais (dismissed):
👌

FabienLelaquais (approved):
👍

PR #871, opened 2024-02-23
Fix/#869 - Investigate using Queue without lock

### LINKED ISSUE #869: Investigate using queue without lock
In standalone job execution mode, when get and put item into queue, we are using a lock.

However, python `Queue()` is already thread safe (as mentioned in [their doc](https://docs.python.org/3/library/queue.html).

Investigate if we can safely remove our lock.

trgiangdo:
Resolves #869

### DIFF
--- a/taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
@@ -37,7 +37,6 @@ class _JobDispatcher(threading.Thread):
         threading.Thread.__init__(self, name="Thread-Taipy-JobDispatcher")
         self.daemon = True
         self.orchestrator = orchestrator
-        self.lock = self.orchestrator.lock  # type: ignore
         Config.block_update()
 
     def start(self):
@@ -64,10 +63,9 @@ class _JobDispatcher(threading.Thread):
         while not self._STOP_FLAG:
             try:
                 if self._can_execute():
-                    with self.lock:
-                        if self._STOP_FLAG:
-                            break
-                        job = self.orchestrator.jobs_to_run.get(block=True, timeout=0.1)
+                    if self._STOP_FLAG:
+                        break
+                    job = self.orchestrator.jobs_to_run.get(block=True, timeout=0.1)
                     self._execute_job(job)
                 else:
                     time.sleep(0.1)  # We need to sleep to avoid busy waiting.
@@ -99,11 +97,10 @@ class _JobDispatcher(threading.Thread):
 
     def _execute_jobs_synchronously(self):
         while not self.orchestrator.jobs_to_run.empty():
-            with self.lock:
-                try:
-                    job = self.orchestrator.jobs_to_run.get()
-                except Exception:  # In case the last job of the queue has been removed.
-                    self._logger.warning(f"{job.id} is no longer in the list of jobs to run.")
+            try:
+                job = self.orchestrator.jobs_to_run.get()
+            except Empty:
+                pass
             self._execute_job(job)
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


jrobinAV (commented):
I am concerned about the `orchestrator.cancel_job()` method.

If we remove the lock on the job dispatcher, we can be in a strange state: 
Let's consider a job A. We can start canceling A by removing some subsequent jobs from the `blocked_job` list. Then in parallel, A is enqueued to be executed. Then finally we continue canceling A trying to remove it from the `jobs_to_run` queue (it is not in the queue anymore).

What do you think?

trgiangdo:
We keep the lock in the `orchestrator.cancel_job()`. We can only remove locks from method that interact with Queue

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
```diff
@@ -93,11 +91,10 @@ def _execute_job(self, job: Job):
 
     def _execute_jobs_synchronously(self):
         while not self.orchestrator.jobs_to_run.empty():
-            with self.lock:
-                try:
-                    job = self.orchestrator.jobs_to_run.get()
-                except Exception:  # In case the last job of the queue has been removed.
-                    self._logger.warning(f"{job.id} is no longer in the list of jobs to run.")
```
trgiangdo, line 106: The `except` here is wrong because in incase the last job has been removed, the `job.id` in the warning will be the previous job id

toan-quach:
I noticed you're focusing only on locks used in dispatcher, what about the ones in orchestrator like submit function?

trgiangdo:
For some code blocks using a lock in the `_orchestrator.py`, they are either:
1. Executing multiple commands, not just get and put from queue or
2. Doesn't interact with Queue, which is not thread-safe.

Either case, the lock is needed there

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
```diff
@@ -96,11 +94,10 @@ def _execute_job(self, job: Job):
...
+            try:
+                job = self.orchestrator.jobs_to_run.get()
+            except Empty:
+                pass
```
jrobinAV, line 103: Well spotted. We cannot log the job id!

However, I would still except all Exceptions, just to be sure that we keep running if any other exception is raised.

```suggestion
            except Exception:
                pass
```

jrobinAV:
> We keep the lock in the `orchestrator.cancel_job()`. We can only remove locks from method that interact with Queue

I believe we cannot remove the lock at all because when canceling a job or when failing subsequent jobs, we do instantiate a new queue

```
    @classmethod
    def _remove_jobs_to_run(cls, jobs):
        new_jobs_to_run = Queue()
```

That means we cannot rely on the locking mechanism of a queue instance since we may have multiple instances.

jrobinAV (changes requested):

trgiangdo:
After more investigation, I believe the mechanism when using `_Orchestrator.lock` and `Queue.lock` are not working properly before

First:
- `_Orchestrator.lock` is a `multiprocessing.Lock` which works between processes
- The `Queue.lock` is a `threading.lock` which works between threads, which is what we actually needed here to properly communicate between the orchestrator thread and dispatcher thread.

Second, all `_remove_jobs_to_run()` calls are also in the `_Orchestrator.lock` context, so:
- If the dispatcher also uses the `_Orchestrator.lock`, the dispatcher can not access the Queue when the orchestrator is using the lock.
- By removing the lock from the dispatcher, the dispatcher can still access the queue when the orchestrator is inside its lock context, but can not access the queue at the same time as the orchestrator since they are handling the same lock of the same queue. This can cause issue for sure.

Initializing a new Queue in `__remove_jobs_to_run()` is also an interesting point and may cause issue.

However, we were still suffer from these problem with the previous implementation since the `_Orchestrator.lock` is a multiprocess lock.

In conclusion, I propose to replace the `_Orchestrator.lock` by a `threading.lock`, and then use that lock in the dispatcher.

toan-quach:
Interesting findings!! Didn't think about that before!

toan-quach (approved):

trgiangdo:
Accessing the `_Orchestrator.jobs_to_run` Queue without the lock is not safe since the `jobs_to_run` Queue can be altered

PR #878, opened 2024-02-26
Fix/#874 - Block config and create version when create entities

### LINKED ISSUE #874: BUG- Version management is messed up when create_scenario or create_global_data_node is called
**Description**
Fix Version management when creating entities (global data node or scenario) without running the Core service. 

**Expected behavior**
Taipy should manage the version and block config when the two methods are called before running the Core service.

**Acceptance Criteria**
- [ ] Create related issues in taipy-doc for documentation and Release Notes if relevant

trgiangdo:
Resolves #874 

In Taipy 3.0, we only create the version entity and block Configuration update when starting the Core service.
This makes application that creating entities without running the Core service will have entities that doesn't have an actual version, which can be dangerous.

In this PR:
- Block configuration update and create the version entity whenever an entity is created via `taipy.create_scenario` or `taipy.create_global_data_node` APIs.
- Add log message when block or unblock the Configuration.


taipy/core/_core.py
```diff
@@ -84,20 +86,37 @@ def stop(self, wait: bool = True, timeout: Optional[float] = None):
         with self.__class__.__lock_is_running:
             self.__class__._is_running = False
 
-    @staticmethod
-    def __update_core_section():
+        with self.__class__.__lock_version_is_initialized:
+            self.__class__._version_is_initialized = False
```
trgiangdo, line 85: This is to unblock the configuration update, mainly for Notebook use-case, where the user wants to add or modify the Config after running the app for a while for testing.

We of course can only `Config.unblock` here, but then when the Core service start again, the version will not be manage. This can be good or bad base on the use-case. 

So the question here is that when re-start the Core service (in notebook), do we want to:
- clean all entities (manage version)
- keep all entities (only unblock the config)

taipy/core/taipy.py
```diff
@@ -846,13 +850,22 @@ def create_scenario(
 
     Returns:
         The new scenario.
+
+    Raises:
+        SystemExit: If the configuration check returns some errors.
+
     """
+    Core._manage_version_and_block_config()
+
```
jrobinAV, line 859:
```suggestion
```

taipy/core/taipy.py
```diff
@@ -861,32 +874,40 @@ def create_global_data_node(config: DataNodeConfig) -> DataNode:
...
     Raises:
         DataNodeConfigIsNotGlobal^: If the data node configuration does not have GLOBAL scope.
+        SystemExit: If the configuration check returns some errors.
     """
     # Check if the data node config has GLOBAL scope
     if config.scope is not Scope.GLOBAL:
         raise DataNodeConfigIsNotGlobal(config.id)
 
+    Core._manage_version_and_block_config()
+
```
jrobinAV, line 884:
```suggestion
```

tests/core/config/checkers/test_migration_config_checker.py
```diff
@@ -24,14 +24,6 @@ def mock_func():
...
 
 def test_check_if_entity_property_key_used_is_predefined(caplog):
-    with patch("sys.argv", ["prog", "--production", "1.0"]):
-        core = Core()
-        core.run()
-    assert caplog.text == ""
-    core.stop()
-
-    caplog.clear()
-
```
jrobinAV, line 34: If I understand correctly, this tests what happens when we run the Core service without any config.
Why do we need to remove this test?

tests/core/config/checkers/test_migration_config_checker.py
```diff
@@ -72,9 +64,9 @@ def test_check_valid_version(caplog):
     Config.unblock_update()
 
     with patch("sys.argv", ["prog", "--production", "2.0"]):
+        # No SystemExit should be raised
         core = Core()
         core.run()
-    assert caplog.text == ""
```
jrobinAV, line 77: Don't we want to test the generated logs anymore?

tests/core/notification/test_events_published.py
```diff
@@ -249,7 +251,7 @@ def test_job_events():
     consumer.start()
 
     # Create scenario
-    scenario = tp.create_scenario(sc_config)
+    scenario = _ScenarioManagerFactory._build_manager()._create(sc_config)
```
jrobinAV, line 254: I don't get why you replaced `scenario = tp.create_scenario(sc_config)` by 
`scenario = _ScenarioManagerFactory._build_manager()._create(sc_config)`.

jrobinAV (approved):
I don't get why you renamed `input` and `output` data node configs, but I am ok with the change.

trgiangdo:
Oh @jrobinAV, it's because "input" and "output" are attributes of a task, which should not be named for a dn config.

In those tests, we did not check the Config before, and now we do.
That's why I renamed those

jrobinAV:
> Oh @jrobinAV, it's because "input" and "output" are attributes of a task, which should not be named for a dn config.
> 
> In those tests, we did not check the Config before, and now we do. That's why I renamed those

Oh, right. Mystery solved! 
Thx

tests/core/config/checkers/test_migration_config_checker.py
trgiangdo: The purpose of this test is to check if the command line has any error. However since we add some new log, it's hard to test this.

Also, this case has already been tested elsewhere in the version_cli test, so it's not needed here.
We just check for the fail cases now.

tests/core/config/checkers/test_migration_config_checker.py
trgiangdo: Same reason as above

tests/core/notification/test_events_published.py
trgiangdo: tp.create_scenario will result in some other event, which should not be testes here

PR #884, opened 2024-02-26
order scenarios (and cycles)

### LINKED ISSUE #165: Scenario Selector - Chronological order for cycles
## Feature Improvement

The scenario selector would benefit from having scenarios in chronological order.

Somehow related to #169.

FredLL-Avaiga:
so that we always get the same order on UI (by creation date)

resolves #165


FabienLelaquais (approved):
👍

PR #885, opened 2024-02-26
:zap: replace multiprocessing.Lock by threading.Lock

jrobinAV:
Replace the orchestrator multiprocessing.lock by a threading.lock.
It is used by:
- the orchestrator to put jobs in the `jobs_to_run` queue when jobs are ok to be run (status is `PENDING`)
- the orchestrator to put jobs in the `blocked_jobs` list when jobs are blocked by another job (status is `BLOCKED`)
- the orchestrator to remove jobs from `jobs_to_run` queue when jobs are cancelled or abandoned (status is `CANCELED` or `ABANDONED`)
- the orchestrator to remove jobs from `blocked_jobs` list when jobs are cancelled or abandoned (status is `CANCELED` or `ABANDONED`)
- the dispatcher done callback thread to move jobs from the `blocked_jobs` list to the `jobs_to_run` queue when jobs are unblocked by another job completion.


toan-quach (approved):
lgtm

PR #904, opened 2024-03-01
feat: update build and release prod action

joaoandre-avaiga:


jrobinAV (approved):
With this action, we are forced to publish the packages. 
That means we cannot test it before pushing it on Pypi.

I am ok with that, but we need to be aware of it.

PR #935, opened 2024-03-05
GUI: Allow dict update on custom frontend (#932)

### LINKED ISSUE #932: BUG-onChange callback not called for JSON update in gui
To reproduce

- Clone or pull from repositoy <https://github.com/bengaid/covid_taipy_chalkit>
- Install Chalk'it with Taipy link
```sh
pip install py-chalk-it-0.4.0.tar.gz
```
- Run the example
```sh
python covid_demo.py
```

When the user changes the country, the Python callback is correctly called, option_e dict is correctly computed. It is sent to front-end but the onChange callback is not called.

dinhlongviolin1:
Resolve #932

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -1002,7 +1002,7 @@ class Gui:
                 elif isinstance(newvalue, _TaipyToJson):
                     newvalue = newvalue.get()
                 if isinstance(newvalue, (dict, _MapDict)):
-                    # Skip in taipy-gui, avaialble in custom frontend
+                    # Skip in taipy-gui, available in custom frontend
                     if request.cookies.get(_Server._RESOURCE_HANDLER_ARG, None) is None:
                         continue  # this var has no transformer
                 if isinstance(newvalue, float) and math.isnan(newvalue):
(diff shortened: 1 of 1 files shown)


FabienLelaquais (dismissed):
😶

taipy/gui/gui.py
```diff
@@ -1002,7 +1002,9 @@ def __send_var_list_update(  # noqa C901
                 elif isinstance(newvalue, _TaipyToJson):
                     newvalue = newvalue.get()
                 if isinstance(newvalue, (dict, _MapDict)):
-                    continue  # this var has no transformer
+                    # Skip in taipy-gui, available in custom frontend
+                    if request.cookies.get(_Server._RESOURCE_HANDLER_ARG, None) is None:
```
FredLL-Avaiga: Are we sure the request is always available ?
what if the update is done from a long callback ?

FredLL-Avaiga: should we at least try/except this or check the availability or request and request.cookies ?

FredLL-Avaiga (dismissed):

taipy/gui/gui.py
dinhlongviolin1: I more than agree with you!

FredLL-Avaiga (dismissed):
😍

FredLL-Avaiga (approved):
🎯

FabienLelaquais (approved):
Sure, why not

Issue #797, opened 2024-02-06
Propagating changes for custom exposed type to parquet and sql


toan-quach:
We have created a proposal for refactoring custom exposed type in csv and excel DN, we would like to propagate the changes to other tabular DN including parquet and SQL

PR #936, opened 2024-03-06
Feature/#797 custom exposed type sql and parquet

toan-quach:
#797


jrobinAV (commented):
I believe the code is working and meets the requirements.

However, the whole inheritance mechanism for all data nodes would benefit from some cleaning. Probably as part of another issue/PR.

For example:
- The `_AbstractTabularDataNode` is not a data node but a mixin. The name is misleading. It is more like a mixin.
- Same for `_AbstractFileDataNode`.
- However, the `_abstract_sql` is a data node sub-class. Why this inconsistency?
-  The filterDataNode is not a data node sub-class nor a mixin. It is just a delegated class. I guess it should be renamed to DataNodeFilter.
- The exposed type code is duplicated in many data node implementations. I believe we can factorize the code in the `_AbstractTabularDataNode` class since the exposed type is for tabular data.
- other minor improvements

@toan-quach What do you think? Should we create another issue to clean this?

toan-quach:
@jrobinAV 
I agree with you points and I think we should have a separate tissue + PR for it

jrobinAV:
> @jrobinAV I agree with you points and I think we should have a separate tissue + PR for it

OK super. Can you create the issue?

taipy/core/data/parquet.py
```diff
@@ -249,13 +251,13 @@ def write_with_kwargs(self, data: Any, job_id: Optional[JobId] = None, **write_k
         }
         kwargs.update(self.properties[self.__WRITE_KWARGS_PROPERTY])
         kwargs.update(write_kwargs)
-        if isinstance(data, pd.DataFrame):
-            data.to_parquet(self._path, **kwargs)
+        if isinstance(data, pd.Series):
+            df = pd.DataFrame(data)
         else:
-            _df = pd.DataFrame(data)
-            # Ensure that the columns are strings, otherwise writing will fail with pandas 1.3.5
```
trgiangdo, line 256: I would keep this comment and move it before line 259

taipy/core/data/sql_table.py
```diff
@@ -123,26 +122,12 @@ def _do_write(self, data, engine, connection) -> None:
...
-            self.__insert_dicts(data, table, connection, delete_table)
-        # If data is a primitive type, it will be inserted as a tuple of one element.
-        else:
-            self.__insert_tuples([(x,) for x in data], table, connection, delete_table)
+        self.__insert_dataframe(
+            self._convert_data_to_dataframe(self.properties[self._EXPOSED_TYPE_PROPERTY], data),
+            table,
+            connection,
+            delete_table,
+        )
```
trgiangdo, line 130: I love this.

Btw, this makes the `__insert_tuples()` method right?
I think we can safely remove it

toan-quach:
@jrobinAV here is the link to the new issue https://github.com/orgs/Avaiga/projects/6/views/2?pane=issue&itemId=55998996

trgiangdo (approved):
LGTM

jrobinAV (approved):

PR #938, opened 2024-03-06
add on_action(state, entity) to scenario_dag

### LINKED ISSUE #768: Scenario DAG - Have Data Node Viewer open when clicking on Data Node
**Description**

Clicking on a Data Node in the Scenario DAG should show a Data Node viewer. (Maybe in a dialog)

**Discussion**

We don't want to enforce a behavior that covers a single use case.
Getting notified when a Taks or a DataNode is selected in the DAG would be good, and let the user decide what's next.
This needs more discussions.

**Acceptance Criteria**

FredLL-Avaiga:
resolves #768

```
def on_dag_action(state: State, entity: Task | DataNode):
    print(f"on_dag_action(state, {entity})")

...

<|{scenario}|scenario_dag|on_action=on_dag_action|>

```

### DIFF
--- a/frontend/taipy/package-lock.json
+++ b/frontend/taipy/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "taipy-gui-core",
-  "version": "3.1.0",
+  "version": "3.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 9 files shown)


FabienLelaquais (approved):
I like that!

PR #939, opened 2024-03-06
Move specific attributes from job_dispatcher to standalone job dispatcher

jrobinAV:
Move `dispatched_processes` and `nb_of_available_workers`

This is in preparation of a better subprocess pid management proposal.

### DIFF
--- a/taipy/core/_orchestrator/_abstract_orchestrator.py
+++ b/taipy/core/_orchestrator/_abstract_orchestrator.py
@@ -21,6 +21,16 @@ from ..task.task import Task
 class _AbstractOrchestrator:
     """Creates, enqueues, and orchestrates jobs as instances of `Job^` class."""
 
+    @property
+    @abstractmethod
+    def jobs_to_run(self):
+        pass
+
+    @property
+    @abstractmethod
+    def blocked_jobs(self):
+        pass
+
     @classmethod
     @abstractmethod
     def initialize(cls):
--- a/taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
@@ -50,5 +50,5 @@ class _DevelopmentJobDispatcher(_JobDispatcher):
     def _is_dispatched(self, job_id: str) -> bool:
         return False
 
-    def _remove_job(self, job_id: str):
+    def _remove_dispatched_job(self, job_id: str):
         pass
--- a/taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
@@ -23,6 +23,9 @@ class _DevelopmentJobDispatcher(_JobDispatcher):
     def __init__(self, orchestrator: _AbstractOrchestrator):
         super().__init__(orchestrator)
 
+    def _can_execute(self) -> bool:
+        return True
+
     def start(self):
         raise NotImplementedError
 
@@ -43,3 +46,9 @@ class _DevelopmentJobDispatcher(_JobDispatcher):
         """
         rs = _TaskFunctionWrapper(job.id, job.task).execute()
         self._update_job_status(job, rs)
+
+    def _is_dispatched(self, job_id: str) -> bool:
+        return False
+
+    def _remove_job(self, job_id: str):
+        pass
--- a/taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
@@ -46,9 +46,3 @@ class _DevelopmentJobDispatcher(_JobDispatcher):
         """
         rs = _TaskFunctionWrapper(job.id, job.task).execute()
         self._update_job_status(job, rs)
-
-    def _is_dispatched(self, job_id: str) -> bool:
-        return False
-
-    def _remove_dispatched_job(self, job_id: str):
-        pass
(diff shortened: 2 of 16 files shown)


taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
```diff
@@ -43,3 +46,9 @@ def _dispatch(self, job: Job):
         """
         rs = _TaskFunctionWrapper(job.id, job.task).execute()
         self._update_job_status(job, rs)
+
+    def _is_dispatched(self, job_id: str) -> bool:
+        return False
+
+    def _remove_job(self, job_id: str):
```
toan-quach: I'd prefer `_remove_dispatched_job` to avoid confusion

taipy/core/_orchestrator/_abstract_orchestrator.py
```diff
@@ -21,6 +21,16 @@
...
 
+    @property
+    @abstractmethod
+    def jobs_to_run(self):
+        pass
+
+    @property
+    @abstractmethod
+    def blocked_jobs(self):
+        pass
```
toan-quach, line 32: I don't think these 2 functions were implemented right?

taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -328,3 +329,6 @@ def _check_and_execute_jobs_if_development_mode():
 
         if dispatcher := _OrchestratorFactory._dispatcher:
             dispatcher._execute_jobs_synchronously()
+
+    def is_dispatched(self, job_id: JobId) -> bool:
```
toan-quach: I thought we want to move this to dispatcher?

toan-quach (dismissed):
lgtm, minor questions

taipy/core/_orchestrator/_abstract_orchestrator.py
jrobinAV: Right. These are two properties. 
The dispatchers are using these two properties. However, since dispatchers take an abstract orchestrator as input, the two properties should be declared at the abstract orchestrator level and not only at the orchestrator implementation level.

taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
jrobinAV: Yes, that's better.

taipy/core/_orchestrator/_orchestrator.py
jrobinAV: Yes, indeed. Let me change that.

taipy/core/_orchestrator/_abstract_orchestrator.py
toan-quach: ahhh my bad, I looked too quickly and thought this is the dispatcher file, was wondering why we're declaring these 2 properties in the dispatchers haha

hmm in that case, I think we should still define it in the orchestrator no? or is it not needed because they already exist as class attribute?

jrobinAV: They are already defined as attributes, indeed.

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
```diff
@@ -29,9 +29,9 @@ class _JobDispatcher(threading.Thread):
     """Manages job dispatching (instances of `Job^` class) on executors."""
 
     _STOP_FLAG = False
-    _dispatched_processes: Dict = {}
```
trgiangdo, line 32: Oh so we do not use this anymore. That's good to know

tests/core/_orchestrator/_dispatcher/mock_standalone_dispatcher.py
```diff
@@ -10,7 +10,7 @@
 # specific language governing permissions and limitations under the License.
 
 from concurrent.futures import Executor, Future
-from typing import List
+from typing import Dict, List
```
trgiangdo: This is redundant right?

tests/core/_orchestrator/test_orchestrator__cancel_jobs.py
```diff
@@ -89,11 +91,10 @@ def test_cancel_job_with_subsequent_jobs_and_parallel_jobs():
     job3 = orchestrator._lock_dn_output_and_create_job(scenario.t3, "s_id", "e_id")
     job2bis = orchestrator._lock_dn_output_and_create_job(scenario.t2bis, "s_id", "e_id")
     job1.completed()
-
-    job2.running()
+    job2.pending()
```
trgiangdo, line 94: Why is this status changed?

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
jrobinAV: Well, that's the purpose of my proposal. It's up to us to reject it.

tests/core/_orchestrator/test_orchestrator__cancel_jobs.py
jrobinAV: When a job was in the dispatched_processes dict, it was necessarily running. The test was wrong here since the job2 status should have already been pending.
Before the current PR, we tested if the job was in the dispatched_processes to know if we could cancel it (that's why the test passed). Now, we are testing if the job status is running instead.

toan-quach (approved):

tests/core/_orchestrator/test_orchestrator__cancel_jobs.py
trgiangdo: Ah I see, good catch. Thank you

PR #951, opened 2024-03-08
Test/integration testing

jrobinAV:


taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -58,15 +58,15 @@ def submit(
...
+             submittable (Union[Scenario^, Sequence^]): The scenario or sequence to submit for execution.
              callbacks: The optional list of functions that should be executed on jobs status change.
              force (bool) : Enforce execution of the scenario's or sequence's tasks even if their output data
                 nodes are cached.
              wait (bool): Wait for the orchestrated jobs created from the scenario or sequence submission to be
                 finished in asynchronous mode.
              timeout (Union[float, int]): The optional maximum number of seconds to wait for the jobs to be finished
                 before returning.
-             **properties (dict[str, any]): A keyworded variable length list of user additional arguments
+             **properties (dict[str, any]): A key worded variable length list of user additional arguments
```
toan-quach, line 69: Wouldnt this be clearer? 
```suggestion
             **properties (dict[str, any]): A dictionary of user additional arguments
```

toan-quach (dismissed):
minor comment, otherwise LGTM!

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
```diff
@@ -55,30 +55,37 @@ def stop(self, wait: bool = True, timeout: Optional[float] = None):
...
+            if job:
+                self._logger.debug(f"Got a job to execute {job.id}.")
+                try:
+                    if not self._STOP_FLAG:
+                        self._execute_job(job)
+                    else:
+                        self.orchestrator.jobs_to_run.put(job)
+                except Exception as e:
+                    self._logger.exception(e)
+        time.sleep(1)
```
jrobinAV: We sleep 1 second when stopping the dispatcher. The purpose is to stay alive and wait for the thread calling the stop method to have enough time to enter the if block (`if wait and self.is_alive():`), so we are sure the subthreads are joined before killing the thread.

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
```diff
@@ -55,30 +55,37 @@ def stop(self, wait: bool = True, timeout: Optional[float] = None):
             wait (bool): If True, the method will wait for the dispatcher to stop.
             timeout (Optional[float]): The maximum time to wait. If None, the method will wait indefinitely.
         """
-        self.stop_wait = wait
-        self.stop_timeout = timeout
         self._STOP_FLAG = True
+        if wait and self.is_alive():
+            self._logger.debug("Waiting for the dispatcher thread to stop...")
+            self.join(timeout=timeout)
```
jrobinAV: Moved from the run method because `join` must be called from another thread.

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
```diff
@@ -55,30 +55,37 @@ def stop(self, wait: bool = True, timeout: Optional[float] = None):
...
-                        if self._STOP_FLAG:
-                            break
+            if not self._can_execute():
+                time.sleep(0.1)  # We need to sleep to avoid busy waiting.
+                continue
+
+            with self.lock:
+                self._logger.debug("Acquiring lock to check jobs to run.")
+                job = None
+                try:
```
jrobinAV, line 73: We now only try the job retrieval and catch empty exception.

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
```diff
@@ -55,30 +55,37 @@ def stop(self, wait: bool = True, timeout: Optional[float] = None):
...
+            self._logger.debug(f"Releasing lock after checking jobs to run.")
+            if job:
+                self._logger.debug(f"Got a job to execute {job.id}.")
+                try:
+                    if not self._STOP_FLAG:
+                        self._execute_job(job)
+                    else:
+                        self.orchestrator.jobs_to_run.put(job)
+                except Exception as e:
+                    self._logger.exception(e)
```
jrobinAV, line 86: We separated the job retrieval and the job execution, so we had a try catch around the job execution.

taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
```diff
@@ -11,6 +11,7 @@
 
 from concurrent.futures import Executor, ProcessPoolExecutor
 from functools import partial
+from threading import RLock
```
jrobinAV: We are using a lock to access the shared variable `_nb_available_workers`

taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -12,7 +12,7 @@
 import itertools
 from datetime import datetime
 from queue import Queue
-from threading import Lock
+from threading import RLock
```
jrobinAV: using reccursive lock

taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -129,16 +131,18 @@ def submit_task(
...
                 itertools.chain([cls._update_submission_status], callbacks or []),
                 force,
             )
-        jobs = [job]
-        submission.jobs = jobs  # type: ignore
-        cls._orchestrate_job_to_run_or_block(jobs)
+            jobs = [job]
+            submission.jobs = jobs  # type: ignore
+            cls._orchestrate_job_to_run_or_block(jobs)
+            cls.__logger.debug(f"Releasing lock after submitting task {task.id}.")
```
jrobinAV: This is now inside a lock context.

taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -223,18 +227,25 @@ def _unlock_edit_on_jobs_outputs(jobs: Union[Job, List[Job], Set[Job]]):
...
             cls.__unblock_jobs()
         elif job.is_failed():
             cls._fail_subsequent_jobs(job)
 
     @classmethod
     def __unblock_jobs(cls):
-        for job in cls.blocked_jobs:
-            if not cls._is_blocked(job):
-                with cls.lock:
+        with cls.lock:
```
jrobinAV, line 235: The lock is global now to ensure all the transactionality of the unblock operation.

taipy/core/_repository/_sql_repository.py
```diff
@@ -47,9 +51,19 @@ def __init__(self, model_type: Type[ModelType], converter: Type[Converter]):
...
+            except DatabaseError as e:
+                self._logger.error(f"Error while updating {entity.id} in {self.table.name}. ")  # type: ignore
+                self._logger.error(f"Error : {e}")
+                raise e
+        try:
+            self.__insert_model(obj)
+        except DatabaseError as e:
+            self._logger.error(f"Error while inserting {entity.id} into {self.table.name}. ")  # type: ignore
+            self._logger.error(f"Error : {e}")
+            raise e
```
jrobinAV, line 66: Better trace errors.

taipy/core/_repository/db/_sql_connection.py
```diff
@@ -87,4 +87,4 @@ def _build_connection() -> Connection:
 
 @lru_cache
 def __build_connection(db_location: str):
-    return sqlite3.connect(db_location, check_same_thread=False)
+    return sqlite3.connect(db_location, check_same_thread=False, timeout=20)
```
jrobinAV, line 87: increasing the sql connection timeout

taipy/logger/_taipy_logger.py
```diff
@@ -34,7 +34,7 @@ def _get_logger(cls):
             cls.__logger.setLevel(logging.INFO)
             ch = logging.StreamHandler(sys.stdout)
             ch.setLevel(logging.INFO)
-            formatter = logging.Formatter("[%(asctime)s][%(name)s][%(levelname)s] %(message)s", "%Y-%m-%d %H:%M:%S")
-            ch.setFormatter(formatter)
+            f = logging.Formatter("[%(asctime)s.%(msecs)03d][%(name)s][%(levelname)s] %(message)s", "%Y-%m-%d %H:%M:%S")
```
jrobinAV, line 37: Add milliseconds to default logging format.

taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
```diff
@@ -36,30 +39,30 @@ def __init__(self, orchestrator: _AbstractOrchestrator, subproc_initializer: Opt
...
         future.add_done_callback(partial(self._update_job_status_from_future, job))
 
-    def _release_worker(self, _):
-        self._nb_available_workers += 1
-
     def _update_job_status_from_future(self, job: Job, ft):
+        with self._nb_available_workers_lock:
+            self._nb_available_workers += 1
+            self._logger.debug(f"Changing nb_available_workers to {self._nb_available_workers} from callback method.")
         self._update_job_status(job, ft.result())
```
jrobinAV: The two callbacks are merged.

taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
trgiangdo: I would revert this back to a normal Lock

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
toan-quach: should we use `self.is_running` instead of `self.is_alive`? They are the same by the way, just asking for semantic-wise

taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
toan-quach: Yep I agree :D

toan-quach (dismissed):

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
jrobinAV: I would go for `is_running` then, since the method name is `run`.

taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
jrobinAV: OK, then. I'll do the change.

taipy/core/_orchestrator/_orchestrator.py
trgiangdo: I would revert it back to normal Lock

tests/core/_orchestrator/_dispatcher/mock_standalone_dispatcher.py
```diff
@@ -39,9 +40,9 @@ def __init__(self, orchestrator: _AbstractOrchestrator):
         super(_StandaloneJobDispatcher, self).__init__(orchestrator)
         self._executor: Executor = MockProcessPoolExecutor()
         self._nb_available_workers = 1
+        self._nb_available_workers_lock = RLock()
```
trgiangdo: I would use normal Lock here as well

toan-quach (dismissed):

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
trgiangdo: Ain't joined and killed thread the same thing?
The `thread.join()` will wait for the thread to stop anyway, and this `sleep()` makes it wait 1 second longer.

taipy/logger/_taipy_logger.py
trgiangdo: Will this cause too much noise?

It affects the whole application so I think we need third opinion about this

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
jrobinAV: In the stop method, there is an if condition to check if the dispatcher thread is running. If it is not running, we don't join the callbacks sub-threads. 

So there is a possibility to be be in the following situation:
- the STOP_FLAG is set to True in the stop method
- The dispatcher thread terminates by completing the main loop.
- We check if it is running to join the sub-threads but can't join them since the dispatcher thread is not running anymore.
- We terminate the tests by removing the entities
- The sub-threads keep running and try to access the entities

Does it make sense?

This sleep(1) allows to give the time to the stop method to check the if condition before the dispatcher thread terminates.

taipy/logger/_taipy_logger.py
jrobinAV: My opinion is that in a multithreading/multiprocessing context, having milliseconds in the logs is more than useful to understand what's happening. This is not only true for us as developers but also for our users.

However, I agree we should discuss it during the daily to have more opinions on that.

taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py
trgiangdo: If the `is_running` returns False, there is no need to join actually.

The main reason we join is so that the `run()` method will be fully stop when the next test runs.

jrobinAV: Yep. OK.

taipy/logger/_taipy_logger.py
jrobinAV: @FlorianJacta @FabienLelaquais Is it ok for you to have a default log formatting with milliseconds?

trgiangdo (approved):

PR #964, opened 2024-03-13
Clean C405 rule for codestyle

ooooo-create:


trgiangdo (approved):
LGTM. Thank you @ooooo-create

PR #970, opened 2024-03-13
Support file reloading on notebook (#460)

### LINKED ISSUE #460: Automatic page reloading
**Description**

As of Taipy 2.3, changing the content of a file that defines a page's content (Markdown or HTML) is not considered. The programmer has to know that something has happened to the file content, then trigger a `Page.set_content()` to let Taipy know the file needs to be read again.
This is slightly cumbersome in the context of using Notebooks (but not only): you usually are in an IDE, and you may want to change a  page definition file... and impact the application on the fly.

I suspect we could actually do something that would solve many of the issues related to automatic refresh: when a page needs to rendered and if that page was created from a file and if that file still exists and if the last modification date is greater than the last time the file was read to render the page... then we trigger the `set_content()`.

I am not saying we should push the change to the browsers. The user would still have to manually refresh the page.
But at least, it would save a line of code, that is different from the page constructor's.

**Acceptance Criteria**

dinhlongviolin1:
Resolve #460


FredLL-Avaiga (dismissed):
Does this work on windows?

FabienLelaquais (dismissed):
WIth a little bit of magic... well done!

dinhlongviolin1:
> Does this work on windows?

No reason why it wont work on windows. Can you double check on your side for me

FredLL-Avaiga (approved):

FabienLelaquais (approved):
👌

PR #971, opened 2024-03-13
feat: replace python-linter with mypy specific action

joaoandre-avaiga:


.github/workflows/partial-tests.yml
```diff
@@ -9,14 +9,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: ricardochaves/python-lint@v1.4.0
+      - uses: jpetrucciani/mypy-check@master
```
trgiangdo, line 12: This action uses mypy 1.9.0. I like it :D

trgiangdo (commented):
The new action raises quite a number of new error. We need to handle these first before merging

.github/workflows/partial-tests.yml
joaoandre-avaiga: Yeah, I chose this one because it updates quite often. Now to fix the erros so the linter passes

trgiangdo (approved):
LGTM

jrobinAV (approved):

PR #984, opened 2024-03-18
Fix MapDict refresh error (#464)

### LINKED ISSUE #464: BUG-Error when calling 'refresh' with dataframes
**Description**
An error occurs when refresh is being used inside a function and called twice.

```
TaipyGuiWarning: on_action(): Exception raised in 'change()':
maximum recursion depth exceeded
TaipyGuiWarning: on_action(): 'change' is not a valid function.
```

**How to reproduce**
Call twice this function:

```python
from taipy.gui import Gui

data = {
    "x": [1, 2, 3],
    "y": [1, 2, 3],
    "y2": [4, 5, 1],
}

properties = {
    "x":"x",
    "y[1]":"y",
    "y[2]":"y2",
}

def change(state):
    state.data['y3'] = [14, 15, 11]
    state.refresh('data')
    print(state.data._dict)

Gui("<|{data}|chart|properties={properties}|> <|change|button|on_action=change|>").run()
```

**Expected behavior**
Calling this function again and again shouldn't cause any problem.

**Runtime environment**
Taipy 2.4

dinhlongviolin1:
Resolve #464

### DIFF
--- a/taipy/gui/utils/_bindings.py
+++ b/taipy/gui/utils/_bindings.py
@@ -41,7 +41,7 @@ class _Bindings:
         def __setter(ud: _Bindings, value: t.Any):
             if isinstance(value, _MapDict):
                 value._update_var = None
-            if isinstance(value, dict):
+            elif isinstance(value, dict):
                 value = _MapDict(value, None)
             ud.__gui._update_var(name, value)
 
--- a/taipy/gui/utils/_bindings.py
+++ b/taipy/gui/utils/_bindings.py
@@ -39,6 +39,8 @@ class _Bindings:
 
     def __get_property(self, name):
         def __setter(ud: _Bindings, value: t.Any):
+            if isinstance(value, _MapDict):
+                value._update_var = None
             if isinstance(value, dict):
                 value = _MapDict(value, None)
             ud.__gui._update_var(name, value)


taipy/gui/utils/_bindings.py
```diff
@@ -39,6 +39,8 @@ def _bind(self, name: str, value: t.Any) -> None:
 
     def __get_property(self, name):
         def __setter(ud: _Bindings, value: t.Any):
+            if isinstance(value, _MapDict):
```
FredLL-Avaiga, line 42: 😗

FredLL-Avaiga (dismissed):
no side effects ?

taipy/gui/utils/_bindings.py
FabienLelaquais: Oh whao

dinhlongviolin1: I have tested it and there should not be any obvious one since it is a setter and the update_var method is called right below it

taipy/gui/utils/_bindings.py
```diff
@@ -39,6 +39,8 @@ def _bind(self, name: str, value: t.Any) -> None:
...
+                value._update_var = None
             if isinstance(value, dict):
```
FabienLelaquais: So what will become a `elif` right?

FabienLelaquais (dismissed):
👌

taipy/gui/utils/_bindings.py
dinhlongviolin1: elif is fine too but not necessary :D

dinhlongviolin1: I change it anyway

FabienLelaquais: Well you avoid a test...

FredLL-Avaiga (approved):

FabienLelaquais (approved):
👍

PR #609, opened 2023-12-14
Refactor for Cleaner Code & Stability

LopeKinz:
**!!!NEVER MERGE INTO MAIN BRANCH!!!**

Refactored by Sourcery AI and Revieved by me.
Everything seems okay.
If Merged then send any issues to [LOPEKINZ/taipy](https://github.com/lopekinz/taipy) or pinkyhax@gmail.com

Stats:
**2** Commits
**153** Files Changed
**1,195** Additions
**1,123** Deletions

Most Common changes:
Replaced i with _ for Counting
Replaced `"String" + "String` with `f"String {string}"`
Changed Formats for better readability.

PR #1007, opened 2024-03-19
Refactor - Cleaner Code & Stability

trgiangdo:
Follow up the outdated PR https://github.com/Avaiga/taipy/pull/609 from a contributor

### DIFF
--- a/taipy/gui_core/_adapters.py
+++ b/taipy/gui_core/_adapters.py
@@ -115,8 +115,8 @@ class _GuiCoreScenarioDagAdapter(_TaipyBase):
         data = super().get()
         if isinstance(data, Scenario):
             try:
-                if core_get(data.id):
-                    dag = data._get_dag()
+                if scenario := core_get(data.id):
+                    dag = scenario._get_dag()
                     nodes = dict()
                     for id, node in dag.nodes.items():
                         entityType = _GuiCoreScenarioDagAdapter.get_entity_type(node)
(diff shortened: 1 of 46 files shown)


taipy/gui_core/_adapters.py
```diff
@@ -116,8 +115,7 @@ def get(self):
         data = super().get()
         if isinstance(data, Scenario):
             try:
-                scenario = core_get(data.id)
-                if scenario:
+                if core_get(data.id):
```
FredLL-Avaiga: looks like this should be
if scenario := core_get(data.id)
    dag = scenario._get_dag()

FredLL-Avaiga (dismissed):
well done, tx @trgiangdo

toan-quach (dismissed):
lgtm

jrobinAV:
It does not pass the linter.

taipy/gui_core/_adapters.py
trgiangdo: I remove the `scenario :=` because we do not use it in the code below.
You're sure about this?

FredLL-Avaiga: that was the mistake : to not use it
i'd rather assign it *and* use it :-)

jrobinAV: 😄

FredLL-Avaiga (approved):

PR #1013, opened 2024-03-19
Decouple preview bundle

dinhlongviolin1:
Preview file for preview pages (without cookies handling)

### DIFF
--- a/frontend/taipy-gui/base/src/index-preview.ts
+++ b/frontend/taipy-gui/base/src/index-preview.ts
@@ -1,0 +1,6 @@
+import { TaipyApp, createApp, OnChangeHandler, OnInitHandler } from "./app";
+import { ModuleData } from "./dataManager";
+
+export default TaipyApp;
+export { TaipyApp, createApp };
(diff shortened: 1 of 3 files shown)


frontend/taipy-gui/base/src/index-preview.ts
```diff
@@ -0,0 +1,6 @@
+import { TaipyApp, createApp, OnChangeHandler, OnInitHandler } from "./app";
+import { ModuleData } from "./dataManager";
+
+export default TaipyApp;
+export { TaipyApp, createApp };
```
FredLL-Avaiga, line 5: are you exporting TaipyApp twice ?

FredLL-Avaiga (dismissed):

frontend/taipy-gui/base/src/index-preview.ts
dinhlongviolin1: Yes it is right now. Thanks for spotting it. I am coming up with an update that allows chunking

FredLL-Avaiga (approved):

PR #1027, opened 2024-03-21
Add name to shared bundle for easier import

dinhlongviolin1:
chunk created by optimization will not be imported by the default entry --> an easier name is needed for 3rd party to import the shared chunk


FabienLelaquais (approved):
Where is this name used again?

dinhlongviolin1:
> Where is this name used again?

it will be used for the shared chunk (shared.taipy-gui-base.js)

FredLL-Avaiga (approved):
good name :-)

PR #1035, opened 2024-03-23
Fix/#497 - Unique section init should not use default values

### LINKED ISSUE #497: BUG -  `CoreSection` init should not use default value
Ok, but I get inspired from what is done in `CoreSection`: https://github.com/Avaiga/taipy-core/blob/develop/src/taipy/core/config/core_section.py#L100
which does pretty much the same...

_Originally posted by @gmarabout in https://github.com/Avaiga/taipy-enterprise/pull/265#discussion_r1406248345_

trgiangdo:
Resolves #497 

In this PR:
- Remove default value usages from CoreSection.__init__
- Refactor the JobConfig class: merge `JobConfig._config` with the `_properties`
- Check the job execution mode in the Config checker

### DIFF
--- a/taipy/core/config/job_config.py
+++ b/taipy/core/config/job_config.py
@@ -17,8 +17,6 @@ from taipy.config._config import _Config
 from taipy.config.common._template_handler import _TemplateHandler as _tpl
 from taipy.config.unique_section import UniqueSection
 
-from ..exceptions.exceptions import ModeNotAvailable
-
 
 class JobConfig(UniqueSection):
     """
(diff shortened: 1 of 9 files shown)


taipy/core/config/job_config.py
```diff
@@ -37,32 +35,34 @@ class JobConfig(UniqueSection):
     _DEVELOPMENT_MODE = "development"
     _DEFAULT_MODE = _DEVELOPMENT_MODE
     _DEFAULT_MAX_NB_OF_WORKERS = 2
-    _MODES = [_STANDALONE_MODE, _DEVELOPMENT_MODE]
+    _MODES = [_DEVELOPMENT_MODE, _STANDALONE_MODE]
```
jrobinAV, line 38: Any specific reason?

jrobinAV (approved):

toan-quach (approved):
lgtm

taipy/core/config/job_config.py
trgiangdo: No, just want to make the default value as the first value of the list 😅

PR #1052, opened 2024-03-27
Feature/#949 - Refactor Taipy CLI

### LINKED ISSUE #949: BUG-Inputting wrong kwargs in CLI does not return warnings
I wanted to create a scenario-management template project using the CLI so I typed:
```bash
taipy create --teamplate scenario-management
```
Unfortunately, I wrote a typo, and wrote "teamplate" instead of "template" and I got the wrong template

Taipy's CLI should return warnings when inputting non-existing kwargs

trgiangdo:
Resolves https://github.com/Avaiga/taipy/issues/949

In this PR:
- Refactoring Taipy CLI. Now all _CLI classes inherits from `_AbstractCLI`.
- Except from `_GuiCLI` and `_CoreCLI`, all other CLI with specific arguments will raise an error if the argument is not supported. If their is a typo that similar enough to an argument, there will be a suggestion. Something like:
```
$ taipy migrate --repository-type filesystem --slip-backup 
usage: taipy migrate [-h] [--repository-type REPOSITORY_TYPE [REPOSITORY_TYPE ...]] [--skip-backup] [--restore] [--remove-backup]

options:
  -h, --help            show this help message and exit
  --repository-type REPOSITORY_TYPE [REPOSITORY_TYPE ...]
                        The type of repository to migrate. If filesystem or sql, a path to the database folder/.sqlite file should be
                        informed. In case of mongo host, port, user and password must be informed, if left empty it is assumed
                        default values
  --skip-backup         Skip the backup of entities before migration.
  --restore             Restore the migration of entities from backup folder.
  --remove-backup       Remove the backup of entities. Only use this option if the migration was successful.
[2024-03-27 19:46:48.675][Taipy][ERROR] Unknown arguments: --slip-backup. Did you mean: --skip-backup?
[2024-03-27 19:46:48.675][Taipy][ERROR]  Please refer to the help message above for more information.
```
Here, we also print the help message for the user to modify the command.

### DIFF
--- a/tests/templates/test_template_cli.py
+++ b/tests/templates/test_template_cli.py
@@ -21,3 +21,11 @@ def test_create_cli_with_wrong_arguments(caplog):
         with pytest.raises(SystemExit):
             _entrypoint()
         assert "Unknown arguments: --teamplaet. Did you mean: --template?" in caplog.text
+
+
+def test_create_cli_with_unsupported_template(capsys):
+    with patch("sys.argv", ["prog", "create", "--template", "not-a-template"]):
+        with pytest.raises(SystemExit):
+            _entrypoint()
+        _, err = capsys.readouterr()
+        assert "invalid choice: 'not-a-template'" in err
--- a/tests/templates/test_template_cli.py
+++ b/tests/templates/test_template_cli.py
@@ -8,3 +8,16 @@
 # Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
+
+from unittest.mock import patch
+
+import pytest
+
+from taipy._entrypoint import _entrypoint
+
+
+def test_create_cli_with_wrong_arguments(caplog):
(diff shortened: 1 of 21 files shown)


tests/templates/test_template_cli.py
```diff
@@ -0,0 +1,23 @@
...
+import pytest
+
+from taipy._entrypoint import _entrypoint
+
+
+def test_create_cli_with_wrong_arguments(caplog):
+    with patch("sys.argv", ["prog", "create", "--teamplaet", "default"]):
+        with pytest.raises(SystemExit):
+            _entrypoint()
+        assert "Unknown arguments: --teamplaet. Did you mean: --template?" in caplog.text
```
jrobinAV, line 23: Can you add a test where the name of the template does not exist?
```
taipy create --template lkjdslklj
```

jrobinAV: It should raise an error.

jrobinAV (dismissed):

tests/templates/test_template_cli.py
trgiangdo: You're right. Good idea

jrobinAV:
@trgiangdo Before merging it, can you check it does not break enterprise edition?

trgiangdo:
it does break 🤣 there is an opening PR to fix that

jrobinAV (approved):

PR #1053, opened 2024-03-27
Fix - Simplify core cli tests

trgiangdo:
This PR simplifies the core cli tests. These tests should not submit jobs in standalone mode.

### DIFF
--- a/tests/core/test_core_cli.py
+++ b/tests/core/test_core_cli.py
@@ -27,7 +27,6 @@ from taipy.core.job._job_manager import _JobManager
 from taipy.core.scenario._scenario_manager import _ScenarioManager
 from taipy.core.sequence._sequence_manager import _SequenceManager
 from taipy.core.task._task_manager import _TaskManager
-from tests.core.utils import assert_true_after_time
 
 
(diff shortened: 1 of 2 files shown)


tests/core/test_core_cli.py
```diff
@@ -457,54 +456,42 @@ def test_force_override_production_version():
         core.stop()
 
 
-@pytest.mark.standalone
 def test_modified_job_configuration_dont_block_application_run(caplog, init_config):
-    scenario_config = config_scenario()
+    _ = config_scenario()
```
toan-quach, line 460: I think we can just keep it even simpler with 
```suggestion
    config_scenario()
```

toan-quach (approved):
lgtm

tests/core/test_core_cli.py
trgiangdo: this will trigger a ruff error where the return of `config_sceario()` is not assigned to any variable. I agree that this looks weird

toan-quach: I see :D that's ok to me too

jrobinAV (approved):

PR #1056, opened 2024-03-28
feature/#957 attempt to clean dn inheritance

### LINKED ISSUE #957: Clean DN inheritance
The inheritance mechanism for all data nodes would benefit from some cleaning.

Suggestions:

- The `_AbstractTabularDataNode` is not a data node but a mixin. The name is misleading. It is more like a mixin.
- Same for `_AbstractFileDataNode`.
- However, the `_abstract_sql` is a data node sub-class. Why this inconsistency?
- The `FilterDataNode` is not a data node sub-class nor a mixin. It is just a delegated class. I guess it should be renamed to `DataNodeFilter`.
- The exposed type code is duplicated in many data node implementations. I believe we can factorize the code in the `_AbstractTabularDataNode` class since the exposed type is for tabular data.

toan-quach:
#957

### DIFF
--- a/taipy/core/data/_abstract_file.py
+++ b/taipy/core/data/_abstract_file.py
@@ -13,7 +13,7 @@ import pathlib
 import shutil
 
 
-class _AbstractFileDataNode(object):
+class _AbstractFileDataNodeMixin(object):
     """Abstract base class for data node implementations (CSVDataNode, ParquetDataNode, ExcelDataNode,
     PickleDataNode and JSONDataNode) that are file based."""
 
--- a/taipy/core/data/_abstract_file.py
+++ b/taipy/core/data/_abstract_file.py
@@ -13,9 +13,9 @@ import pathlib
 import shutil
 
 
-class _AbstractFileDataNodeMixin(object):
-    """Abstract base class for data node implementations (CSVDataNode, ParquetDataNode, ExcelDataNode,
-    PickleDataNode and JSONDataNode) that are file based."""
+class _FileDataNodeMixin(object):
+    """Mixin class designed to handle file-based data nodes
+    (CSVDataNode, ParquetDataNode, ExcelDataNode, PickleDataNode, JSONDataNode, etc.)."""
 
     __EXTENSION_MAP = {"csv": "csv", "excel": "xlsx", "parquet": "parquet", "pickle": "p", "json": "json"}
 
--- a/taipy/core/data/_abstract_tabular.py
+++ b/taipy/core/data/_abstract_tabular.py
@@ -17,9 +17,9 @@ import pandas as pd
 from ..exceptions.exceptions import InvalidExposedType
 
 
-class _AbstractTabularDataNodeMixin(object):
-    """Abstract base class for tabular data node implementations (CSVDataNode, ParquetDataNode, ExcelDataNode,
-    SQLTableDataNode and SQLDataNode) that are tabular representable."""
+class _TabularDataNodeMixin(object):
+    """Mixin class designed to handle tabular representable data nodes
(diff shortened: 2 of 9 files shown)


taipy/core/data/_abstract_file.py
```diff
@@ -13,7 +13,7 @@
 import shutil
 
 
-class _AbstractFileDataNode(object):
+class _AbstractFileDataNodeMixin(object):
```
jrobinAV: At the end, it is not abstract and it is not a data node.

```suggestion
class _FileMixin(object):
```

taipy/core/data/_abstract_file.py
```diff
@@ -13,7 +13,7 @@
...
     """Abstract base class for data node implementations (CSVDataNode, ParquetDataNode, ExcelDataNode,
     PickleDataNode and JSONDataNode) that are file based."""
```
jrobinAV:
```suggestion
    """Mixin class designed to handle file-based data nodes (CSVDataNode, ParquetDataNode, ExcelDataNode, PickleDataNode, JSONDataNode, etc.)."""
```

taipy/core/data/_abstract_tabular.py
```diff
@@ -17,7 +17,7 @@
 from ..exceptions.exceptions import InvalidExposedType
 
 
-class _AbstractTabularDataNode(object):
+class _AbstractTabularDataNodeMixin(object):
```
jrobinAV:
```suggestion
class _TabularMixin(object):
```

taipy/core/data/_abstract_tabular.py
```diff
@@ -17,7 +17,7 @@
...
     """Abstract base class for tabular data node implementations (CSVDataNode, ParquetDataNode, ExcelDataNode,
     SQLTableDataNode and SQLDataNode) that are tabular representable."""
```
jrobinAV: We can change the docstring to make it up-to-date regarding the new name, no?

taipy/core/data/_abstract_tabular.py
```diff
@@ -53,9 +53,20 @@ def _convert_data_to_dataframe(self, exposed_type: Any, data: Any) -> Union[pd.D
...
+    @classmethod
+    def _get_valid_exposed_type(cls, properties: Dict):
+        if (
+            cls._EXPOSED_TYPE_PROPERTY not in properties.keys()
+            or properties[cls._EXPOSED_TYPE_PROPERTY] == cls._EXPOSED_TYPE_MODIN
+        ):
+            # Default exposed type is pandas
+            # Deprecated modin exposed type in favor of pandas since 3.1.0
+            return cls._EXPOSED_TYPE_PANDAS
+        return properties[cls._EXPOSED_TYPE_PROPERTY]
```
jrobinAV, line 65: Nice!

taipy/core/data/_abstract_file.py
FabienLelaquais: On the other hand... _FileMixin is really really vague...
But local to core/data... I'm good with that

FabienLelaquais (dismissed):
To me, after the renaming as proposed by JR, that's good.
It IS a mixin indeed, and not an abstract class...

taipy/core/data/_abstract_tabular.py
toan-quach: about time 🤣

taipy/core/data/_abstract_file.py
toan-quach: how about `_FileDataNodeMixin`?

taipy/core/data/_abstract_tabular.py
toan-quach: I'd lean on `_TabularDataNodeMixin` since it's exclusively for dn use only. What do you think?

jrobinAV (approved):

PR #1078, opened 2024-04-02
Refactor - All file-based datanode should have consistent default data writing behavior

### LINKED ISSUE #535: Generalize default_data attribute for all predefined data nodes
Today, a default data attribute in Data Node config is implemented for csv, excel, in memory, json, parquet, pickle storage types.
We want to make it generic for all data node types, in a similar way.

- [ ] Implement the data writing with default data when creating the data node for the first time.
- [ ] Unit tests
- [ ] Release notes
- [ ] User manual

trgiangdo:
In this PR:
- Refactor the `_FileDataNodeMixin`
- Fix clean all data entities doesn't delete all generated data.
- Other minor code format fixes


taipy/core/data/csv.py
```diff
@@ -100,11 +96,17 @@ def __init__(
         properties[self._EXPOSED_TYPE_PROPERTY] = _TabularDataNodeMixin._get_valid_exposed_type(properties)
         self._check_exposed_type(properties[self._EXPOSED_TYPE_PROPERTY])
 
+        default_value = properties.pop(self._DEFAULT_DATA_KEY, None)
+        _FileDataNodeMixin.__init__(self, properties)
+        _TabularDataNodeMixin.__init__(self, **properties)
+        properties[self._IS_GENERATED_KEY] = self._is_generated
+        properties[self._PATH_KEY] = self._path
```
jrobinAV: Should this be in FileMixin?

taipy/core/data/json.py
```diff
@@ -62,9 +60,9 @@ class JSONDataNode(DataNode, _FileDataNodeMixin):
     """
 
     __STORAGE_TYPE = "json"
-    __DEFAULT_DATA_KEY = "default_data"
-    __DEFAULT_PATH_KEY = "default_path"
-    __PATH_KEY = "path"
+    _DEFAULT_DATA_KEY = "default_data"
+    _DEFAULT_PATH_KEY = "default_path"
+    _PATH_KEY = "path"
```
jrobinAV: Is it necessary?

taipy/core/data/json.py
```diff
@@ -159,7 +143,8 @@ def path(self):
     @path.setter
     def path(self, value):
         self._path = value
```
jrobinAV: COuld path be defined in the FileMixin?

tests/core/_orchestrator/test_orchestrator__submit_task.py
```diff
@@ -46,14 +47,15 @@ def test_submit_task_development_mode():
...
-    submit_time = datetime.now()
+    submit_time = datetime.now() + timedelta(seconds=1)  # +1 to ensure the edit time of dn_0 is before the submit time
     with freezegun.freeze_time(submit_time):
         submission = orchestrator.submit_task(
             scenario.t1, no_of_retry=10, log=True, log_file="file_path"
         )  # t1 is executed directly in development mode
         job = submission.jobs[0]
 
     # task output should have been written
+    assert scenario.dn_0.last_edit_date < submit_time
```
jrobinAV: This assertion does not belong to the test_submit_task function. It should be part of the create scenario test function. No?

tests/core/data/test_data_manager.py
```diff
@@ -77,7 +78,9 @@ def test_create_and_get_csv_data_node(self):
         assert _DataManager._get(csv_dn.id).job_ids == csv_dn.job_ids
         assert not _DataManager._get(csv_dn.id).is_ready_for_reading
         assert _DataManager._get(csv_dn.id).is_ready_for_reading == csv_dn.is_ready_for_reading
-        assert len(_DataManager._get(csv_dn.id).properties) == 4
+        assert (
+            len(_DataManager._get(csv_dn.id).properties) == 5
+        )  # path, encoding, has_header, exposed_type, is_generated
         assert _DataManager._get(csv_dn.id).properties.get("path") == "bar"
```
jrobinAV, line 84: Please add a test on the 'is_generated' property value to avoid regression

tests/core/data/test_excel_data_node.py
```diff
@@ -132,12 +133,13 @@ def test_set_path(self):
...
         [
-            ({}, False),
             ({"default_data": {"a": ["foo", "bar"]}}, True),
+            ({}, False),
         ],
     )
     def test_create_with_default_data(self, properties, exists):
-        dn = ExcelDataNode("foo", Scope.SCENARIO, DataNodeId("dn_id"), properties=properties)
+        dn = ExcelDataNode("foo", Scope.SCENARIO, DataNodeId(f"dn_id_{uuid.uuid4()}"), properties=properties)
+        assert dn.path == Config.core.storage_folder + "excels/" + dn.id + ".xlsx"
```
jrobinAV: This is not compatible with windows. IS it?

tests/core/data/test_csv_data_node.py
```diff
@@ -102,7 +103,8 @@ def test_new_csv_data_node_with_existing_file_is_ready_for_reading(self):
         ],
     )
     def test_create_with_default_data(self, properties, exists):
-        dn = CSVDataNode("foo", Scope.SCENARIO, DataNodeId("dn_id"), properties=properties)
+        dn = CSVDataNode("foo", Scope.SCENARIO, DataNodeId(f"dn_id_{uuid.uuid4()}"), properties=properties)
+        assert dn.path == Config.core.storage_folder + "csvs/" + dn.id + ".csv"
```
jrobinAV: IS this windows compatible?

tests/core/test_taipy.py
```diff
@@ -715,10 +715,10 @@ def test_export_scenario_filesystem(self):
         assert sorted(os.listdir("./tmp/exp_scenario_1/jobs")) == sorted([f"{job_2.id}.json"])
         assert sorted(os.listdir("./tmp/exp_scenario_1/cycles")) == sorted([f"{scenario_2.cycle.id}.json"])
 
-        with pytest.raises(InvalidExportPath):
-            tp.export_scenario(scenario_2.id, Config.core.taipy_storage_folder)
+        # with pytest.raises(InvalidExportPath):
+        #     tp.export_scenario(scenario_2.id, Config.core.taipy_storage_folder)
 
-        shutil.rmtree("./tmp", ignore_errors=True)
+        # shutil.rmtree("./tmp", ignore_errors=True)
```
jrobinAV: Why this change?

taipy/core/data/csv.py
trgiangdo: It should, yes. I forgot a dict is passed by reference to the function

tests/core/test_taipy.py
trgiangdo: Sorry, I pushed this by mistake

tests/core/_orchestrator/test_orchestrator__submit_task.py
trgiangdo: you're right

taipy/core/data/_abstract_file.py
```diff
@@ -19,16 +30,73 @@ class _FileDataNodeMixin(object):
...
+    def _write(self, data):
+        raise NotImplementedError
 
+    def _build_path(self, storage_type) -> str:
         folder = f"{storage_type}s"
         dir_path = pathlib.Path(Config.core.storage_folder) / folder
         if not dir_path.exists():
             dir_path.mkdir(parents=True, exist_ok=True)
-        return dir_path / f"{self.id}.{self.__EXTENSION_MAP.get(storage_type)}"
+        return str(dir_path / f"{self.id}.{self.__EXTENSION_MAP.get(storage_type)}")  # type: ignore[attr-defined]
```
jrobinAV: Why do we need a type ignore?

taipy/core/data/_abstract_file.py
```diff
@@ -19,16 +30,73 @@ class _FileDataNodeMixin(object):
...
+        self.properties[self._PATH_KEY] = value
+        self.properties[self._IS_GENERATED_KEY] = False
+
+    @classmethod
+    def storage_type(cls) -> str:
+        raise NotImplementedError
+
+    @abstractmethod
+    def _write(self, data):
+        raise NotImplementedError
```
jrobinAV: These are inherited from the Datanode class. They should not be in a Mixin. If it is needed because of the write default data, that means either the write default data should be in a separate class or it should not be a Mixin but an implementation of the datanode class.

What do you think?

My two cents.

taipy/core/data/_abstract_file.py
```diff
@@ -19,16 +30,73 @@ class _FileDataNodeMixin(object):
...
+            self._last_edit_date = datetime.now()
+            self._edits.append(  # type: ignore[attr-defined]
+                Edit(
+                    {
+                        "timestamp": self._last_edit_date,
+                        "writer_identifier": "TAIPY",
+                        "comments": "Default data written.",
+                    }
+                )
+            )
```
jrobinAV: There is nothing specific to file data nodes in this part. Could we move it to the data node class?

taipy/core/data/_abstract_file.py
jrobinAV, line 1: The name of the file does not match the class name. 
Either `FileMixin` or `AbstractFileDatanode` depending on its inheritance from Datanode class or not.

taipy/core/data/csv.py
```diff
@@ -116,56 +115,25 @@ def __init__(
...
-                    {
-                        "timestamp": self._last_edit_date,
-                        "writer_identifier": "TAIPY",
-                        "comments": "Default data written.",
-                    }
-                )
-            )
-        if not self._last_edit_date and isfile(self._path):
-            self._last_edit_date = datetime.now()
+        self._write_default_data(default_value)
```
jrobinAV, line 119: Same here. Nothing specific to csv data node. It should be moved to the appropriate class.

taipy/core/data/_abstract_file.py
trgiangdo: Somehow when editing this line, the `self.id` raise the mypy's `attr-defined` error. The error was there before, but not detected

taipy/core/data/_abstract_file.py
trgiangdo: I added these to avoid the `type: ignore`. But yes, these should not be here

taipy/core/data/_abstract_file.py
trgiangdo: If we decide to have `default_data` on all data node types, we can do that.

But if we decide to have the `default_data` on file data nodes only, it should be here

taipy/core/data/_abstract_file.py
jrobinAV: ok thx.

taipy/core/data/_abstract_file.py
jrobinAV: @FlorianJacta Seems to agree with you. default_data should not be generalized to all data node types.

I don't like it when it is asymmetric, but I don't have a functional or technical reason to generalize it.

Let's forget about this. We can close ticket #535 as not planned and resolve this comment.

taipy/core/data/_abstract_file.py
trgiangdo: I renamed the file to `_file_datanode_mixin.py` to match the class name. The same for the `_tabular_datanode_mixin.py`

jrobinAV:
It looks fine. I want to wait for the tests to pass. Unfortunately, they don't pass for the numpy dependency error.

jrobinAV (dismissed):

jrobinAV:
Do we need to update the documentation and particularly the release notes?

toan-quach (dismissed):

trgiangdo:
> Do we need to update the documentation and particularly the release notes?

No we don't. The behavior is nearly identical. We're mainly refactoring the code in this PR.

tests/core/_orchestrator/test_orchestrator__submit.py
```diff
@@ -54,18 +53,16 @@ def test_submit_scenario_development_mode():
...
 
-    submit_time = datetime.now() + timedelta(seconds=1)  # +1 to ensure the edit time of dn_0 is before the submit time
-    with freezegun.freeze_time(submit_time):
-        submission = orchestrator.submit(
-            scenario, no_of_retry=10, log=True, log_file="file_path"
-        )  # scenario is executed directly in development mode
-        jobs = submission.jobs
+    submit_time = datetime.now()
+    submission = orchestrator.submit(
+        scenario, no_of_retry=10, log=True, log_file="file_path"
```
jrobinAV: what are these properties for? If we want to test the custom properties mechanism, we should add some assertions, and it should be in a different test function.

tests/core/_orchestrator/test_orchestrator__submit.py
```diff
@@ -54,18 +53,16 @@ def test_submit_scenario_development_mode():
...
+    jobs = submission.jobs
 
-    # data nodes should have been written (except the input dn_0)
+    # the edit time of dn_0 is before the submit time
     assert scenario.dn_0.last_edit_date < submit_time
-    assert scenario.dn_1.last_edit_date == submit_time
-    assert scenario.dn_2.last_edit_date == submit_time
-    assert scenario.dn_3.last_edit_date == submit_time
+    # data nodes should have been written after the submit_time
+    assert submit_time < scenario.dn_1.last_edit_date < scenario.dn_2.last_edit_date < scenario.dn_3.last_edit_date
```
jrobinAV: I would not assert that. At least we should assert with `<=` instead of `<`.

tests/core/_orchestrator/test_orchestrator__submit.py
jrobinAV, line 1: Two global remarks:

1. The tests on the various timestamps are dangerous to me. Regarding the accuracy of the time measurement, we cannot ensure a strict inequality. First, I would replace all `<`with `<=`.

2. since the data nodes are file-based, the various timestamps we compare are not computed similarly. Some are computed by the Python interpreter through Taipy code, while others are computed by the so itself (when the file is written). I believe this depends on some external components we don't control: OS, hard drive drivers, etc... 

I am not comfortable with this, but I don't know what to propose. Maybe the test needs to be more 'unitary'. The scope of this long test function needs to be clarified for me. Before removing the `freezegun` code, we tested the `creation_date` and  `last_edit_date ` attributes to ensure they are populated with the current time. Now, with your new code, you also test the correct ordering of the timestamps. 
Has it already been tested somewhere else? 
How to compare Python timestamps with file timestamps?

tests/core/_orchestrator/test_orchestrator__submit.py
trgiangdo: Idk. It was there before

jrobinAV: They are used below. Let s keep them. Thx.

jrobinAV (approved):

PR #1079, opened 2024-04-02
Better document default_data for pickle datanodes

jrobinAV:


jrobinAV:
Oh... When the `Update config.pyi` action runs, and auto commits something, all the test actions are stopped and they don't run again.

PR #1086, opened 2024-04-04
bug/#746 new notifier mechanism for submittable status

### LINKED ISSUE #746: "Cannot submit scenario" in scenario visual element without explanation
The scenario visual element can display a grey send icon with "Cannot submit scenario" without an explanation:
image

This bug gets triggered in the [Scenario Management Overview](https://docs.taipy.io/en/develop/knowledge_base/tutorials/scenario_management_overview/) tutorial when the user creates a new scenario and tries to run it

**How to reproduce:**

1. Use the code provided in the Scenario Management Overview tutorial:

```python
from taipy import Config
import taipy as tp
import pandas as pd
import datetime as dt


data = pd.read_csv("https://raw.githubusercontent.com/Avaiga/taipy-getting-started-core/develop/src/daily-min-temperatures.csv")


def predict(historical_temperature: pd.DataFrame, date_to_forecast: str) -> float:
    print(f"Running baseline...")
    historical_temperature['Date'] = pd.to_datetime(historical_temperature['Date'])
    historical_same_day = historical_temperature.loc[
        (historical_temperature['Date'].dt.day == date_to_forecast.day) &
        (historical_temperature['Date'].dt.month == date_to_forecast.month)
    ]
    return historical_same_day['Temp'].mean()

# Configuration of Data Nodes
historical_temperature_cfg = Config.configure_data_node("historical_temperature")
date_to_forecast_cfg = Config.configure_data_node("date_to_forecast")
predictions_cfg = Config.configure_data_node("predictions")

# Configuration of tasks
predictions_cfg = Config.configure_task("predict",
                                        predict,
                                        [historical_temperature_cfg, date_to_forecast_cfg],
                                        predictions_cfg)

# Configuration of scenario
scenario_cfg = Config.configure_scenario(id="my_scenario", task_configs=[predictions_cfg])

Config.export('config.toml')

if __name__ == '__main__':
    # Run of the Core
    tp.Core().run()

    # Creation of the scenario and execution
    scenario = tp.create_scenario(scenario_cfg)
    scenario.historical_temperature.write(data)
    scenario.date_to_forecast.write(dt.datetime.now())
    tp.submit(scenario)

    print("Value at the end of task", scenario.predictions.read())

    def save(state):
        state.scenario.historical_temperature.write(data)
        state.scenario.date_to_forecast.write(state.date)
        tp.gui.notify(state, "s", "Saved! Ready to submit")

    date = None
    scenario_md = """
<|{scenario}|scenario_selector|>

Put a Date
<|{date}|date|on_change=save|active={scenario}|>

Run the scenario
<|{scenario}|scenario|>
<|{scenario}|scenario_dag|>

View all the information on your prediction here
<|{scenario.predictions if scenario else None}|data_node|>
"""

    tp.Gui(scenario_md).run()
```

2. Run the app and create a new scenario
image

3. The scenario visual element won't let you run the new scenario
image

toan-quach:
#746


taipy/core/data/data_node.py
```diff
@@ -29,12 +30,35 @@
...
+def _update_parent_submittable_cache_on_edit_in_progress(fct):
+    @functools.wraps(fct)
+    def _do_update_parent_submittable_cache_on_edit_in_progress(dn, *args, **kwargs):
+        fct(dn, *args, **kwargs)
+        if not dn._edit_in_progress:
+            SubmittableStatusCache.remove(dn.id)
+        else:
+            from ..scenario.scenario import Scenario
+            from ..sequence.sequence import Sequence
+            from ..task.task import Task
```
trgiangdo: This will create a recursive import error if put outside?

taipy/core/data/data_node.py
```diff
@@ -167,7 +191,10 @@ def last_edit_date(self):
     @last_edit_date.setter  # type: ignore
     @_self_setter(_MANAGER_NAME)
     def last_edit_date(self, val):
+        prev_last_edit_date = self._last_edit_date
         self._last_edit_date = val
+        if prev_last_edit_date is None and self._last_edit_date:
+            SubmittableStatusCache.remove(self.id)
```
trgiangdo: what does this mean?

taipy/core/data/data_node.py
```diff
@@ -29,12 +30,35 @@
...
+            from ..sequence.sequence import Sequence
+            from ..task.task import Task
+
+            parent_entities = dn.get_parents()
+            for scenario_parent in parent_entities.get(Scenario._MANAGER_NAME, []):
+                SubmittableStatusCache.add(scenario_parent.id, dn.id)
+            for sequence_parent in parent_entities.get(Sequence._MANAGER_NAME, []):
+                SubmittableStatusCache.add(sequence_parent.id, dn.id)
+            for task_parent in parent_entities.get(Task._MANAGER_NAME, []):
+                SubmittableStatusCache.add(task_parent.id, dn.id)
```
trgiangdo: I believe we can do this with 2 nested loop so we can avoid the import.

However, I'm not sure about it, since it can make the code less readable.

trgiangdo (commented):
I think the PR looks good. I have a few questions in the comment.

Also, from the changes, the only reason a scenario can't be run is that a data node is not ready to be read?

taipy/core/_entity/submittable.py
```diff
@@ -86,7 +88,13 @@ def is_ready_to_run(self) -> bool:
...
             True if the given entity is ready to be run. False otherwise.
         """
-        return all(dn.is_ready_for_reading for dn in self.get_inputs())
+        if self._submittable_id not in SubmittableStatusCache.submittable_id_datanodes:
+            for dn in self.get_inputs():
+                if not dn.is_ready_for_reading:
+                    SubmittableStatusCache.add(self._submittable_id, dn.id)
+        if len(SubmittableStatusCache.submittable_id_datanodes[self._submittable_id]) == 0:
+            return True
+        return False
```
jrobinAV: What about delegating more by targeting the following code:

```suggestion
        if self._submittable_id not in SubmittableStatusCache:
            SubmittableStatusCache.add(self._submittable_id, self.get_inputs())
        return SubmittableStatusCache.is_ready_to_run(self._submittable_id)
```

I did another suggestion in the `SubmittableStatusCache` class to match this usage.

taipy/core/notification/dn_scenario_cache.py
```diff
@@ -0,0 +1,35 @@
...
+if TYPE_CHECKING:
+    from ..data.data_node import DataNode, DataNodeId
+
+
+class SubmittableStatusCache:
+    datanode_id_submittables: Dict["DataNodeId", Set[str]] = defaultdict(lambda: set())
+    submittable_id_datanodes: Dict[str, Set["DataNodeId"]] = defaultdict(lambda: set())
+
+    @classmethod
+    def add(cls, entity_id: str, datanode_id: "DataNodeId"):
```
jrobinAV: What about adding more responsibility to this class. Instead of only storing the result. we can also compute the value.

```suggestion
    @classmethod
    def add(cls, entity_id: str, datanodes: Iterable[DataNode]):
        for dn in datanodes if not dn.is_ready_for_reading::
            clas._add(entity_id, dn.id)
    
    def _add(cls, entity_id: str, datanode_id: "DataNodeId"):
```

taipy/core/notification/dn_scenario_cache.py
```diff
@@ -0,0 +1,35 @@
...
+# specific language governing permissions and limitations under the License.
+
+from collections import defaultdict
+from typing import TYPE_CHECKING, Dict, Set
+
+if TYPE_CHECKING:
+    from ..data.data_node import DataNode, DataNodeId
+
+
+class SubmittableStatusCache:
```
jrobinAV: Ultimately, if we give more responsibility to this class, it computes, stores, and returns the `is_ready_to_run`property.

What about calling it `_ReadyToRunManager`?

I believe it should be prefixed by an `_` so it is not exposed.

taipy/core/notification/dn_scenario_cache.py
```diff
@@ -0,0 +1,35 @@
...
+    datanode_id_submittables: Dict["DataNodeId", Set[str]] = defaultdict(lambda: set())
+    submittable_id_datanodes: Dict[str, Set["DataNodeId"]] = defaultdict(lambda: set())
+
+    @classmethod
+    def add(cls, entity_id: str, datanode_id: "DataNodeId"):
+        cls.datanode_id_submittables[datanode_id].add(entity_id)
+        cls.submittable_id_datanodes[entity_id].add(datanode_id)  # type: ignore
```
jrobinAV: We need to add some comments here to explain what is the content of the two dict.

taipy/core/notification/dn_scenario_cache.py
```diff
@@ -0,0 +1,35 @@
...
+
```
jrobinAV:
```suggestion

     @classmethod
     def is_ready_to_run(cls, entity_id):
         return len(cls.submittable_id_datanodes.get(entity_id, [])) == 0
```

taipy/core/data/data_node.py
jrobinAV: I believe this does not work.
We don't know if we need to remove the dn from the cache here, since we don't know the value of `_edits_in_progress`.

taipy/core/data/data_node.py
```diff
@@ -29,12 +30,35 @@
...
 from .data_node_id import DataNodeId, Edit
 from .operator import JoinOperator
 
 
+def _update_parent_submittable_cache_on_edit_in_progress(fct):
+    @functools.wraps(fct)
+    def _do_update_parent_submittable_cache_on_edit_in_progress(dn, *args, **kwargs):
+        fct(dn, *args, **kwargs)
+        if not dn._edit_in_progress:
+            SubmittableStatusCache.remove(dn.id)
```
jrobinAV: Same here, we don't know if we should remove the dn from the cache or not. It depends on the value of `_last_edit_date`.

jrobinAV (commented):
@toan-quach I might have a proposal to share with you. I can show it to you in a call.

taipy/core/data/_data_model.py
```diff
@@ -69,7 +69,7 @@ def from_dict(data: Dict[str, Any]):
             scope=Scope._from_repr(data["scope"]),
             storage_type=data["storage_type"],
             owner_id=data.get("owner_id"),
-            parent_ids=data.get("parent_ids", []),
+            parent_ids=_BaseModel._deserialize_attribute(data.get("parent_ids", [])),
```
jrobinAV: Can we isolate this bug fix in a separate PR. 
The topic is big, so I would like to keep the PRs as small as possible and split them as much as possible.

I would also like to add a non regression unit test about this fix.

tests/core/sequence/test_sequence.py
```diff
@@ -412,27 +412,43 @@ def test_get_inputs():
...
+    scenario_id = "SCENARIO_scenario_id"
+    task_1_id, task_2_id, task_3_id, task_4_id = (
+        TaskId("TASK_t1"),
+        TaskId("TASK_t2"),
+        TaskId("TASK_t3"),
+        TaskId("TASK_t4"),
+    )
+    data_node_1 = PickleDataNode("foo", Scope.SCENARIO, "s1", parent_ids={task_1_id}, properties={"default_data": 1})
+    data_node_2 = PickleDataNode("bar", Scope.SCENARIO, "s2", parent_ids={task_1_id}, properties={"default_data": 2})
+    data_node_4 = PickleDataNode(
```
jrobinAV: We can take the opportunity to rename the data nodes (1, 2, 3, 4, 5, and 6) instead of (1, 2, 4, 5, 6, and 7)

taipy/core/data/data_node.py
toan-quach: yep, in Scenario, we also import DataNode for additional datanode, same for task. Maybe Sequence won't have this issue but for cleanliness I decided it's better to keep things together

taipy/core/data/data_node.py
toan-quach: if previously, the data node is not written -> prev_last_edit_date is None, and now it's been written -> last_edit_date is no longer None, -> Then the entity is now submittable as the entity is not submittable when last_edit_date is None (or not written ever)

taipy/core/_entity/submittable.py
toan-quach: I don't think it's good considering we're adding all input DNs without checking?

toan-quach: I'm proposing a new design, will let you know once pushed

taipy/core/data/_data_model.py
toan-quach: I'm afraid it might be difficult as I need it to make the tests passed. But let me have a look into it

taipy/core/data/data_node.py
toan-quach: For now I'd lean more on readability and simplicity as the topic is a little complex :D

taipy/core/data/data_node.py
toan-quach: this has been fixed with your proposal from yesterday @jrobinAV

taipy/core/notification/dn_scenario_cache.py
toan-quach: calling it a manager can be confusing IMHO, as of right now, I don't really see any features to extend from this class also :/

taipy/core/notification/dn_scenario_cache.py
toan-quach: I'm moving some of the computing responsibility from DataNode file to this class

taipy/core/notification/dn_scenario_cache.py
toan-quach: I'm thinking of proposing a new design for it

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,67 @@
...
+# specific language governing permissions and limitations under the License.
+
+from collections import defaultdict
+from typing import TYPE_CHECKING, Dict, Set
+
+if TYPE_CHECKING:
+    from ..data.data_node import DataNode, DataNodeId
+
+
+class SubmittableStatusCache:
```
jrobinAV: This should be a hidden class prefixed by an `_`.

taipy/core/data/data_node.py
```diff
@@ -29,12 +30,24 @@
...
 from ..exceptions.exceptions import DataNodeIsBeingEdited, NoData
 from ..job.job_id import JobId
+from ..notification._submittable_status_cache import SubmittableStatusCache
 from ..notification.event import Event, EventEntityType, EventOperation, _make_event
 from ._filter import _FilterDataNode
 from .data_node_id import DataNodeId, Edit
 from .operator import JoinOperator
 
 
+def _recompute_submittable_cache_wrapper(fct):
```
jrobinAV: I would call it `_update_ready_for_reading` and move it into the Datanode class, close to ready_for_reading property.

taipy/core/notification/_submittable_status_cache.py
jrobinAV: In the end, this class behaves like a Cache of the `ready_to_run` property.
We can call it `_ReadyToRunPropertyCache`. No?

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,67 @@
...
+    # A dictionary representing not ready_to_read datanode and its equivalent submittable entities
```
jrobinAV:
```suggestion
    # A dictionary of the data nodes not ready_to_read and their corresponding submittable entities.
```

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class SubmittableStatusCache:
+    IS_SUBMITTABLE_FCT_NAME = "is_submittable"
+
+    # A dictionary representing not ready_to_read datanode and its equivalent submittable entities
+    _datanode_id_submittables: Dict["DataNodeId", Set[str]] = defaultdict(lambda: set())
+
+    # A nested dictionary representing the submittable entities (Scenario, Sequence, Task) and
+    # its equivalent not ready_to_read input datanodes
```
jrobinAV:
```suggestion
    # A nested dictionary of the submittable entities (Scenario, Sequence, Task) and
    # the data nodes that make it not ready_to_run with the reason(s)
```

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,86 @@
...
+    _submittable_id_datanodes: Dict[str, Dict["DataNodeId", str]] = defaultdict(defaultdict)
```
jrobinAV: I believe in the end we need to store multiple reasons per data nodes so we can display all of them.

```suggestion
    _submittable_id_datanodes: Dict[str, Dict["DataNodeId", Set[str]]] = defaultdict(defaultdict)
```

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,86 @@
...
+
+if TYPE_CHECKING:
+    from ..data.data_node import DataNode, DataNodeId
+
+
+class SubmittableStatusCache:
+    IS_SUBMITTABLE_FCT_NAME = "is_submittable"
+
+    # A dictionary representing not ready_to_read datanode and its equivalent submittable entities
+    _datanode_id_submittables: Dict["DataNodeId", Set[str]] = defaultdict(lambda: set())
```
jrobinAV: I don't get the purpose of initializing it with an empty set lambda. If the data node is not in the dictionary, that means it is ready for reading, and we don't want to add it automatically.
No?

jrobinAV:
```suggestion
    _datanode_id_submittables: Dict["DataNodeId", Set["ScenarioId"]] = defaultdict(lambda: set())
```

taipy/core/notification/_submittable_status_cache.py
jrobinAV: Same comment about the lambda initialization. Why?

jrobinAV: I believe we should use an object Reason.

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # A nested dictionary representing the submittable entities (Scenario, Sequence, Task) and
+    # its equivalent not ready_to_read input datanodes
+    _submittable_id_datanodes: Dict[str, Dict["DataNodeId", str]] = defaultdict(defaultdict)
+
+    @classmethod
+    def __add(cls, entity_id: str, datanode_id: "DataNodeId", reason: str):
```
jrobinAV:
```suggestion
    def __add(cls, submittable_id: str, datanode_id: "DataNodeId", reason: str):
```

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,86 @@
...
+        cls._datanode_id_submittables[datanode_id].add(entity_id)
+        cls._submittable_id_datanodes[entity_id][datanode_id] = reason
+
+    @classmethod
+    def __remove(cls, datanode_id: "DataNodeId"):
```
jrobinAV: Actually we don't want to remove all the reasons at once, but one at a time, no?

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,86 @@
...
+            SubmittableStatusCache.__add(task_parent.id, dn.id, reason)
+
+    @classmethod
+    def _compute_if_dn_is_ready_for_reading(cls, dn: "DataNode"):
+        if dn._edit_in_progress:
+            cls.__add_parent_entities_to_submittable_cache(dn, f"DataNode {dn.id} is being edited")
+        elif not dn._last_edit_date:
+            cls.__add_parent_entities_to_submittable_cache(dn, f"DataNode {dn.id} is not written")
+        elif dn.is_ready_for_reading:
+            SubmittableStatusCache.__remove(dn.id)
```
jrobinAV: This method is the responsibility of the data node to me. It should be located in the DataNode class, next to the `is_read_for_reading` property. 
Plus, I would explicitly split the remove responsibility in two methods: 
- Removing the data node from the cache
- Notifying the data node is ready_for_reading.

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,86 @@
...
+                    )
+                )
+                cls._submittable_id_datanodes.pop(submittable_id, None)
+
+    @classmethod
+    def _check_submittable_is_ready_to_submit(cls, entity_id: str):
+        return len(SubmittableStatusCache._submittable_id_datanodes.get(entity_id, [])) == 0
+
+    @classmethod
+    def __add_parent_entities_to_submittable_cache(cls, dn: "DataNode", reason: str):
```
jrobinAV: Why not just call this method `add`? In the end, we add a data node to the cache.
The way it is implemented behind does not need to be exposed in the API name.

```suggestion
    def _add(cls, dn: "DataNode", reason: str):
```

taipy/core/_entity/submittable.py
```diff
@@ -86,7 +88,10 @@ def is_ready_to_run(self) -> bool:
         Returns:
             True if the given entity is ready to be run. False otherwise.
         """
-        return all(dn.is_ready_for_reading for dn in self.get_inputs())
+        if self._submittable_id not in SubmittableStatusCache._submittable_id_datanodes:
+            for dn in self.get_inputs():
+                SubmittableStatusCache._compute_if_dn_is_ready_for_reading(dn)
+        return SubmittableStatusCache._check_submittable_is_ready_to_submit(self._submittable_id)
```
jrobinAV: This does not work since calling `SubmittableStatusCache._compute_if_dn_is_ready_for_reading(dn)` triggers a notification if the scenario is submittable, while the value may not have changed recently.

In any case, why do you need to both access the cache and recompute the value?

I would simply rollback this part to use the previous code using `dn.is_ready_for_reading` directly, which should return the reasons as well.

tests/core/notification/test_events_published.py
```diff
@@ -166,6 +166,7 @@ def test_events_published_for_scenario_submission():
     register_id_0, register_queue_0 = Notifier.register()
     all_evts = RecordingConsumer(register_id_0, register_queue_0)
     all_evts.start()
+    # Write value to the unwritten data node trigger 3 is_submittable update events for the scenario, sequence and task
```
jrobinAV: 👍🏻 
Can you also update the fiollowing comment to make is correct regarding the three new events?

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,86 @@
...
+
+        submittable_ids: Set = cls._datanode_id_submittables.pop(datanode_id, set())
+        for submittable_id in submittable_ids:
+            cls._submittable_id_datanodes[submittable_id].pop(datanode_id, None)
+            if len(cls._submittable_id_datanodes[submittable_id]) == 0:
+                Notifier.publish(
+                    _make_event(
+                        tp_get(submittable_id),
+                        EventOperation.UPDATE,
+                        attribute_name=cls.IS_SUBMITTABLE_FCT_NAME,
```
jrobinAV: I propose the name : `IS_SUBMITTABLE_PROPERTY_NAME`

taipy/core/notification/_submittable_status_cache.py
```diff
@@ -0,0 +1,86 @@
...
+                        attribute_value=True,
+                    )
```
jrobinAV: The remove method and the notification should be split into multiple methods.

taipy/core/notification/_submittable_status_cache.py
toan-quach: sure why not :D

taipy/core/notification/_submittable_status_cache.py
toan-quach: would it be too complicated? and considering we're checking 1 by 1 conditions, not sure if it's useful enough?

taipy/core/data/data_node.py
toan-quach: since it's a wrapper function, would moving it into the DataNode class be more complicated w the extra parameters?

taipy/core/notification/_submittable_status_cache.py
toan-quach: to initiate an empty dictionary in case it has not been done yet. This is to avoid having to check if we have the key value pair in the dictionary or not? if not then we will initiate it with some value. But instead, everything will be done by default

taipy/core/notification/_submittable_status_cache.py
toan-quach: to initiate an empty dictionary in case it has not been done yet. This is to avoid having to check if we have the key value pair in the dictionary or not? if not then we will initiate it with some value. But instead, everything will be done by default

taipy/core/notification/_submittable_status_cache.py
toan-quach: because outside of adding to the cache, it's also responsible for getting the parents of the DN and getting only scenario, sequence and task entities and looping through all to add them into the cache. we also have an `__add` function that simply does the adding already

taipy/core/notification/_submittable_status_cache.py
jrobinAV: Actually, we may want to drop the Cache as well. `_ReadyToRunProperty`. What do you think?

toan-quach: I think having Cache makes the class more self-explanatory, for me, that's good :D

taipy/core/_entity/submittable.py
toan-quach: but the `is_ready_for_reading` in DN doesn't update the ReadyToRunCache and only checking the value of the data node? I don't think it will work no?

toan-quach: but the `is_ready_for_reading` in DN doesn't update the ReadyToRunCache and only checking the value of the data node? I don't think it will work no?

taipy/core/_entity/submittable.py
```diff
@@ -33,7 +34,8 @@ class Submittable:
         subscribers (List[Callable]): The list of callbacks to be called on `Job^`'s status change.
     """
 
-    def __init__(self, subscribers: Optional[List[_Subscriber]] = None):
+    def __init__(self, submittable_id: str, subscribers: Optional[List[_Subscriber]] = None):
```
jrobinAV: Please take the opportunity to add the return type to `__init__` method.

taipy/core/_entity/submittable.py
```diff
@@ -86,7 +88,10 @@ def is_ready_to_run(self) -> bool:
         Returns:
             True if the given entity is ready to be run. False otherwise.
         """
-        return all(dn.is_ready_for_reading for dn in self.get_inputs())
+        if self._submittable_id not in _ReadyToRunCache._submittable_id_datanodes:
+            for dn in self.get_inputs():
+                _compute_if_dn_is_ready_for_reading(dn)
+        return _ReadyToRunCache._check_submittable_is_ready_to_submit(self._submittable_id)
```
jrobinAV: I don't understand why we need to access the cache here. When a client requests the property value, we can compute it on the fly. There is no problem with that.

The cache is not made to make it faster to get the value. It is made to notify when the value changes.

I would rollback this code.

taipy/core/data/data_node.py
```diff
@@ -29,12 +30,33 @@
...
 from ..exceptions.exceptions import DataNodeIsBeingEdited, NoData
 from ..job.job_id import JobId
+from ..notification._ready_to_run_cache import _ReadyToRunCache
 from ..notification.event import Event, EventEntityType, EventOperation, _make_event
 from ._filter import _FilterDataNode
 from .data_node_id import DataNodeId, Edit
 from .operator import JoinOperator
 
 
+def _compute_if_dn_is_ready_for_reading(dn: "DataNode"):
```
jrobinAV: If this is not called by the submittable class anymore, it can be directly merged with the `_update_ready_for_reading`.

taipy/core/data/data_node.py
```diff
@@ -29,12 +30,33 @@
...
+
+def _update_ready_for_reading(fct):
+    # This decorator must be wrapped before self_setter decorator as self_setter will run the function twice.
+
+    @functools.wraps(fct)
+    def _recompute_is_ready_for_reading(dn: "DataNode", *args, **kwargs):
+        fct(dn, *args, **kwargs)
+        _compute_if_dn_is_ready_for_reading(dn)
+
+    return _recompute_is_ready_for_reading
```
jrobinAV: Moving this inside the data node class shouldn't make it more complex.

https://stackoverflow.com/questions/1263451/python-decorators-in-classes

taipy/core/notification/_ready_to_run_cache.py
```diff
@@ -0,0 +1,90 @@
...
+from . import EventOperation, Notifier, _make_event
+
+if TYPE_CHECKING:
+    from ..data.data_node import DataNode, DataNodeId
+    from ..scenario.scenario import ScenarioId
+    from ..sequence.sequence import SequenceId
+    from ..task.task import TaskId
+
+
+class _ReadyToRunCache:
```
jrobinAV: After more thinking, I am changing my mind. 
I believe we should not use the term cache in the class name. It is a technical consideration and does not bring any information on what is the responsibility of this class or the functionality it brings.

`_ReadyToRun`, `_ReadyToRunHandler`, `_ReadyToRunNotifier`, `_ReadyToRunController`, `_ReadyToRunProperty` are better names to me. 

I would personally simply go for `_ReadyToRun`.

taipy/core/data/data_node.py
jrobinAV: I don't think so.

taipy/core/notification/_submittable_status_cache.py
jrobinAV: I think it is useful. The end user wants to know all the reasons why its scenario is not submittable.

taipy/core/notification/_submittable_status_cache.py
jrobinAV: From the client's perspective of the ReadyToRun class, we just add a data node with a reason to the ReadyToRun class so it can handle it properly. The fact that it is retrieving the parents is an implementation detail to me. Not a big deal though.

taipy/core/_entity/submittable.py
jrobinAV: My understanding is that the `ReadyToRun` class is updated only when attributes  `edits_in_progress` or `last_edit_date` of a data node are updated. So we can track when the value changes and so when we need to send an event.

If the values of the attributes don't change, there is no need to update `ReadyToRun`, since there is no need to send events.

Am I missing something?

taipy/core/notification/_ready_to_run_cache.py
```diff
@@ -0,0 +1,90 @@
...
+            if len(cls._submittable_id_datanodes[submittable_id]) == 0:
+                cls._publish_submittable_property_event(submittable_id, True)
+                cls._submittable_id_datanodes.pop(submittable_id, None)
+
+    @classmethod
+    def _check_submittable_is_ready_to_submit(cls, submittable_id: Union["ScenarioId", "SequenceId", "TaskId"]):
+        return len(_ReadyToRunCache._submittable_id_datanodes.get(submittable_id, [])) == 0
+
+    @classmethod
+    def _add_parent_entities_to_submittable_cache(cls, dn: "DataNode", reason: str):
```
jrobinAV: I would call it `_add`.

taipy/core/notification/_ready_to_run_cache.py
```diff
@@ -0,0 +1,90 @@
...
+            _make_event(
+                tp_get(submittable_id),
+                EventOperation.UPDATE,
+                attribute_name=_ReadyToRunCache.IS_SUBMITTABLE_PROPERTY_NAME,
+                attribute_value=submittable_property,
+            )
+        )
+
+    @classmethod
+    def _add(cls, submittable_id: Union["ScenarioId", "SequenceId", "TaskId"], datanode_id: "DataNodeId", reason: str):
```
jrobinAV: I would rename it with a double underscore. 
Something like: `__add_for_submittable`, `__do_add`, `__update_cache`, `__add_and_notify`, etc...

taipy/core/notification/_ready_to_run_cache.py
```diff
@@ -0,0 +1,90 @@
...
+    def _remove(cls, datanode_id: "DataNodeId"):
+        submittable_ids: Set = cls._datanode_id_submittables.pop(datanode_id, set())
+        for submittable_id in submittable_ids:
+            cls._submittable_id_datanodes[submittable_id].pop(datanode_id, None)
+            if len(cls._submittable_id_datanodes[submittable_id]) == 0:
+                cls._publish_submittable_property_event(submittable_id, True)
+                cls._submittable_id_datanodes.pop(submittable_id, None)
+
+    @classmethod
+    def _check_submittable_is_ready_to_submit(cls, submittable_id: Union["ScenarioId", "SequenceId", "TaskId"]):
```
jrobinAV: If we don't access the `ReadyToRun` class, I believe we don't need this method anymore.

taipy/core/notification/_ready_to_run_cache.py
```diff
@@ -0,0 +1,90 @@
...
+        for sequence_parent in parent_entities.get(Sequence._MANAGER_NAME, []):
+            _ReadyToRunCache._add(sequence_parent.id, dn.id, reason)
+        for task_parent in parent_entities.get(Task._MANAGER_NAME, []):
+            _ReadyToRunCache._add(task_parent.id, dn.id, reason)
+
+    @classmethod
+    def get_reason_submittable_is_not_ready_to_submit(cls, submittable_id: Union["ScenarioId", "SequenceId", "TaskId"]):
+        if reasons := list(cls._submittable_id_datanodes.get(submittable_id, {}).values()):
+            return "; ".join(reasons) + "."
+        return ""
```
jrobinAV: This method is never used. 

In any case, we don't want to build the message to display. We want to return a Reason object or a list of Reasons. Defining how the message should be formatted and displayed should be the front-end responsibility on the graphical side.

taipy/core/notification/_ready_to_run_cache.py
jrobinAV, line 1: Sorry, but I am changing my mind again about where this class should be located.

I believe it is purely related to the management of the ready_to_run property of a submittable. It should be located close to the submittable class in the `_entity` package.

jrobinAV (commented):
In order to keep the comments readable, and to move on step by step, I did not add any comment on the possibility of storing multiple reasons. However, I still believe it is important.

tests/core/notification/test_events_published.py
```diff
@@ -166,6 +166,9 @@ def test_events_published_for_scenario_submission():
     register_id_0, register_queue_0 = Notifier.register()
     all_evts = RecordingConsumer(register_id_0, register_queue_0)
     all_evts.start()
+    # Before and after writing value to the unwritten data node trigger:
+    # 3 is_submittable update events for the scenario, sequence and task being not submittable
+    # 3 is_submittable update events for the scenario, sequence and task being submittable
```
jrobinAV: 👍🏻

taipy/core/_entity/submittable.py
toan-quach: 2 reasons why I decided to put it here:
- A way to trigger tracking the submittable status of the scenario
- To make sure, in case the Gui team check but the scenario is not submittable yet, it's still loaded into the cache so that when it's available for submission, it can notify the user/gui about that

toan-quach: we will rollback due to the unneeded initiation of the cache

taipy/core/notification/_ready_to_run_cache.py
toan-quach: we will go with `_ReadyToRunProperty`

taipy/core/data/data_node.py
toan-quach: the linter in GH Action didnt allow it so I moved it back outside the class

jrobinAV: Oh! 

OK thx for the information.

jrobinAV (dismissed):

tests/core/scenario/test_scenario_manager.py
```diff
@@ -589,7 +590,22 @@ def test_notification_subscribe(mocker):
...
+            notify_1,
+            notify_1,
+            notify_1,
+            notify_1,
+            notify_2,
+            notify_2,
+            notify_2,
+            notify_2,
+            notify_2,
+        ],
```
jrobinAV: I don't get this change. What is this side effect?

taipy/core/_entity/_ready_to_run_property.py
```diff
@@ -0,0 +1,93 @@
...
+                cls._submittable_id_datanodes[submittable_id].pop(datanode_id, None)
+                if len(cls._submittable_id_datanodes[submittable_id]) == 0:
+                    cls._publish_submittable_property_event(submittable_id, True)
+                    cls._submittable_id_datanodes.pop(submittable_id, None)
+
+        if to_remove_dn:
+            cls._datanode_id_submittables.pop(datanode_id)
+
+    @classmethod
+    def _add_parent_entities_to_submittable_cache(cls, dn: "DataNode", reason: str) -> None:
```
jrobinAV: The API would be simpler if the names were matching and simply _add and _remove: 

```
_ReadyToRunProperty._add(dn, "foo")
_ReadyToRunProperty._add(dn, "bar")
_ReadyToRunProperty._remove(dn, "foo")
_ReadyToRunProperty._add(dn2, "bar")
_ReadyToRunProperty._remove(dn2, "bar")
```

in the end, the client does not need to know how the dn reason is propagated to the submittables using parents. This is internal behavior.

tests/core/notification/test_events_published.py
```diff
@@ -156,8 +156,8 @@ def test_events_published_for_writing_dn():
 
 
 def test_events_published_for_scenario_submission():
-    input_config = Config.configure_data_node("the_input")
-    output_config = Config.configure_data_node("the_output")
+    input_config = Config.configure_pickle_data_node("the_input")
+    output_config = Config.configure_pickle_data_node("the_output")
```
jrobinAV: Any particular reason for this change?

tests/core/notification/test_events_published.py
```diff
@@ -167,6 +167,9 @@ def test_events_published_for_scenario_submission():
     register_id_0, register_queue_0 = Notifier.register()
     all_evts = RecordingConsumer(register_id_0, register_queue_0)
     all_evts.start()
+    # Before and after writing value to the unwritten data node trigger:
+    # 3 is_submittable update events for the scenario, sequence and task being not submittable
+    # 3 is_submittable update events for the scenario, sequence and task being submittable
```
jrobinAV: Actually, I don't get anymore why we need to change that. 

The input data node is already written when we start capturing the events. 
That means it is ready for reading. As a consequence, the task, sequence, and scenario are never unsubmittable, and they never turn submittable. We should not send any notification about that.

What am I missing here?

tests/core/notification/test_events_published.py
jrobinAV, line 1: I believe we are missing a test file for testing behaviors of the ReadToRunProperty class. It should also contain tests about the notifications.

tests/core/scenario/test_scenario_manager.py
```diff
@@ -934,21 +949,70 @@ def test_hard_delete_shared_entities():
...
+        f"DataNode {dn_2.id} is being edited"
+    }
     assert not _ScenarioManager._is_submittable(scenario)
     assert not _ScenarioManager._is_submittable(scenario.id)
 
-    scenario.dn.edit_in_progress = False
+    dn_2.edit_in_progress = False
+    assert dn_1.id not in _ReadyToRunProperty._datanode_id_submittables
+    assert dn_2.id not in _ReadyToRunProperty._datanode_id_submittables
+    assert scenario.id not in _ReadyToRunProperty._submittable_id_datanodes
```
jrobinAV: I believe the changes you made to this tests should not be here.
It should be tested in another test function in another test file dedicated to the ReadyToRunProperty.

tests/core/sequence/test_sequence_manager.py
```diff
@@ -189,25 +192,111 @@ def test_get_all_on_multiple_versions_environment():
...
+    dn_1 = PickleDataNode("dn_1", Scope.SCENARIO, parent_ids={task_id, scenario_id}, properties={"default_data": 10})
+    dn_2 = PickleDataNode("dn_2", Scope.SCENARIO, parent_ids={task_id, scenario_id}, properties={"default_data": 10})
+    task = Task("task", {}, print, [dn_1, dn_2], id=task_id, parent_ids={scenario_id})
+    scenario = Scenario("scenario", {task}, {}, set(), scenario_id=scenario_id)
+    _DataManager._set(dn_1)
+    _DataManager._set(dn_2)
+    _TaskManager._set(task)
     _ScenarioManager._set(scenario)
 
+    dn_1 = scenario.dn_1
```
jrobinAV: The new test cases should be moved to a dedicated file for ReadyToRunProperty class.

tests/core/task/test_task_manager.py
```diff
@@ -303,23 +306,107 @@ def test_hard_delete():
...
-    dn_config = Config.configure_in_memory_data_node("dn", 10)
-    task_config = Config.configure_task("task", print, [dn_config])
-    task = _TaskManager._bulk_get_or_create([task_config])[0]
-
-    assert len(_TaskManager._get_all()) == 1
-    assert _TaskManager._is_submittable(task)
-    assert _TaskManager._is_submittable(task.id)
-    assert not _TaskManager._is_submittable("Task_temp")
-
-    task.input["dn"].edit_in_progress = True
```
jrobinAV: Same remark on this function. The new test cases should be in a dedicated file.

tests/core/notification/test_events_published.py
toan-quach: none in particular, probably I forgot to revert when testing, will revert now

taipy/core/_entity/_ready_to_run_property.py
toan-quach: will rename:
-  `_add_parent_entities_to_submittable_cache` -> `_add`
- `__add_unsubmittable_reason` -> `__add`

tests/core/scenario/test_scenario_manager.py
toan-quach: from how I understand it, this is due to trying to load the task again and again

tests/core/notification/test_events_published.py
toan-quach: - when you write to the input data -> the dn will be converted to `edit_in_progress` -> triggering the events for entities that has the dn as the input that they're not ready to be submitted
- after the dn has been written -> `edit_in_progress` is triggered once again -> triggering the events for entities that has the dn as the input that they are ready to be submitted

tests/core/notification/test_events_published.py
toan-quach: have moved `test_is_submittable` from `test_scenario_manager`, `test_sequence_manager`, `test_task_manager` to `tests/core/_entity/test_ready_to_run_property.py`

tests/core/scenario/test_scenario_manager.py
jrobinAV: Ok, I got it. This _load_fct is used when we load the scenario from the repo to deserialize the subscription callback. 

That means before your change, a scenario submission was triggering one reload. Now it triggers 5 reloads. That's a lot.

jrobinAV (commented):
I am afraid of a performance decrease due to scenario reload when publishing events related to data node edition.

Let me think a few hours before approving the PR.

toan-quach:
@jrobinAV If I get the flow correctly, it must be due to the `get_parents` to try and get the parent entities to update their "submittable status"

taipy/core/_entity/_ready_to_run_property.py
```diff
@@ -0,0 +1,94 @@
...
+    @classmethod
+    def _add(cls, dn: "DataNode", reason: str) -> None:
+        from ..scenario.scenario import Scenario
+        from ..sequence.sequence import Sequence
+        from ..task.task import Task
+
+        parent_entities = dn.get_parents()
+
+        for scenario_parent in parent_entities.get(Scenario._MANAGER_NAME, []):
+            _ReadyToRunProperty.__add(scenario_parent, dn, reason)
```
jrobinAV:
```suggestion
            if dn in scenario_parent.get_inputs():
                _ReadyToRunProperty.__add(scenario_parent, dn, reason)
```

and same for other entity types

taipy/core/_entity/_ready_to_run_property.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @classmethod
+    def _remove(cls, datanode: "DataNode", reason: str) -> None:
+        # check the data node status to determine the reason to be removed
+        submittable_ids: Set = cls._datanode_id_submittables.get(datanode.id, set())
+
+        to_remove_dn = False
+        for submittable_id in submittable_ids:
+            # check remove the reason
+            if reason in cls._submittable_id_datanodes[submittable_id].get(datanode.id, set()):
```
jrobinAV:
```suggestion
            if reason in cls._submittable_id_datanodes.get(submittable_id,dict()).get(datanode.id, set()):
```

taipy/core/_entity/_ready_to_run_property.py
```diff
@@ -0,0 +1,94 @@
...
+    IS_SUBMITTABLE_PROPERTY_NAME: str = "is_submittable"
+
+    # A dictionary of the data nodes not ready_to_read and their corresponding submittable entities.
+    _datanode_id_submittables: Dict["DataNodeId", Set[Union["ScenarioId", "SequenceId", "TaskId"]]] = defaultdict(set)
+
+    # A nested dictionary of the submittable entities (Scenario, Sequence, Task) and
+    # the data nodes that make it not ready_to_run with the reason(s)
+    _submittable_id_datanodes: Dict[
+        Union["ScenarioId", "SequenceId", "TaskId"], Dict["DataNodeId", Set[str]]
+    ] = defaultdict(lambda: defaultdict(set))
```
jrobinAV: The lambda initialization is suspicious to me. I would remove it.

taipy/core/_entity/_ready_to_run_property.py
```diff
@@ -0,0 +1,94 @@
...
+        to_remove_dn = False
+        for submittable_id in submittable_ids:
+            # check remove the reason
+            if reason in cls._submittable_id_datanodes[submittable_id].get(datanode.id, set()):
+                cls._submittable_id_datanodes[submittable_id].get(datanode.id, set()).remove(reason)
+            if len(cls._submittable_id_datanodes[submittable_id][datanode.id]) == 0:
+                to_remove_dn = True
+                cls._submittable_id_datanodes[submittable_id].pop(datanode.id, None)
+                if len(cls._submittable_id_datanodes[submittable_id]) == 0:
+                    from ..taipy import get as tp_get
```
jrobinAV: This line should be moved outside the for loop.

tests/core/_entity/test_ready_to_run_property.py
jrobinAV, line 1: In this file, we should have functional tests that a mock publish is properly called according to some data node writing. 
Examples:
1. One test function should verify that writing an output does not trigger any event.
2. One test function should verify that writing a scenario intermediate data node triggers an event for a sequence with the data node as input.
3. etc.

tests/core/_entity/test_ready_to_run_property.py
```diff
@@ -0,0 +1,317 @@
...
+
+    dn_config_1 = Config.configure_in_memory_data_node("dn_1", 10)
+    dn_config_2 = Config.configure_in_memory_data_node("dn_2", 10)
+    task_config = Config.configure_task("task", print, [dn_config_1, dn_config_2])
+    scenario_config = Config.configure_scenario("sc", {task_config}, set(), Frequency.DAILY)
+    scenario = scenario_manager._create(scenario_config)
+    dn_1 = scenario.dn_1
+    dn_2 = scenario.dn_2
+
+    assert len(scenario_manager._get_all()) == 1
```
jrobinAV: this should be tested in the scenario manager test file

tests/core/_entity/test_ready_to_run_property.py
```diff
@@ -0,0 +1,317 @@
...
+    assert scenario.id not in _ReadyToRunProperty._submittable_id_datanodes
+    assert scenario_manager._is_submittable(scenario)
+    assert scenario_manager._is_submittable(scenario.id)
+    assert not scenario_manager._is_submittable("Scenario_temp")
```
jrobinAV: same for this

tests/core/notification/test_events_published.py
jrobinAV: Still the events should not be in all_evts since we start it after the events are published.

tests/core/notification/test_events_published.py
jrobinAV, line 1: We should have more tests impacted in this file, or add new tests. 
For instance, writing a data should trigger some events.

tests/core/scenario/test_scenario_manager.py
```diff
@@ -931,28 +933,6 @@ def test_hard_delete_shared_entities():
...
-
-    scenario.dn.edit_in_progress = True
-    assert not _ScenarioManager._is_submittable(scenario)
-    assert not _ScenarioManager._is_submittable(scenario.id)
-
-    scenario.dn.edit_in_progress = False
-    assert _ScenarioManager._is_submittable(scenario)
-    assert _ScenarioManager._is_submittable(scenario.id)
-
-
```
jrobinAV: Please rollback this removal.

tests/core/sequence/test_sequence_manager.py
```diff
@@ -188,30 +188,6 @@ def test_get_all_on_multiple_versions_environment():
...
-
-    scenario.dn.edit_in_progress = True
-    assert not _SequenceManager._is_submittable(sequence)
-    assert not _SequenceManager._is_submittable(sequence.id)
-
-    scenario.dn.edit_in_progress = False
-    assert _SequenceManager._is_submittable(sequence)
-    assert _SequenceManager._is_submittable(sequence.id)
-
-
```
jrobinAV: Please rollback this removal.

tests/core/task/test_task_manager.py
```diff
@@ -302,26 +302,6 @@ def test_hard_delete():
...
-    assert not _TaskManager._is_submittable(task)
-    assert not _TaskManager._is_submittable(task.id)
-
-    task.input["dn"].edit_in_progress = False
-    assert _TaskManager._is_submittable(task)
-    assert _TaskManager._is_submittable(task.id)
-
-
```
jrobinAV: Please rollback this removal.

jrobinAV (dismissed):

tests/core/notification/test_events_published.py
```diff
@@ -142,22 +142,23 @@ def test_events_published_for_writing_dn():
 
     # Write input manually trigger 4 data node update events
     # for last_edit_date, editor_id, editor_expiration_date and edit_in_progress
+    scenario.the_input.lock_edit()
```
jrobinAV: This line should not be necessary since input has never been written. So it is already not ready. This is a bug to me. When the data node is created and not written, it should be added to the ReadToRunProperty.

jrobinAV (commented):
I reworked the tests. I wanted to have better code coverage.
Plus I split them so we have one function for one use case. 

I believe I exhibited 3 bugs:
- When a data node is created without default data, the data node is not added to the ReadyToRunProperty.
- When a new scenario is created with one input. If the input already existed and was not ready for reading (in case of cycle or global scope), the scenario is not added to the ReadyToRunProperty.
- When a sequence is added to a scenario. If the input of the sequence is not ready for reading, the sequence is not added to the ReadyToRunProperty.

toan-quach:
@jrobinAV I believe that's due to 1 case, which is that we don't have an entry point for the ready to run check to be triggered. I believed we discussed about keeping it in the `is_ready_to_run` of class `Submittable` but decided that it's not necessary anymore as any changes to the `edit_in_progress` attribute `last_edit_date` would trigger the check and it's sufficient as it is.

jrobinAV:
> @jrobinAV I believe that's due to 1 case, which is that we don't have an entry point for the ready to run check to be triggered. I believed we discussed about keeping it in the `is_ready_to_run` of class `Submittable` but decided that it's not necessary anymore as any changes to the `edit_in_progress` attribute `last_edit_date` would trigger the check and it's sufficient as it is.

OK, let me change the tests. Thank you.

jrobinAV (dismissed):

jrobinAV (approved):

taipy/core/_entity/_ready_to_run_property.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # A nested dictionary of the submittable entities (Scenario, Sequence, Task) and
+    # the data nodes that make it not ready_to_run with the reason(s)
+    _submittable_id_datanodes: Dict[Union["ScenarioId", "SequenceId", "TaskId"], Dict["DataNodeId", Set[str]]] = {}
+
+    @classmethod
+    def _add(cls, dn: "DataNode", reason: str) -> None:
+        from ..scenario.scenario import Scenario
+        from ..sequence.sequence import Sequence
+        from ..task.task import Task
```
trgiangdo, line 37: If we can
```python
parent_entities.get("scenario", []
```
instead of
```python
parent_entities.get(Scenario._MANAGER_NAME, []
```
we probably don't need these import lines. However, the practice of using _MANAGER_NAME is ok to me as well

taipy/core/_entity/_ready_to_run_property.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for scenario_parent in parent_entities.get(Scenario._MANAGER_NAME, []):
+            if dn in scenario_parent.get_inputs():
+                _ReadyToRunProperty.__add(scenario_parent, dn, reason)
+        for sequence_parent in parent_entities.get(Sequence._MANAGER_NAME, []):
+            if dn in sequence_parent.get_inputs():
+                _ReadyToRunProperty.__add(sequence_parent, dn, reason)
+        for task_parent in parent_entities.get(Task._MANAGER_NAME, []):
+            if dn in task_parent.input.values():
+                _ReadyToRunProperty.__add(task_parent, dn, reason)
```
trgiangdo, line 49:
```suggestion
                cls.__add(scenario_parent, dn, reason)
        for sequence_parent in parent_entities.get(Sequence._MANAGER_NAME, []):
            if dn in sequence_parent.get_inputs():
                cls.__add(sequence_parent, dn, reason)
        for task_parent in parent_entities.get(Task._MANAGER_NAME, []):
            if dn in task_parent.input.values():
                cls.__add(task_parent, dn, reason)
```

taipy/core/_entity/_ready_to_run_property.py
```diff
@@ -0,0 +1,103 @@
...
+            if reason in cls._submittable_id_datanodes.get(submittable_id, {}).get(datanode.id, set()):
+                cls._submittable_id_datanodes[submittable_id][datanode.id].remove(reason)
+            if len(cls._submittable_id_datanodes.get(submittable_id, {}).get(datanode.id, set())) == 0:
+                to_remove_dn = True
+                cls._submittable_id_datanodes.get(submittable_id, {}).pop(datanode.id, None)
+                if (
+                    submittable_id in cls._submittable_id_datanodes
+                    and len(cls._submittable_id_datanodes[submittable_id]) == 0
+                ):
+                    submittable = tp_get(submittable_id)
```
trgiangdo, line 70: How about call the `_get` from the Manager instead of the `taipy.get()` to avoid the local import?

```suggestion
                    submittable = _SubmissionManagerFactory._build_manager()._get(submittable_id)
```

trgiangdo (approved):
Minor comments

taipy/core/_entity/_ready_to_run_property.py
jrobinAV: it is not the submission manager, and you don't know which manager you need. Sequence, Task or Scenario manager.

trgiangdo: Oh, you're right. Sorry for the confusion

taipy/core/_entity/_ready_to_run_property.py
toan-quach: I understand it's a little painful but I'll keep the import for manager_name instead to avoid hard-coding things to make it easier to make changes just in case :D

Issue #410, opened 2023-09-29
Possbility to import a scenario (after an export)


FlorianJacta:
**What would that feature address**
The possibility of importing a scenario/scenarios after exporting it with Taipy. The problem is that in the context of a project, different scenarios are present in different environments (Test, QA, Production). 

For example, the end user tests its application on the Test environment and creates meaningful scenarios. When the test is done, the application goes into QA and Production. The end user wants the scenarios he created in the other environments.

Another use case could be if the end-user encounters an issue with a scenario/scenarios he has created; he would then want to send it/them to its technical team (Data Scientist, Python Developers, ...) to debug them.

***Description of the ideal solution***
This export/import system should work with the different storage methods (Filesystem, SQL, Mongo). It could be another option of the CLI.
The goal would be to have the possibility to import a scenario. The difficulty comes from ensuring the consistency of the import regarding the entities related to the scenario and the data attached to data nodes.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Create related issue in taipy-doc for documentation and Release Notes
- [ ] Check if a new demo could be provided based on this, or if legacy demos could be benefit from it
- [ ] Ensure any change is well documented

trgiangdo:
For the current implementation of the `taipy.export()` API, user data is not exported alongside the entities.

We will need to check if the data is Taipy generated, we should also export the user data.

The new signature of the export method can be:
```
taipy.export(include_data=True)
```

When including the data, we will only export the file-based data node's data.

PR #1093, opened 2024-04-05
Feature - Expanding the `export_scenario()` api

trgiangdo:
Related to issue #410.

In this PR:
- `override` option is added to the `export_scenario()` method. Previously, we always delete the folder destination if existed. This is really not good. Now, by default, if the destination exporting folder has already exist, an error will be raised.
- `include_data` option is added to the `export_scenario()` method. This will only export data for file-based data nodes. If there is any non file-based data node, a warning will be raised and the data for that data node will not be exported.
- Split all tests for the `export_scenario()` to a dedicated file.

### DIFF
--- a/taipy/core/_manager/_manager.py
+++ b/taipy/core/_manager/_manager.py
@@ -155,9 +155,6 @@ class _Manager(Generic[EntityType]):
 
     @classmethod
     def _export(cls, id: str, folder_path: Union[str, pathlib.Path], **kwargs):
-        """
-        Export an entity.
-        """
         return cls._repository._export(id, folder_path)
 
     @classmethod
--- a/taipy/core/_manager/_manager.py
+++ b/taipy/core/_manager/_manager.py
@@ -154,7 +154,7 @@ class _Manager(Generic[EntityType]):
         _SubmissionManagerFactory._build_manager()._delete_many(_entity_ids.submission_ids)
 
     @classmethod
-    def _export(cls, id: str, folder_path: Union[str, pathlib.Path]):
+    def _export(cls, id: str, folder_path: Union[str, pathlib.Path], **kwargs):
         """
         Export an entity.
         """
--- a/taipy/core/_repository/_filesystem_repository.py
+++ b/taipy/core/_repository/_filesystem_repository.py
@@ -131,9 +131,6 @@ class _FileSystemRepository(_AbstractRepository[ModelType, Entity]):
             export_dir.mkdir(parents=True)
 
         export_path = export_dir / f"{entity_id}.json"
-        # Delete if exists.
-        if export_path.exists():
-            export_path.unlink()
 
         shutil.copy2(self.__get_path(entity_id), export_path)
 
--- a/taipy/core/_repository/_filesystem_repository.py
+++ b/taipy/core/_repository/_filesystem_repository.py
@@ -19,7 +19,7 @@ from taipy.config.config import Config
 
 from ..common._utils import _retry_repository_operation
 from ..common.typing import Converter, Entity, Json, ModelType
-from ..exceptions import FileCannotBeRead, InvalidExportPath, ModelNotFound
+from ..exceptions import FileCannotBeRead, ModelNotFound
 from ._abstract_repository import _AbstractRepository
 from ._decoder import _Decoder
 from ._encoder import _Encoder
@@ -122,9 +122,6 @@ class _FileSystemRepository(_AbstractRepository[ModelType, Entity]):
             folder: pathlib.Path = pathlib.Path(folder_path)
         else:
             folder = folder_path
-
-        if folder.resolve() == self._storage_folder.resolve():
-            raise InvalidExportPath("The export folder must not be the storage folder.")
 
         export_dir = folder / self._dir_name
         if not export_dir.exists():
--- a/taipy/core/data/_data_manager.py
+++ b/taipy/core/data/_data_manager.py
@@ -166,12 +166,7 @@ class _DataManager(_Manager[DataNode], _VersionMixin):
         return cls._repository._load_all(filters)
 
     @classmethod
-    def _export(
-        cls,
-        id: DataNodeId,
-        folder_path: Union[str, pathlib.Path],
-        **kwargs,
-    ):
+    def _export(cls, id: str, folder_path: Union[str, pathlib.Path], **kwargs):
         cls._repository._export(id, folder_path)
 
         if not kwargs.get("include_data"):
--- a/taipy/core/data/_data_manager.py
+++ b/taipy/core/data/_data_manager.py
@@ -168,20 +168,18 @@ class _DataManager(_Manager[DataNode], _VersionMixin):
     @classmethod
     def _export(
         cls,
-        data_node_id: DataNodeId,
+        id: DataNodeId,
         folder_path: Union[str, pathlib.Path],
-        include_data: Optional[bool] = False,
+        **kwargs,
     ):
-        cls._repository._export(data_node_id, folder_path)
+        cls._repository._export(id, folder_path)
 
-        if not include_data:
+        if not kwargs.get("include_data"):
             return
 
-        data_node = cls._get(data_node_id)
+        data_node = cls._get(id)
         if not isinstance(data_node, _FileDataNodeMixin):
-            cls._logger.warning(
-                f"Data node {data_node_id} is not a file-based data node and the data will not be exported."
-            )
+            cls._logger.warning(f"Data node {id} is not a file-based data node and the data will not be exported.")
             return
 
         if isinstance(folder_path, str):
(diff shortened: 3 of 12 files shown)


taipy/core/scenario/scenario.py
```diff
@@ -585,15 +585,22 @@ def submit(
     def export(
         self,
         folder_path: Union[str, pathlib.Path],
+        override: Optional[bool] = False,
```
jrobinAV: SHould not be an optional. Right?

taipy/core/data/_data_manager.py
```diff
@@ -161,3 +164,35 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
         for fil in filters:
             fil.update({"config_id": config_id})
         return cls._repository._load_all(filters)
+
+    @classmethod
+    def _export(
+        cls,
+        data_node_id: DataNodeId,
+        folder_path: Union[str, pathlib.Path],
+        include_data: Optional[bool] = False,
```
jrobinAV: SHould not be optional, right?

taipy/core/taipy.py
```diff
@@ -951,18 +954,33 @@ def export_scenario(
...
+    if os.path.exists(folder_path):
+        if override:
+            __logger.warn(f"Override the existing folder '{folder_path}'")
+            shutil.rmtree(folder_path, ignore_errors=True)
+        else:
+            __logger.error(
+                f"Folder '{folder_path}' already exists and can not be used to export scenario '{scenario_id}'."
+                " Please use the 'override' parameter to override it."
+            )
+            raise SystemExit()
```
jrobinAV: I believe we don't want to exit the system here. The raised exception must be catched.
Indeed, this is not a CLI, but a Python API. So it can be used by an end-user. We don't want the application to stop if a wrong path is provided at run time.

taipy/core/scenario/scenario.py
```diff
@@ -585,15 +585,15 @@ def submit(
...
         folder_path: Union[str, pathlib.Path],
-        override: Optional[bool] = False,
-        include_data: Optional[bool] = False,
+        override: bool = False,
+        include_data: bool = False,
     ):
         """Export all related entities of this scenario to a folder.
 
         Parameters:
             folder_path (Union[str, pathlib.Path]): The folder path to export the scenario to.
```
jrobinAV, line 594:
```suggestion
            folder_path (Union[str, pathlib.Path]): The folder path to export the scenario to.
                The path must not exist.
```

taipy/core/scenario/scenario.py
```diff
@@ -585,15 +585,15 @@ def submit(
...
-            override (Optional[bool]): If True, the existing folder will be overridden. Default is False.
-            include_data (Optional[bool]): If True, the file-based data nodes are exported as well.
+            override (bool): If True, the existing folder will be overridden. Default is False.
+            include_data (bool): If True, the file-based data nodes are exported as well.
                 This includes Pickle, CSV, Excel, Parquet, and JSON data nodes.
                 If the scenario has a data node that is not file-based, a warning will be logged, and the data node
                 will not be exported. The default value is False.
```
jrobinAV, line 600:
```suggestion
                will not be exported. The default value is False.
        
        Raises:
            ExportFolderAlreadyExists^: If the `folder_path` already exists.
```

taipy/core/taipy.py
```diff
@@ -954,8 +955,8 @@ def export_scenario(
     Parameters:
         scenario_id (ScenarioId): The ID of the scenario to export.
         folder_path (Union[str, pathlib.Path]): The folder path to export the scenario to.
-        override (Optional[bool]): If True, the existing folder will be overridden. Default is False.
-        include_data (Optional[bool]): If True, the file-based data nodes are exported as well.
+        override (bool): If True, the existing folder will be overridden. Default is False.
+        include_data (bool): If True, the file-based data nodes are exported as well.
             This includes Pickle, CSV, Excel, Parquet, and JSON data nodes.
             If the scenario has a data node that is not file-based, a warning will be logged, and the data node
             will not be exported. The default value is False.
```
jrobinAV, line 964:
```suggestion
            will not be exported. The default value is False.
        
        Raises:
            ExportFolderAlreadyExists^: If the `folder_path` already exists.
```

taipy/core/taipy.py
```diff
@@ -954,8 +955,8 @@ def export_scenario(
     Parameters:
         scenario_id (ScenarioId): The ID of the scenario to export.
         folder_path (Union[str, pathlib.Path]): The folder path to export the scenario to.
```
jrobinAV, line 958:
```suggestion
        folder_path (Union[str, pathlib.Path]): The folder path to export the scenario to.
            The path must not exist.
```

taipy/core/scenario/scenario.py
trgiangdo: What do you think about
```suggestion
            folder_path (Union[str, pathlib.Path]): The folder path to export the scenario to.
                If the path exists and the override parameter is False, an exception is raised.
```

The path can exist, as long as the `override` is True

taipy/core/scenario/scenario.py
trgiangdo: What do you think about
```suggestion
                will not be exported. The default value is False.

        Raises:
            ExportFolderAlreadyExist^: If the `folder_path` already exists and the override parameter is False.
```

jrobinAV: Much better. Thx!

taipy/core/scenario/scenario.py
jrobinAV: Much better.

jrobinAV (dismissed):

taipy/core/_manager/_manager.py
```diff
@@ -154,7 +154,7 @@ def _delete_entities_of_multiple_types(cls, _entity_ids: _EntityIds):
         _SubmissionManagerFactory._build_manager()._delete_many(_entity_ids.submission_ids)
 
     @classmethod
-    def _export(cls, id: str, folder_path: Union[str, pathlib.Path]):
+    def _export(cls, id: str, folder_path: Union[str, pathlib.Path], **kwargs):
         """
         Export an entity.
         """
```
jrobinAV: It should be up to date if we want to keep this docstring (Which is not mandatory since the function is not exposed as a public API). Otherwise, we can remove it.

taipy/core/scenario/_scenario_manager.py
```diff
@@ -416,7 +416,7 @@ def _get_children_entity_ids(cls, scenario: Scenario) -> _EntityIds:
         submissions = _SubmissionManagerFactory._build_manager()._get_all()
         submitted_entity_ids = list(entity_ids.scenario_ids.union(entity_ids.sequence_ids, entity_ids.task_ids))
         for submission in submissions:
-            if submission.entity_id in submitted_entity_ids:
+            if submission.entity_id in submitted_entity_ids or submission.entity_id == scenario.id:
```
jrobinAV, line 419: Good catch !

jrobinAV (dismissed):

toan-quach (dismissed):
lgtm, I think my comments from https://github.com/Avaiga/taipy/pull/1194#pullrequestreview-2005359960 would also apply 🤣

taipy/core/_repository/_filesystem_repository.py
```diff
@@ -123,17 +123,11 @@ def _export(self, entity_id: str, folder_path: Union[str, pathlib.Path]):
         else:
             folder = folder_path
 
-        if folder.resolve() == self._storage_folder.resolve():
-            raise InvalidExportPath("The export folder must not be the storage folder.")
```
trgiangdo, line 127: @toan-quach asked:
> why are we removing this code?

This is moved to the `taipy.import()` since we only need to check it once.

taipy/core/_repository/_filesystem_repository.py
```diff
@@ -123,17 +123,11 @@ def _export(self, entity_id: str, folder_path: Union[str, pathlib.Path]):
...
-
         export_dir = folder / self._dir_name
         if not export_dir.exists():
             export_dir.mkdir(parents=True)
 
         export_path = export_dir / f"{entity_id}.json"
-        # Delete if exists.
-        if export_path.exists():
-            export_path.unlink()
```
trgiangdo, line 136: @toan-quach asked:
> why are we removing this code?

It's is already handled in the `taipy.import()` function, so this is duplicated.

taipy/core/scenario/_scenario_manager.py
trgiangdo: Previously, when we find children submissions of a scenario, we only check if the `submission.entity_id` is one of the children tasks or sequences.

However, the `submission.entity_id` can also be the scenario itself.
@toan-quach Hope this answer your question.

taipy/core/_manager/_manager.py
trgiangdo: I removed it since we already explain these in the `taipy.export()` method.

jrobinAV (approved):
Looks good to me.

PR #1109, opened 2024-04-09
Fix - Failing update dependencies action

### LINKED ISSUE #1192: Run overall-test workflow when creating Update Dependencies PRs
**Description**

Currently, the step that runs overall-test workflow is failing when creating Update Dependencies PRs. We need to make sure that the action:
- Create PRs with proper changes
- Run the tests to make sure that new dependencies work with the current implementation of Taipy

trgiangdo:
This PR fixed problems with the dependency management workflow.
Runs on each Sunday, creating PR with dependencies updates on the repository:
image

A link to the test workflow is put in the description:
image


.github/workflows/dependencies-management.yml
```diff
@@ -76,12 +85,24 @@ jobs:
...
+          github-token: ${{ secrets.TRIGGER_GITHUB_PR }}
           script: |
-            github.rest.actions.createWorkflowDispatch({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              workflow_id: 'overall-tests.yml',
-              ref: 'dependencies/update-python${{matrix.python-version}}',
-            })
+            const runTests = require('.github/scripts/run-workflow.js')
+            const linkTests = require('.github/scripts/link-workflow-to-pr.js')
```
jrobinAV, line 93: Two questions: 
1. Why in js and not in Python like other scripts?
2. Why not move this into a specific file?

.github/workflows/overall-tests.yml
```diff
@@ -12,6 +12,7 @@ jobs:
     uses: ./.github/workflows/partial-tests.yml
 
   coverage:
+    timeout-minutes: 40
```
jrobinAV, line 15: This step runs in // of other tests. there is no point in limiting this to a smaller timeout than the other steps.

I would rollback this.

.github/scripts/link-workflow-to-pr.js
jrobinAV, line 1: Why JS?

.github/scripts/run-workflow.js
jrobinAV, line 1: Why JS?

.github/workflows/dependencies-management.yml
florian-vuillemot: To be able to use the native "Github" (technically, it is [octokit](https://github.com/octokit/octokit.js/)) object, which simplifies GitHub interactions.

I created two files under the `.github` folder for clarity - reading/updating code in the middle of a YAML is not friendly - and save workflow size by 50 lines. Unfortunately, there is limitation and I need some code in the middle.

.github/scripts/run-workflow.js
florian-vuillemot: Explained below.

.github/scripts/link-workflow-to-pr.js
florian-vuillemot: Explained below.

.github/workflows/overall-tests.yml
florian-vuillemot: Good point! So I would prefer to put the same timeout as the other steps. I don't like the idea of not having a timeout; we will consume unnecessary energy in case of an infinite loop/proceess stuck. What do you think?

jrobinAV: The other steps have a timeout. I believe it is 50 minutes. Can you check and align?

.github/workflows/dependencies-management.yml
jrobinAV: oki

jrobinAV (approved):

.github/workflows/overall-tests.yml
trgiangdo: Other steps are not align in term of timeout. Some 40, some 50

jrobinAV: Ok, so either we align all the test actions to max 50 minutes, or we accept to have a different time out.

There is nothing blocking to me here. I am ok with both solutions. I let you decide.

trgiangdo: We can align the timeout in a different PR

jrobinAV:
LEt's merge it then.

PR #1161, opened 2024-04-12
fix: add numpy to pipfile

joaoandre-avaiga:


jrobinAV (commented):
Looks good to me.
 Is it still a draftt?

Pipfile
```diff
@@ -22,6 +22,7 @@ marshmallow = "==3.20.1"
 networkx = "==2.6"
 openpyxl = "==3.1.2"
 pandas = "==1.3.5"
+numpy = "<2.0.0"
```
trgiangdo: Should we put this under "dev-packages"?
Just to indicate that this condition is only used when developing taipy, not an actual taipy requirement.

trgiangdo (dismissed):
Minor suggestion. Other than that, the change looks good to me

Pipfile
joaoandre-avaiga: I like it. Changing it now

trgiangdo (approved):

toan-quach (dismissed):

trgiangdo (changes requested):
The coverage test is failing. Probably because it's not using the same Pipfile?

jrobinAV (dismissed):

toan-quach:
I have the same thinking with @trgiangdo but I just checked and not sure what the problem is 🤣

jrobinAV (approved):

trgiangdo (approved):

PR #1193, opened 2024-04-16
Removing useless type annotation

jrobinAV:
- Add none type return to init methods

### DIFF
--- a/taipy/core/_entity/_entity_ids.py
+++ b/taipy/core/_entity/_entity_ids.py
@@ -11,16 +11,20 @@
 
 from __future__ import annotations
 
+from typing import Set
+
 
 class _EntityIds:
     def __init__(self) -> None:
-        self.data_node_ids = set()
-        self.task_ids = set()
-        self.scenario_ids = set()
-        self.sequence_ids = set()
-        self.job_ids = set()
-        self.cycle_ids = set()
-        self.submission_ids = set()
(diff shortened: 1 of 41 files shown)


taipy/core/_entity/_entity_ids.py
```diff
@@ -11,16 +11,20 @@
 
 from __future__ import annotations
 
+from typing import Set
```
trgiangdo:
```suggestion
from typing import TYPE_CHECKING, Set

if TYPE_CHECKING:
    from taipy import CycleId, DataNodeId, JobId, ScenarioId, SequenceId, SubmissionId, TaskId
```

I would do this to avoid the local import

taipy/core/_entity/_entity_ids.py
```diff
@@ -11,16 +11,20 @@
...
-        self.data_node_ids = set()
-        self.task_ids = set()
-        self.scenario_ids = set()
-        self.sequence_ids = set()
-        self.job_ids = set()
-        self.cycle_ids = set()
-        self.submission_ids = set()
+    def __init__(self) -> None:
+        from taipy import CycleId, DataNodeId, JobId, ScenarioId, SequenceId, SubmissionId, TaskId
+
```
trgiangdo:
```suggestion
```

trgiangdo (approved):
There is no mypy warning anymore. Great work :D

PR #1194, opened 2024-04-16
Feature/#410 - Import scenario from an exported one

### LINKED ISSUE #410: Possbility to import a scenario (after an export)

trgiangdo:
Resolves #410


taipy/core/taipy.py
```diff
@@ -973,6 +998,104 @@ def export_scenario(
         _ScenarioManagerFactory._build_manager()._export(scenario_id, folder_path)
     for job_id in entity_ids.job_ids:
         _JobManagerFactory._build_manager()._export(job_id, folder_path)
+    for submission_id in entity_ids.submission_ids:
+        _SubmissionManagerFactory._build_manager()._export(submission_id, folder_path)
+    _VersionManagerFactory._build_manager()._export(scenario.version, folder_path)
+
+
+def import_scenario(folder_path: Union[str, pathlib.Path], override: bool = False):
+    """Import a folder contains an exported scenario into the current Taipy application.
```
jrobinAV:
```suggestion
    """Import a folder containing an exported scenario into the current Taipy application.
```

jrobinAV: can you add an explanation for the version number change?

jrobinAV (commented):
sounds good

taipy/core/_repository/_filesystem_repository.py
```diff
@@ -123,20 +123,19 @@ def _export(self, entity_id: str, folder_path: Union[str, pathlib.Path]):
         else:
             folder = folder_path
 
-        if folder.resolve() == self._storage_folder.resolve():
-            raise InvalidExportPath("The export folder must not be the storage folder.")
-
```
toan-quach: why are we removing this code?

taipy/core/_repository/_filesystem_repository.py
```diff
@@ -123,20 +123,19 @@ def _export(self, entity_id: str, folder_path: Union[str, pathlib.Path]):
...
         export_dir = folder / self._dir_name
         if not export_dir.exists():
             export_dir.mkdir(parents=True)
 
         export_path = export_dir / f"{entity_id}.json"
-        # Delete if exists.
-        if export_path.exists():
-            export_path.unlink()
```
toan-quach: why are we removing this code?

toan-quach: why are we removing this code?

taipy/core/data/data_node.py
```diff
@@ -121,6 +121,13 @@ def __init__(
 
         self._properties = _Properties(self, **kwargs)
 
+    @staticmethod
+    def _new_id(config_id: str) -> DataNodeId:
+        """Generate a unique datanode identifier."""
+        return DataNodeId(
+            DataNode.__ID_SEPARATOR.join([DataNode._ID_PREFIX, _validate_id(config_id), str(uuid.uuid4())])
+        )
```
toan-quach: 👍

taipy/core/scenario/_scenario_manager.py
```diff
@@ -416,7 +416,7 @@ def _get_children_entity_ids(cls, scenario: Scenario) -> _EntityIds:
         submissions = _SubmissionManagerFactory._build_manager()._get_all()
         submitted_entity_ids = list(entity_ids.scenario_ids.union(entity_ids.sequence_ids, entity_ids.task_ids))
         for submission in submissions:
-            if submission.entity_id in submitted_entity_ids:
+            if submission.entity_id in submitted_entity_ids or submission.entity_id == scenario.id:
```
toan-quach: not sure if I understand this part :/

taipy/core/taipy.py
```diff
@@ -973,6 +998,104 @@ def export_scenario(
...
+
+        # Skip the version folder as it is already checked
+        if entity_folder.name == "version":
+            continue
+
+        entity_type = entity_folder.name
+        manager = entity_managers[entity_type]()
+        for entity_file in entity_folder.iterdir():
+            imported_entity = manager._import(
+                entity_file,
```
toan-quach: should we put in a try-catch to raise an error saying if a file cannot be imported?

toan-quach (dismissed):
lgtm, minor comments 😃

taipy/core/exceptions/exceptions.py
```diff
@@ -373,6 +373,47 @@ class FileCannotBeRead(Exception):
...
+    """Raised when the import folder doesn't contain any scenario"""
+
+    def __init__(self, folder_path):
+        self.message = f"The import folder {folder_path} doesn't contain any scenario."
+
+
+class ImportScenarioDoesntHaveAVersion(Exception):
+    """Raised when the import folder doesn't contain any scenario"""
+
+    def __init__(self, folder_path):
```
jrobinAV: Please take the opportunity of this PR to add the return type for `__init__` methods. Thx

taipy/core/taipy.py
```diff
@@ -973,6 +998,106 @@ def export_scenario(
...
+        "data_nodes": _DataManagerFactory._build_manager,
+        "data_node": _DataManagerFactory._build_manager,
+        "tasks": _TaskManagerFactory._build_manager,
+        "task": _TaskManagerFactory._build_manager,
+        "scenarios": _ScenarioManagerFactory._build_manager,
+        "scenario": _ScenarioManagerFactory._build_manager,
+        "jobs": _JobManagerFactory._build_manager,
+        "job": _JobManagerFactory._build_manager,
+        "submission": _SubmissionManagerFactory._build_manager,
+        "version": _VersionManagerFactory._build_manager,
```
jrobinAV: Really strange... Is it some kind of temporary code until we make this consistent?

taipy/core/taipy.py
```diff
@@ -973,6 +998,106 @@ def export_scenario(
...
+                version=_VersionManagerFactory._build_manager()._get_latest_version(),
+                data_folder=folder / valid_data_folder,
+            )
+
+            imported_entities.append(imported_entity)
+            if entity_type in ["scenario", "scenarios"]:
+                imported_scenario = imported_entity
+
+    __logger.info(f"Scenario {imported_scenario.id} has been successfully imported.")  # type: ignore[union-attr]
+    return imported_scenario
```
jrobinAV: This function is complex. It seems too complex to remain in the taipy.py file, which is the top-level API. 
Would delegating most of this logic to the manager classes make sense?

tests/core/data/test_csv_data_node.py
```diff
@@ -102,7 +103,8 @@ def test_new_csv_data_node_with_existing_file_is_ready_for_reading(self):
         ],
     )
     def test_create_with_default_data(self, properties, exists):
-        dn = CSVDataNode("foo", Scope.SCENARIO, DataNodeId("dn_id"), properties=properties)
+        dn = CSVDataNode("foo", Scope.SCENARIO, DataNodeId(f"dn_id_{uuid.uuid4()}"), properties=properties)
```
jrobinAV: What s the rational for this unique id?

jrobinAV (commented):
Looks good to me.

taipy/core/taipy.py
```diff
@@ -1015,12 +1015,11 @@ def import_scenario(folder_path: Union[str, pathlib.Path], override: bool = Fals
         override (bool): If True, override the entities if existed. Default value is False.
 
     Return:
-        The imported scenario.
+        The imported scenario if the import is successful.
```
jrobinAV: I believe it is not necessary to add that. It is implicit for me.

taipy/core/taipy.py
trgiangdo: we now have rollback mechanism where if there is any exception, all imported entities will be deleted.

taipy/core/_repository/_filesystem_repository.py
trgiangdo: Answered in #1093.

It's is already handled in the `taipy.export()` function, so this is duplicated.

taipy/core/_repository/_filesystem_repository.py
trgiangdo: Answered in #1093.

It's is already handled in the `taipy.export()` function, so this is duplicated.

taipy/core/scenario/_scenario_manager.py
trgiangdo: Previously, when we find children submissions of a scenario, we only check if the `submission.entity_id` is one of the children tasks or sequences.

However, the `submission.entity_id` can also be the scenario itself.

taipy/core/exceptions/exceptions.py
trgiangdo: We are not having the `-> None` on the Exception classes. Should we do it all?

taipy/core/taipy.py
trgiangdo: It's not temporary. Since we are relying on the entity type to see which manager is needed, this dictionary is the best I can think of that still make the code readable.

The duplicated entries are because some have "s" and some don't.

taipy/core/taipy.py
trgiangdo: I agree. I will try to refactor this next

tests/core/data/test_csv_data_node.py
trgiangdo: To avoid duplicated id. I added this during testing and it's probably not needed anymore. But I think it's a good addition to the test

taipy/core/taipy.py
trgiangdo: We can discuss this

taipy/core/taipy.py
```diff
@@ -973,6 +998,121 @@ def export_scenario(
...
+    except Exception as err:
+        __logger.error(f"An error occurred during the import: {err}. Rollback the import.")
+
+        # Rollback the import
+        for entity_type, entity_ids in list(imported_entities.items())[::-1]:
+            manager = entity_managers[entity_type]()
+            for entity_id in entity_ids:
+                if manager._exists(entity_id):
+                    manager._delete(entity_id)
+        return None
```
trgiangdo: Here, when there is any Exception, we will return None. Do we want to raise the Exception again here, after the rollback?

taipy/core/exceptions/exceptions.py
jrobinAV: Yes, I think so. 
I would suggest doing this kind of changes iteratively, small by small, instead of doing everything in the same big PR. That's why I was talking about taking the opportunity.
However, since your PR is already big enough if you believe it will make it too difficult to review, we can decide not to do it in this PR.

Nothing is blocking for me anyway. It's up to you.

tests/core/data/test_csv_data_node.py
jrobinAV: Make sense. Thx.

taipy/core/job/_job_manager.py
```diff
@@ -58,7 +58,9 @@ def _create(
         return job
 
     @classmethod
-    def _delete(cls, job: Job, force=False):
+    def _delete(cls, job: Union[Job, JobId], force=False):
+        if isinstance(job, str):
```
jrobinAV, line 62: Why str and not JobId ?
```suggestion
        if isinstance(job, JobId):
```

taipy/core/scenario/_scenario_manager.py
```diff
@@ -450,3 +457,85 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
...
+            except Exception as err:
+                cls._logger.error(f"An error occurred during the import: {err}. Rollback the import.")
+
+                # Rollback the import
+                for entity_type, entity_ids in list(imported_entities.items())[::-1]:
+                    manager = entity_managers[entity_type]
+                    for entity_id in entity_ids:
+                        if manager._exists(entity_id):
+                            manager._delete(entity_id)
+                return None
```
trgiangdo: Here, when there is any Exception, we will return None. Do we want to raise the Exception again here, after the rollback?
```suggestion
                raise err
```

jrobinAV: Yes, I would raise the error again.

tests/core/test_taipy/test_export.py
```diff
@@ -84,86 +101,97 @@ def test_export_scenario_with_cycle():
...
-    assert os.path.exists("./tmp/exp_scenario/tasks")
-    assert os.path.exists("./tmp/exp_scenario/scenarios")
-    assert os.path.exists("./tmp/exp_scenario/jobs")
-    assert os.path.exists("./tmp/exp_scenario/submission")
-    # The cycles folder should be removed when overriding
-    assert not os.path.exists("./tmp/exp_scenario/cycles")
-
-
-def test_export_scenario_filesystem_with_data():
+    tp.export_scenario(scenario_2.id, "tmp.zip", override=True)
```
jrobinAV, line 165: Should we add tmp.zip to the .gitignore?
Should we remove it in the conftest?

jrobinAV (dismissed):
Minor questions

taipy/core/job/_job_manager.py
trgiangdo: Probably a copy-paste systematic error.

We use the `if isinstance(job, str)` quite a lot, and similarly for other entity types, like `if isinstance(scenario, str)` in the _ScenarioManager

tests/core/test_taipy/test_export.py
trgiangdo: I would not.

All tests exported to the `tmp.zip` file, so we might get the error where the later test using the exported file from the previous test.

To override the `tmp.zip` from the previous test, all tests should use the `tp.export("tmp.zip", override=True)`, which is not desirable

jrobinAV (approved):

PR #1224, opened 2024-04-23
<DataNode>.is_up_to_date raises error when never written before

### LINKED ISSUE #1198: BUG- <DataNode>.is_up_to_date raises error when never written before
**Description**

The `<DataNode>.is_up_to_date` method breaks when the Data Node has never been written.

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "c:\Users\jacta\.conda\envs\bar-cutting\lib\site-packages\taipy\core\data\data_node.py", line 528, in is_up_to_date
    and ancestor_node.last_edit_date > self.last_edit_date
TypeError: '>' not supported between instances of 'datetime.datetime' and 'NoneType'
```

**How to reproduce**

```python
from taipy.config import Config
import taipy as tp

Config.configure_job_executions(mode="standalone")


def double(nb):
    return nb * 2


# Configuration of Data Nodes
input_cfg = Config.configure_data_node(id="somedata", default_data=21)
output_cfg = Config.configure_data_node(id="result")


# Configuration of tasks
first_task_cfg = Config.configure_task(id="double",
                                       function=double,
                                       input=input_cfg,
                                       output=output_cfg)


# Configuration of scenario
scenario_cfg = Config.configure_scenario(id="my_scenario",
                                         task_configs=[first_task_cfg],
                                         name="my_scenario")


if __name__=="__main__":
    tp.Core().run()
    scenario_1 = tp.create_scenario(scenario_cfg)

    print(scenario_1.result.is_up_to_date)
```

**Expected behavior**
We should not get an error; the result should be False if the Data Node is invalid.

Possible solution:

Change the *is_up_to_date* function in *taipy/core/data/data_node.py*, line 510:

```python
@property
def is_up_to_date(self) -> bool:
    """Indicate if this data node is up-to-date.

    Returns:
        False if a preceding data node has been updated before the selected data node
        or the selected data is invalid.<br/>
        True otherwise.
    """

    if self.is_valid:
        from ..scenario.scenario import Scenario
        from ..taipy import get_parents

        parent_scenarios: Set[Scenario] = get_parents(self)["scenario"]  # type: ignore
        for parent_scenario in parent_scenarios:
            for ancestor_node in nx.ancestors(parent_scenario._build_dag(), self):
                if (
                    isinstance(ancestor_node, DataNode)
                    and ancestor_node.last_edit_date
                    and ancestor_node.last_edit_date > self.last_edit_date
                ):
                    return False
    return False

```

**Runtime environment**
 - Taipy: 3.1.1

**Acceptance Criteria**
- [ ] #1200

yaten2302:
# Fixes #1198


trgiangdo:
Hello @yaten2302, thank you for your contribution.

Can you add a unittest that fails before your fix, and passes after your fix?

yaten2302:
Hey @trgiangdo, actually, I'm not really familiar on how to add tests. Could you please guide that how can I create and add the same?

taipy/core/data/data_node.py
```diff
@@ -505,25 +505,27 @@ def is_valid(self) -> bool:
...
-        Returns:
-            False if a preceding data node has been updated before the selected data node
-            or the selected data is invalid.<br/>
-            True otherwise.
-        """
+    Returns:
+        False if a preceding data node has been updated before the selected data node
+        or the selected data is invalid.<br/>
+        True otherwise.
+    """
```
jrobinAV: We have a wrong indentation here.
```suggestion
        Returns:
            False if a preceding data node has been updated before the selected data node
            or the selected data is invalid.<br/>
            True otherwise.
        """
```

trgiangdo:
@yaten2302 You can take a look at https://github.com/Avaiga/taipy/blob/c3aaf92e767c0844c698701a9c18d15d19bf60ba/tests/core/data/test_data_node.py.
Your test maybe similar to the tests in this file.

Basically, the test should:
- Create a DataNode entity (InMemoryDataNode or PickleDataNode for example)
- Try to assert if the `dn.is_up_to_date` attribute is False. Before your fix, this will raise an error.

taipy/core/data/data_node.py
yaten2302: Made the changes👍

yaten2302:
> @yaten2302 You can take a look at https://github.com/Avaiga/taipy/blob/c3aaf92e767c0844c698701a9c18d15d19bf60ba/tests/core/data/test_data_node.py. Your test maybe similar to the tests in this file.
> 
> Basically, the test should:
> 
> * Create a DataNode entity (InMemoryDataNode or PickleDataNode for example)
> * Try to assert if the `dn.is_up_to_date` attribute is False. Before your fix, this will raise an error.

Do I've to do like this? In this file: https://github.com/Avaiga/taipy/blob/c3aaf92e767c0844c698701a9c18d15d19bf60ba/tests/core/data/test_data_node.py<br>
```python
class TestData:
   def testDataNode:
       dn = InMemoryDataNode("dn", Scope.SCENARIO)
       assert dn.is_up_to_date == False
```

trgiangdo:
Yes, and you can also try to write to the data node, then `assert dn.is_up_to_date == True` since it should be True then

yaten2302:
Thanks, @trgiangdo. I'll make the changes by today. Also, could you please tell that if we want to test the changes, then do we've to run the command - `pipenv run pytest`?

yaten2302:
Hey @trgiangdo, I've put this code in the file [tests/core/data/test_data_node.py](https://github.com/Avaiga/taipy/blob/develop/tests/core/data/test_data_node.py). After that when I run the `pipenv run pytest` command, I'm getting this. Ig this means failure in a test right? Could you please guide that how can this be fixed? <br>
image

trgiangdo:
Seems like the modification you made causes the `tests/core/data/test_data_node.py::TestDataNode::test_is_up_to_date_across_scenarios()` and `tests/core/data/test_data_node.py::TestDataNode::test_is_up_to_date()` to failed.

The reason should be shown the terminal when you run the test. You can run the specific test by running the `pytest <<path_to_test_file>>` to run the 1 failing file only.

yaten2302:
@trgiangdo, when I'm running the command `pytest test_data_node.py`, it's showing this error.<br>
image
I've checked the dir as well, the `test_data_node.py` file is located in the `taipy/tests/core/data` dir only. Could you please guide that how can this be fixed?

trgiangdo:
You are not in the virtualenv. Please run pytest with the "pipenv run ..." or activate the virtualenv

yaten2302:
It's again showing the same error @trgiangdo <br>
image
This time it created a Pipfile as well in the `tests/core/data` dir.

trgiangdo:
It's because you are running the pipenv command in a different directory. You need to cd back to the taipy folder where you created the previous virtualenv and run the test

yaten2302:
@trgiangdo, but when I'm running this command in the `taipy` folder, then it's showing the error that the file `test_data_node.py` not found.<br>
image

trgiangdo:
It's because the path you provided after the `pipenv run pytest <<path>>` is wrong. You need to provide the full path.

yaten2302:
Yes, now it's working 👍 @trgiangdo. It's showing error in 2 functions - `test_is_up_to_date` (which I created) and `test_is_up_to_date_across_scenarios` (which was already present). <br>
image
<br>
image
Could you please guide that how can these be fixed?

trgiangdo:
You modified the `.is_up_to_date` attribute so that it will be False when the data node is not written. I don't see the data node is written in those tests, am I right?

yaten2302:
Yes, @trgiangdo , what my test does is that it'll return False if the data node is not written.

trgiangdo:
Then why are you testing if it is up to date with the `assert dn.is_up_to_date`?

yaten2302:
> > @yaten2302 You can take a look at https://github.com/Avaiga/taipy/blob/c3aaf92e767c0844c698701a9c18d15d19bf60ba/tests/core/data/test_data_node.py. Your test maybe similar to the tests in this file.
> > 
> > Basically, the test should:
> > 
> > * Create a DataNode entity (InMemoryDataNode or PickleDataNode for example)
> > * Try to assert if the `dn.is_up_to_date` attribute is False. Before your fix, this will raise an error.
> 
> Do I've to do like this? In this file: https://github.com/Avaiga/taipy/blob/c3aaf92e767c0844c698701a9c18d15d19bf60ba/tests/core/data/test_data_node.py<br>
> ```python
> class TestData:
>    def testDataNode:
>        dn = InMemoryDataNode("dn", Scope.SCENARIO)
>        assert dn.is_up_to_date == False
> ```

I've added the data node in the test, but it's showing the error(ss I've sent in the above comment)

taipy/core/data/data_node.py
```diff
@@ -511,19 +511,21 @@ def is_up_to_date(self) -> bool:
...
+            parent_scenarios: Set[Scenario] = get_parents(self)["scenario"]  # type: ignore
+            for parent_scenario in parent_scenarios:
+                for ancestor_node in nx.ancestors(parent_scenario._build_dag(), self):
+                    if (
+                        isinstance(ancestor_node, DataNode)
+                        and ancestor_node.last_edit_date
+                        and ancestor_node.last_edit_date > self.last_edit_date
+                    ):
+                        return False
+        return False
```
trgiangdo: The logic of this block will always return `False`.

It should be like:
```suggestion
        if self.is_valid:
            from ..scenario.scenario import Scenario
            from ..taipy import get_parents

            parent_scenarios: Set[Scenario] = get_parents(self)["scenario"]  # type: ignore
            for parent_scenario in parent_scenarios:
                for ancestor_node in nx.ancestors(parent_scenario._build_dag(), self):
                    if (
                        isinstance(ancestor_node, DataNode)
                        and ancestor_node.last_edit_date
                        and ancestor_node.last_edit_date > self.last_edit_date
                    ):
                        return False
            return True

        return False
```

trgiangdo:
Then, I would add this test to the `tests/core/data/test_data_node.py`, right above the 2 tests that are failing.

```python
    def test_is_up_to_date_when_not_written(self):
        dn_confg_1 = Config.configure_in_memory_data_node("dn_1", default_data="a")
        dn_confg_2 = Config.configure_in_memory_data_node("dn_2")
        task_config_1 = Config.configure_task("t1", funct_a_b, [dn_confg_1], [dn_confg_2])
        scenario_config = Config.configure_scenario("sc", [task_config_1])

        scenario = tp.create_scenario(scenario_config)

        assert scenario.dn_1.is_up_to_date is True
        assert scenario.dn_2.is_up_to_date is False

        tp.submit(scenario)
        assert scenario.dn_1.is_up_to_date is True
        assert scenario.dn_2.is_up_to_date is True
```

The idea is to have 2 data nodes:
- the `dn_1` has the default_data, which will be written to the data node right at the creation. Because of that, the `scenario.dn_1.is_up_to_date` should be True.
- the `dn_2` is only written when the scenario is submitted, so the `scenario.dn_2.is_up_to_date` should be False before the submission, and True after.

yaten2302:
Hey @trgiangdo , sorry for inconvenience, actually when this issue was created, the code which I've copied was given in the description, and that's why I copied that. I'll change the code and add the tests as well, as told by you.

trgiangdo:
Thanks for your feedback.

The code in the issue description is just a prototype and not actually tested. The one who is assigned to the issue needs to handle the actual implementation and test the fix.

taipy/core/data/data_node.py
yaten2302: Fixed error in `is_up_to_date` function in data_node.py

yaten2302:
> Thanks for your feedback.
> 
> The code in the issue description is just a prototype and not actually tested. The one who is assigned to the issue needs to handle the actual implementation and test the fix.

Sure, I'll take care of this while contributing to other issues in this project 👍

yaten2302:
Also, I've added the code and the tests as well, it's working now 👍, only 1 test is getting failed (which was already present) - `test_is_up_to_date_across_scenarios`. Rest everything is working fine.

trgiangdo:
What do you mean the failed test `test_is_up_to_date_across_scenarios` is already presented?
The test is not failing before and if it's failing now, it's due to the changes in your branch.

Also, there is a linter error:
```
tests/core/data/test_data_node.py:77:1: W293 [*] Blank line contains whitespace
```
where you left some redundant whitespaces at the end of the line.

yaten2302:
Sorry for the inconvenienced caused, ig the error was due to my mistake I accidentally left those white spaces in my code, I'll take care of it from the next time👍

trgiangdo (approved):
I added 2 commits, 1 fix the whitespaces, and 1 fix the return in the wrong scope. The tests are all passing now.

Thank you again for your contribution.

jrobinAV (approved):
Neat!
Thank you very much for your contribution.

PR #1227, opened 2024-04-24
feature/#1099 added check instance types in equal functions

### LINKED ISSUE #1099: When comparing Scenario to other, check that other isinstance Scenario
**Description**
When building a frontend with taipy, I often have a `selected_scenario` state variable which is `Optional[Scenario]`. Now, when I use an expression like `selected_scenario in tp.get_scenarios()`, I have to prepend a `isinstance(selected_scenario, Scenario) and` to avoid an exception.

Example as follows:

```python
from taipy import Config, Scenario
import taipy as tp
from typing import Optional

x_cfg = Config.configure_data_node(id="x")
y_cfg = Config.configure_data_node(id="y")

def square_fn(x: int) -> int: x**2
square_task_cfg = Config.configure_task(id="square", function=square_fn, input=[x_cfg], output=[y_cfg])

scenario_cfg = Config.configure_scenario(id="scenario", task_configs=[square_task_cfg])

if __name__ == "__main__":
    tp.Core().run()
    scenario = tp.create_scenario(scenario_cfg)

    # __eq__
    selected_scenario: Optional[Scenario] = None
    try:
        selected_scenario == scenario
    except AttributeError as ae:
        print(ae)  # 'NoneType' object has no attribute 'id'

    # in, indirectly calls __eq__
    selected_scenario: Optional[Scenario] = None
    try:
        selected_scenario in tp.get_scenarios()
    except AttributeError as ae:
        print(ae)  # 'NoneType' object has no attribute 'id'
```

Would it be appropriate to add a check in `Scenario.__eq__` for this? Same question for the other entities too

**Acceptance Criteria**

toan-quach:
#1099

Changes:
- Add a check in Entities's `__eq__` for type instance

### DIFF
--- a/taipy/core/cycle/cycle.py
+++ b/taipy/core/cycle/cycle.py
@@ -157,7 +157,7 @@ class Cycle(_Entity, _Labeled):
         raise AttributeError(f"{attribute_name} is not an attribute of cycle {self.id}")
 
     def __eq__(self, other):
-        return self.id == other.id
+        return isinstance(other, Cycle) and self.id == other.id
 
     def __hash__(self):
         return hash(self.id)
--- a/tests/core/scenario/test_scenario.py
+++ b/tests/core/scenario/test_scenario.py
@@ -32,6 +32,21 @@ from taipy.core.task._task_manager_factory import _TaskManagerFactory
 from taipy.core.task.task import Task, TaskId
 
 
+def test_scenario_equals(scenario):
+    scenario_manager = _ScenarioManagerFactory()._build_manager()
+
+    scenario_id = scenario.id
+    scenario_manager._set(scenario)
+
+    # To test if instance is same type
+    task = Task("task", {}, print, [], [], scenario_id)
+
+    scenario_2 = scenario_manager._get(scenario_id)
+    assert scenario == scenario_2
+    assert scenario != scenario_id
+    assert scenario != task
+
+
 def test_create_primary_scenario(cycle):
     scenario = Scenario("foo", set(), {"key": "value"}, is_primary=True, cycle=cycle)
     assert scenario.id is not None
--- a/taipy/core/data/data_node.py
+++ b/taipy/core/data/data_node.py
@@ -277,7 +277,7 @@ class DataNode(_Entity, _Labeled):
         return {key: value for key, value in self.properties.items() if key not in self._TAIPY_PROPERTIES}
 
     def __eq__(self, other):
-        return self.id == other.id
+        return isinstance(other, DataNode) and self.id == other.id
 
     def __ne__(self, other):
         return not self == other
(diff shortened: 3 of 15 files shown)


taipy/core/cycle/cycle.py
```diff
@@ -157,7 +157,7 @@ def __getattr__(self, attribute_name):
         raise AttributeError(f"{attribute_name} is not an attribute of cycle {self.id}")
 
     def __eq__(self, other):
-        return self.id == other.id
+        return isinstance(other, Cycle) and self.id == other.id
```
trgiangdo, line 160: should we raise an error instead of just False if the `other` is not a Cycle?

trgiangdo (dismissed):
LGTM. I just have a question

taipy/core/cycle/cycle.py
jrobinAV: I would not raise an error. This code looks correct to me.

tests/core/scenario/test_scenario.py
```diff
@@ -32,6 +32,21 @@
...
+    scenario_manager._set(scenario)
+
+    # To test if instance is same type
+    task = Task("task", {}, print, [], [], scenario_id)
+
+    scenario_2 = scenario_manager._get(scenario_id)
+    assert scenario == scenario_2
+    assert scenario != scenario_id
+    assert scenario != task
+
```
jrobinAV, line 48:
```suggestion
We should add a test closer to our user's case using optionals.

```suggestion
    scenario = tp.create_scenario(scenario_cfg)

    # __eq__
    selected_scenario: Optional[Scenario] = None
    try:
        selected_scenario == scenario
    except AttributeError as ae:
        print(ae)  # 'NoneType' object has no attribute 'id'

    # in, indirectly calls __eq__
    selected_scenario: Optional[Scenario] = None
    try:
        selected_scenario in tp.get_scenarios()
    except AttributeError as ae:
        print(ae)  # 'NoneType' object has no attribute 'id'
```

What do you think?

jrobinAV (commented):
I just have a proposal to enhance the test suite.

taipy/core/cycle/cycle.py
toan-quach: same, I wouldn't raise an error also, the purpose is to check if this `other` value is equal to the Cycle, if it's not, then the right thing to do is return False. We don't need to raise any errors for it

tests/core/scenario/test_scenario.py
toan-quach: I think all the cases have been covered (None, same Scenario, different type instance). Not sure if it's needed :/

jrobinAV: I don't see where we check for None values.

jrobinAV (dismissed):
Minor test improvement about asserting `scenario != None` values<

toan-quach:
@jrobinAV 
I actually added but pre-commit doesn't allow it and `type: ignore` doesn't help also :/

jrobinAV (approved):

PR #1228, opened 2024-04-24
Fix/#1202 - Catch cookiecutter.OutputDirExistsException and exit

### LINKED ISSUE #1202: BUG-taipy create fails if the project folder already exists
**Description**
`taipy create` command fails if the project folder already exists.

**How to reproduce**

```sh
taipy create 
# accept defaults. Default root folder name is 'taipy_application'

# Run the create again and accept defaults
taipy create
```

Fails with error -

```sh
❯ taipy create --template=scenario-management
Application root folder name [taipy_application]:
Application main Python file [main.py]:
Application title [Taipy Application]:
Does the application use TOML Config? [No]:
Traceback (most recent call last):
  File "/Users/anunayasri/workspace/taipy/./extaipy.py", line 8, in <module>
    sys.exit(_entrypoint())
             ^^^^^^^^^^^^^
  File "/Users/anunayasri/workspace/taipy/taipy/_entrypoint.py", line 62, in _entrypoint
    _ScaffoldCLI.handle_command()
  File "/Users/anunayasri/workspace/taipy/taipy/_cli/_scaffold_cli.py", line 49, in handle_command
    cookiecutter(cls._TEMPLATE_MAP[args.template])
  File "/Users/anunayasri/workspace/taipy/.venv/lib/python3.11/site-packages/cookiecutter/main.py", line 114, in cookiecutter
    result = generate_files(
             ^^^^^^^^^^^^^^^
  File "/Users/anunayasri/workspace/taipy/.venv/lib/python3.11/site-packages/cookiecutter/generate.py", line 291, in generate_files
    project_dir, output_directory_created = render_and_create_dir(
                                            ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/anunayasri/workspace/taipy/.venv/lib/python3.11/site-packages/cookiecutter/generate.py", line 223, in render_and_create_dir
    raise OutputDirExistsException(msg)
cookiecutter.exceptions.OutputDirExistsException: Error: "taipy_application" directory already exists


**Expected behavior**
 It should exit gracefully, after declaring that the folder already exists. Or it can ask for a different folder name if the folder already exists.


**Runtime environment**
 - Taipy version (or branch name): `develop`
 - OS: Mac OS

### PR DESCRIPTION
**Pull request by trgiangdo**
Resolves #1202

### DIFF
--- a/taipy/_cli/_scaffold_cli.py
+++ b/taipy/_cli/_scaffold_cli.py
@@ -50,6 +50,6 @@ class _ScaffoldCLI(_AbstractCLI):
             cookiecutter(cls._TEMPLATE_MAP[args.template])
         except OutputDirExistsException as err:
             error_msg = f"{str(err)}. Please remove the existing directory or provide a new folder name."
-            print(error_msg)
+            print(error_msg)  # noqa: T201
             sys.exit(1)
         sys.exit(0)
--- a/taipy/_cli/_scaffold_cli.py
+++ b/taipy/_cli/_scaffold_cli.py
@@ -49,7 +49,7 @@ class _ScaffoldCLI(_AbstractCLI):
         try:
             cookiecutter(cls._TEMPLATE_MAP[args.template])
         except OutputDirExistsException as err:
-           error_msg = f"{str(err)}. Please remove the existing directory or provide a new folder name."
-           print(error_msg)
-           sys.exit(1)
+            error_msg = f"{str(err)}. Please remove the existing directory or provide a new folder name."
+            print(error_msg)
+            sys.exit(1)
         sys.exit(0)
--- a/taipy/_cli/_scaffold_cli.py
+++ b/taipy/_cli/_scaffold_cli.py
@@ -12,6 +12,7 @@
 import pathlib
 import sys
 
+from cookiecutter.exceptions import OutputDirExistsException
 from cookiecutter.main import cookiecutter
 
 import taipy
@@ -45,6 +46,10 @@ class _ScaffoldCLI(_AbstractCLI):
         args = cls._parse_arguments()
         if not args:
             return
-
-        cookiecutter(cls._TEMPLATE_MAP[args.template])
+        try:
+            cookiecutter(cls._TEMPLATE_MAP[args.template])
+        except OutputDirExistsException as err:
+            error_msg = f"{str(err)[7:]}. Please remove the existing directory or provide a new folder name."
+            cls._logger.error(error_msg)
+            sys.exit(1)
         sys.exit(0)
--- a/taipy/_cli/_scaffold_cli.py
+++ b/taipy/_cli/_scaffold_cli.py
@@ -51,5 +51,5 @@ class _ScaffoldCLI(_AbstractCLI):
         except OutputDirExistsException as err:
            error_msg = f"{str(err)}. Please remove the existing directory or provide a new folder name."
            print(error_msg)
-            sys.exit(1)
+           sys.exit(1)
         sys.exit(0)
--- a/taipy/_cli/_scaffold_cli.py
+++ b/taipy/_cli/_scaffold_cli.py
@@ -49,7 +49,7 @@ class _ScaffoldCLI(_AbstractCLI):
         try:
             cookiecutter(cls._TEMPLATE_MAP[args.template])
         except OutputDirExistsException as err:
-            error_msg = f"{str(err)[7:]}. Please remove the existing directory or provide a new folder name."
-            cls._logger.error(error_msg)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: taipy/_cli/_scaffold_cli.py [hunk 1]
```diff
@@ -45,6 +46,10 @@ def handle_command(cls):
         args = cls._parse_arguments()
         if not args:
             return
-
-        cookiecutter(cls._TEMPLATE_MAP[args.template])
+        try:
+            cookiecutter(cls._TEMPLATE_MAP[args.template])
+        except OutputDirExistsException as err:
+            error_msg = f"{str(err)[7:]}. Please remove the existing directory or provide a new folder name."
+            cls._logger.error(error_msg)
```
jrobinAV: Can't we simply get rid of the cookie-cutter exception?

Plus I believe we don't need to use the taipy logger here. There is no application running. We just exit the command line with a message.

trgiangdo: We can not get the `output_dir` since it's handle by cookiecutter and is only returned if the command success.

jrobinAV: Oh, right, it is not part of the command line arguments. It is in the answer to the application name question!

trgiangdo: That may be to vague though, because we do not mention the folder name in the log. It would look weird, something like:
```
Application root folder name [taipy_application]: a
Application main Python file [main.py]: 
Application title [Taipy Application]: 
Page names in multi-page application? []: 
Does the application use scenario management or version management? [No]: 
Does the application use Rest API? [No]: 
[2024-04-25 14:32:17.716][Taipy][ERROR] Please remove the existing directory or provide a new folder name.
```

There is nothing indicate that the "a" folder exists, but only a message said to remove the existing folder.

jrobinAV: That's true. And what is the output of this piece of code: `{str(err)[7:]}`?

trgiangdo: In cookiecutter, the message is defined as
```python
f'Error: "{dir_to_create}" directory already exists'
```

So the `[7:]` is to remove the "Error: " part of the message

jrobinAV: OK, so indeed that's the right message.

```suggestion
            error_msg = f"{str(err)[7:]}. Please remove the existing directory or provide a new folder name."
            print(error_msg)
```

trgiangdo: I use the `logger.error` so that we can indicate this is an error raised by Taipy. Print would not have the same effect and we should keep the `"Error: "` part of the message

jrobinAV: OK, I got it now, and I agree. Thank you.

However, since this is for a command line and not for runtime, I would still not use a logger but the print function directly. 
That means we don't need to remove the 'Error' part of the cookiedcutter message.

 ```suggestion
            error_msg = f"{str(err)}. Please remove the existing directory or provide a new folder name."
            print(error_msg)
```

What do you think?

Comment (jrobinAV) [reply taipy/_cli/_scaffold_cli.py]: SOrry about that. We sent our messages at the same time.

Comment (trgiangdo) [reply taipy/_cli/_scaffold_cli.py]: Yes that makes sense

File: taipy/_cli/_scaffold_cli.py [hunk 2]
```diff
@@ -45,6 +46,10 @@ def handle_command(cls):
...
         if not args:
             return
-
-        cookiecutter(cls._TEMPLATE_MAP[args.template])
+        try:
+            cookiecutter(cls._TEMPLATE_MAP[args.template])
+        except OutputDirExistsException as err:
+           error_msg = f"{str(err)}. Please remove the existing directory or provide a new folder name."
+           print(error_msg)
+            sys.exit(1)
```
jrobinAV, line 54:
```suggestion
           sys.exit(1)
```

jrobinAV (dismissed):

taipy/_cli/_scaffold_cli.py
```diff
@@ -45,6 +46,10 @@ def handle_command(cls):
...
         if not args:
             return
-
-        cookiecutter(cls._TEMPLATE_MAP[args.template])
+        try:
+            cookiecutter(cls._TEMPLATE_MAP[args.template])
+        except OutputDirExistsException as err:
+           error_msg = f"{str(err)}. Please remove the existing directory or provide a new folder name."
+           print(error_msg)
+           sys.exit(1)
```
jrobinAV:
```suggestion
            error_msg = f"{str(err)}. Please remove the existing directory or provide a new folder name."
            print(error_msg)
            sys.exit(1)
```

jrobinAV (approved):
Neat

PR #1229, opened 2024-04-24
Fixed File Upload Notification 

### LINKED ISSUE #1060: BUG - Using File Upload with a large file generates a large notification
## **Description**
When uploading a file using the file selector visual element on a large file (bigger than 10Mb), Taipy will create a very long success notification with `file_name,file_name,file_name,file_name,file_name,file_name,file_name Uploaded successfully` Here is an example with a 200Mb dataset:

image

## **How to reproduce**

Use this example Taipy app:
```python
from taipy.gui import Gui

content = ""

page = """
<|{content}|file_selector|>
"""

Gui(page).run()
```

and upload this example 8Mb dataset:
[example_data.csv](https://github.com/Avaiga/taipy/files/14790078/example_data.csv)

This will result in the notification issue:
image

## **Expected behavior**
The notification should only display the file name once

## **Runtime environment**
 - Taipy version: 3.1.0
 - OS: Windows
 - Browser: Chrome

Satoshi-Sh:
## Related Issue 
closes #1060  
## Update 
Moved uploadedFiles.push(blob.name) outside of while loop
## Result 
Tested locally 
notification-demo

### DIFF
--- a/frontend/taipy-gui/src/workers/fileupload.worker.ts
+++ b/frontend/taipy-gui/src/workers/fileupload.worker.ts
@@ -82,8 +82,8 @@ const process = (files: FileList, uploadUrl: string, varName: string, id: string
 
                 start = end;
                 end = start + BYTES_PER_CHUNK;
-                uploadedFiles.push(blob.name);
             }
+            uploadedFiles.push(blob.name);
         }
         self.postMessage({
             progress: 100,


trgiangdo:
Hello @FredLL-Avaiga, can you take a look at this fix from @Satoshi-Sh for the bug in #1060?

FredLL-Avaiga (approved):
Good catch, thank you @Satoshi-Sh

PR #1245, opened 2024-04-29
scenario selector list of scenarios

### LINKED ISSUE #1040: Scenario-selector - provide an explicit list of scenarios
**Improvement:**
Adding the possibility for a developer to filter scenarios by scenario_config can be useful in the scenario selector.

**Possible solutions:**
Here are some alternatives:

1. Having a simple optional property to pass the scenario_config
    
    `<|{scenario}|scenario_selector|scenario_config=my_scenario_config_id|>`

2. Having an optional property to pass a custom filter method to apply to the list of scenarios to display.
    
    `<|{scenario}|scenario_selector|filter=my_filter|>`

    Here is an example of a filter method that a user can build:
    ```python
    def my_filter(scenario):
        if scenario.config_id == "my_scenario_config_id":
            return True
        return False
    ```

3. Having an optional lov-like property allows the user to pass a scenario list. By default, the property should be bound to the result of `tp.get_scenarios()` if the property is not defined.

    `<|{scenario}|scenario_selector|lov=scenarios|>`

     ```python
    scenarios = [sc for sc in tp.get_scenarios() if sc.config_id == "my_scenario_config_id"]
    ```

FredLL-Avaiga:
Allows user to specify a list of scenarios/Cycles to display in the scenario selector
resolves #1040 

Also fix a bug in CoreSelector that would prevent to expand branches :-(

`<|{scenario}|scenario_selector|scenarios={get_scenarios()}|>`

### DIFF
--- a/frontend/taipy/src/CoreSelector.tsx
+++ b/frontend/taipy/src/CoreSelector.tsx
@@ -267,12 +267,12 @@ const CoreSelector = (props: CoreSelectorProps) => {
     useDispatchRequestUpdateOnFirstRender(dispatch, id, module, updateVars, undefined, true);
 
     const onItemExpand = useCallback((e: SyntheticEvent, itemId: string, expanded: boolean) => {
-        setExpandedItems(old => {
+        setExpandedItems((old) => {
             if (!expanded) {
-                return old.filter(id => id != itemId);
+                return old.filter((id) => id != itemId);
             }
             return [...old, itemId];
-        })
+        });
     }, []);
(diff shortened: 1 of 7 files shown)


FabienLelaquais (dismissed):
Brilliant.
Can you mark the doc for 'scenarios' as 'TODO' please? That's not quite clear yet, and this would remind me to add an example. Thx

FabienLelaquais (approved):
👍

PR #1247, opened 2024-04-30
Add gitignore and git init to templates

### LINKED ISSUE #671: Add .gitignore and init git repo for taipy cli
When using django cli or reactjs cli to generate a project, it always comes with a folder that already has git initialized along with a .gitignore. This is very convenient to have in taipy cli as well

anunayasri:
Fixes #671


trgiangdo (commented):
Thank you @anunayasri for your contribution.

The PR looks really good to me. I would like to propose 2 additional changes:
1. The content of `.gitignore` is fully loaded with usual Python files to be ignored. However, there is no `.vscode` and Vscode related file. Furthermore, there are a lot of ignore lines for packages that are not closely related to Taipy, like "Translations", "Scrapy", "pdm", "Sphinx", ... We may need a discussion to decide what go into this file to make it suitable for a simple template.
2. In `tests/templates/test_default_template.py` and `tests/templates/test_scenario_mgt_template.py`, we do sometimes test for which files and folders are generated after running the `taipy run ...` command. Those tests now should take the new `.gitignore` and `.git` into account as well.

anunayasri:
@trgiangdo Apologies! I was not able to reply to the comment earlier. 

I made the `.gitignore` based on @jrobinAV 's comment [here](https://github.com/Avaiga/taipy/issues/671#issuecomment-1910262020). I also agree with the same. Different users use different tools. It's meaningful to have a generic `.gitignore` that can serve most users. Users can remove the sections that are unnecessary to them.

I will make the changes for the 2nd comment.

taipy/templates/default/{{cookiecutter.__root_folder_name}}/.gitignore
jrobinAV, line 1: Compared to the [official template](https://github.com/github/gitignore/blob/main/Python.gitignore), you remove a few sections. 

I would not remove them to keep the file as standard as possible.

What do you think?

jrobinAV (commented):
Just a small remark. Otherwise, it is great. Thank you.

PR #1256, opened 2024-05-02
allows to have 2 instances of the same core viz element in a page 

FredLL-Avaiga:
with different configuration/error message !

WiP scenario_selector.filter_by

### DIFF
--- a/frontend/taipy-gui/packaging/taipy-gui.d.ts
+++ b/frontend/taipy-gui/packaging/taipy-gui.d.ts
@@ -117,6 +117,21 @@ export interface TableProps extends TaipyPaginatedTableProps {
 }
 export declare const Table: (props: TableProps) => JSX.Element;
 
+export interface FilterDesc {
+    col: string;
+    action: string;
+    value: string | number | boolean | Date;
+}
+export interface TableFilterProps {
+    columns: Record<string, ColumnDesc>;
+    colsOrder?: Array<string>;
+    onValidate: (data: Array<FilterDesc>) => void;
+    appliedFilters?: Array<FilterDesc>;
+    className?: string;
+    filteredCount: number;
+}
+export declare const TableFilter: (props: TableFilterProps) => JSX.Element;
+
 export declare const Router: () => JSX.Element;
 
 /**
--- a/taipy/gui/extension/library.py
+++ b/taipy/gui/extension/library.py
@@ -159,12 +159,17 @@ class Element:
         properties: t.Optional[t.Dict[str, t.Any]],
         lib: "ElementLibrary",
         is_html: t.Optional[bool] = False,
-        counter: int = 0
+        counter: int = 0,
     ) -> t.Union[t.Any, t.Tuple[str, str]]:
         attributes = properties if isinstance(properties, dict) else {}
         if self.inner_properties:
             uniques: t.Dict[str, int] = {}
-            self.attributes.update(self.inner_properties)
+            self.attributes.update(
+                {
+                    prop: ElementProperty(attr.property_type, None, attr._js_name, attr.with_update)
+                    for prop, attr in self.inner_properties.items()
+                }
+            )
             for prop, attr in self.inner_properties.items():
                 val = attr.default_value
                 if val:
@@ -184,7 +189,9 @@ class Element:
                         if id is None:
                             id = len(uniques) + 1
                             uniques[m.group(1)] = id
-                        val = f"{val[: m.start()]}'{counter}.{id}'{val[m.end() :]}"
+                        val = f"{val[: m.start()]}{counter}{id}{val[m.end() :]}"
+                        if gui._is_expression(val):
+                            gui._evaluate_expr(val, True)
 
                 attributes[prop] = val
         # this modifies attributes
--- a/taipy/gui/utils/_adapter.py
+++ b/taipy/gui/utils/_adapter.py
@@ -12,6 +12,7 @@
 from __future__ import annotations
 
 import typing as t
+from operator import add
 
 from .._warnings import _warn
 from ..icon import Icon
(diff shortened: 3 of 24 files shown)


taipy/gui/utils/_evaluator.py
```diff
@@ -105,7 +108,15 @@ def _analyze_expression(self, gui: Gui, expr: str) -> t.Tuple[t.Dict[str, t.Any]
                     var_name = node.id.split(sep=".")[0]
                     if var_name not in args and var_name not in targets and var_name not in non_vars:
                         try:
-                            encoded_var_name = gui._bind_var(var_name)
+                            if lazy_declare and var_name.startswith("__"):
```
dinhlongviolin1, line 111: Really nice!

taipy/gui/utils/_evaluator.py
```diff
@@ -105,7 +108,15 @@ def _analyze_expression(self, gui: Gui, expr: str) -> t.Tuple[t.Dict[str, t.Any]
...
+                                with warnings.catch_warnings(record=True) as warns:
+                                    warnings.resetwarnings()
+                                    encoded_var_name = gui._bind_var(var_name)
+                                    if len(warns):
```
dinhlongviolin1: Not sure if it is right but we might want to catch specific warnings instead of just checking for the len

taipy/gui/utils/_evaluator.py
```diff
@@ -105,7 +108,15 @@ def _analyze_expression(self, gui: Gui, expr: str) -> t.Tuple[t.Dict[str, t.Any]
...
+                                        gui._get_locals_bind()[var_name] = None
```
dinhlongviolin1: correct me if I am wrong but var_name might not be in the locals since it is coming from the user

taipy/gui/utils/_adapter.py
```diff
@@ -129,8 +130,8 @@ def _run(
                 if not id_only and len(tpl_res) > 2 and isinstance(tpl_res[2], list) and len(tpl_res[2]) > 0:
                     tpl_res = (tpl_res[0], tpl_res[1], self.__on_tree(adapter, tpl_res[2]))
                 return (
-                    (tpl_res + result[len(tpl_res) :])
-                    if isinstance(result, tuple) and isinstance(tpl_res, tuple)
+                    add(type(result)(tpl_res), result[len(tpl_res) :])
```
dinhlongviolin1, line 133: nice nice!!!

taipy/gui_core/_adapters.py
```diff
@@ -177,32 +179,32 @@ def _is_tabular_data(datanode: DataNode, value: t.Any):
         return False
 
     def __get_data(self, dn: DataNode):
-            if dn._last_edit_date:
-                if isinstance(dn, _TabularDataNodeMixin):
+        if dn._last_edit_date:
```
dinhlongviolin1, line 182: might need to document the return types a little bit

taipy/gui_core/_adapters.py
```diff
@@ -177,32 +179,32 @@ def _is_tabular_data(datanode: DataNode, value: t.Any):
...
-                    return (None, None, None, f"read data_node: {e}")
-            return (None, None, None, f"Data unavailable for {dn.get_simple_label()}")
+                val_type = (
+                    "date"
+                    if "date" in type(value).__name__
+                    else type(value).__name__
+                    if isinstance(value, Number)
+                    else None
+                )
+                if isinstance(value, float):
```
dinhlongviolin1: we can merge the if conditions

taipy/gui/extension/library.py
```diff
@@ -159,12 +159,17 @@ def _call_builder(
...
         lib: "ElementLibrary",
         is_html: t.Optional[bool] = False,
-        counter: int = 0
+        counter: int = 0,
     ) -> t.Union[t.Any, t.Tuple[str, str]]:
         attributes = properties if isinstance(properties, dict) else {}
         if self.inner_properties:
             uniques: t.Dict[str, int] = {}
-            self.attributes.update(self.inner_properties)
+            self.attributes.update(
```
dinhlongviolin1, line 167: is it to copy the object over and not have the same object

dinhlongviolin1 (dismissed):
LGTM

frontend/taipy-gui/packaging/taipy-gui.d.ts
```diff
@@ -117,6 +117,21 @@ export interface TableProps extends TaipyPaginatedTableProps {
 }
 export declare const Table: (props: TableProps) => JSX.Element;
 
+export interface FilterDesc {
+    col: string;
+    action: string;
+    value: string | number | boolean | Date;
```
FabienLelaquais, line 123: Will this be a problem in the future when supporting filters on enumerated values?
Or string will do uh?

FabienLelaquais (dismissed):
Nothing wrong I can spot.
Really, really nice job!

taipy/gui/utils/_evaluator.py
FredLL-Avaiga: it's coming from the vis elements so we main as well have them in locals (and there won't be any clash with the leading double underscore)

taipy/gui_core/_adapters.py
FredLL-Avaiga: all documented in the front end

taipy/gui/extension/library.py
FredLL-Avaiga: inner properties use the default value of Element Property to hold the expression ... we don't want those strings has `real` default values

taipy/gui_core/_adapters.py
FredLL-Avaiga: you're right

taipy/gui/utils/_evaluator.py
FredLL-Avaiga: again you're right

taipy/gui/utils/_evaluator.py
FredLL-Avaiga: I've removed it as it is not needed

frontend/taipy-gui/packaging/taipy-gui.d.ts
FredLL-Avaiga: I guess it will do as value holds the choice of the user
The enumeration will have to go into the colDesc

dinhlongviolin1 (approved):
Amazingly done!

PR #1277, opened 2024-05-10
feature/#1197 pass job as metadata for event

### LINKED ISSUE #1197: BUG- Scenario Viewer - Job is not accessible in on_submission_change
**Description**
There are no keys for the 'job' in the *details* variable. We cannot provide correct user feedback.

This is not in line with the doc.
image

**How to reproduce**

```python
from taipy.config import Config
import taipy as tp
from taipy.gui import Gui, notify

Config.configure_job_executions(mode="standalone")


def double(nb):
    return nb * 2


# Configuration of Data Nodes
input_cfg = Config.configure_data_node(id="somedata", default_data=21)
output_cfg = Config.configure_data_node(id="result")


# Configuration of tasks
first_task_cfg = Config.configure_task(id="double",
                                       function=double,
                                       input=input_cfg,
                                       output=output_cfg)


# Configuration of scenario
scenario_cfg = Config.configure_scenario(id="my_scenario",
                                         task_configs=[first_task_cfg],
                                         name="my_scenario")



def notify_from_submissions(state, submittable, details):
    print(details.keys())

if __name__=="__main__":
    tp.Core().run()
    scenario_1 = tp.create_scenario(scenario_cfg)

    scenario_md = """
<|{scenario_1}|scenario|on_submission_change=notify_from_submissions|>
"""
    Gui(scenario_md).run()
```

image

**Expected behavior**
There should be the job that changes the status of the scenario.

**Runtime environment**
 - Taipy: 3.1.1

**Acceptance Criteria**

toan-quach:
#1197

### DIFF
--- a/taipy/core/_entity/_reload.py
+++ b/taipy/core/_entity/_reload.py
@@ -106,4 +106,4 @@ def _get_manager(manager: str) -> _Manager:
         "job": _JobManagerFactory._build_manager(),
         "task": _TaskManagerFactory._build_manager(),
         "submission": _SubmissionManagerFactory._build_manager(),
-    }[manager]  # type: ignore
+    }[manager]
--- a/taipy/core/_entity/_reload.py
+++ b/taipy/core/_entity/_reload.py
@@ -61,7 +61,7 @@ def _self_reload(manager: str):
     return __reload
 
 
-def _self_setter(manager, publish=True):
+def _self_setter(manager):
     def __set_entity(fct):
         @functools.wraps(fct)
         def _do_set_entity(self, *args, **kwargs):
@@ -78,11 +78,9 @@ def _self_setter(manager, publish=True):
                 entity = _Reloader()._reload(manager, self)
                 fct(entity, *args, **kwargs)
                 entity_manager._set(entity)
-                if publish:
-                    Notifier.publish(event)
+                Notifier.publish(event)
             else:
-                if publish:
-                    self._in_context_attributes_changed_collector.append(event)
+                self._in_context_attributes_changed_collector.append(event)
 
         return _do_set_entity
 
--- a/taipy/core/_entity/_reload.py
+++ b/taipy/core/_entity/_reload.py
@@ -60,7 +60,7 @@ def _self_reload(manager: str):
     return __reload
 
 
-def _self_setter(manager):
+def _self_setter(manager, publish=True):
     def __set_entity(fct):
         @functools.wraps(fct)
         def _do_set_entity(self, *args, **kwargs):
@@ -77,9 +77,11 @@ def _self_setter(manager):
                 entity = _Reloader()._reload(manager, self)
                 fct(entity, *args, **kwargs)
                 entity_manager._set(entity)
-                Notifier.publish(event)
+                if publish:
+                    Notifier.publish(event)
             else:
-                self._in_context_attributes_changed_collector.append(event)
+                if publish:
+                    self._in_context_attributes_changed_collector.append(event)
 
         return _do_set_entity
 
@@ -104,6 +106,4 @@ def _get_manager(manager: str) -> _Manager:
         "job": _JobManagerFactory._build_manager(),
         "task": _TaskManagerFactory._build_manager(),
         "submission": _SubmissionManagerFactory._build_manager(),
-    }[
-        manager
-    ]  # type: ignore
+    }[manager]  # type: ignore
--- a/taipy/core/_entity/_reload.py
+++ b/taipy/core/_entity/_reload.py
@@ -10,7 +10,6 @@
 # specific language governing permissions and limitations under the License.
 
 import functools
-from typing import Dict
 
 from .._manager._manager import _Manager
 from ..notification import EventOperation, Notifier, _make_event
@@ -97,7 +96,7 @@ def _get_manager(manager: str) -> _Manager:
     from ..submission._submission_manager_factory import _SubmissionManagerFactory
     from ..task._task_manager_factory import _TaskManagerFactory
 
-    manager_dict: Dict[str, _Manager] = {
+    return {
         "scenario": _ScenarioManagerFactory._build_manager(),
         "sequence": _SequenceManagerFactory._build_manager(),
         "data": _DataManagerFactory._build_manager(),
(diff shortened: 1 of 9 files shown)


taipy/core/_entity/_reload.py
```diff
@@ -104,6 +106,4 @@ def _get_manager(manager: str) -> _Manager:
         "job": _JobManagerFactory._build_manager(),
         "task": _TaskManagerFactory._build_manager(),
         "submission": _SubmissionManagerFactory._build_manager(),
-    }[
-        manager
-    ]  # type: ignore
+    }[manager]  # type: ignore
```
jrobinAV, line 107: WHy do we need to add a "# type: ignore" here?

jrobinAV (commented):
- We are missing the unit tests covering the new code on both reload.py and submission.py modules.
- Even if the code looks smart, I don't like giving the notification responsibility to the entity itself. For me, it should be the repository/self_setter layer.


We can discuss and try to challenge your solution to see if we can propose a better design.

taipy/core/_entity/_reload.py
toan-quach: apparently we don't need it!

taipy/core/submission/_submission_manager.py
```diff
@@ -47,6 +53,88 @@ def _create(cls, entity_id: str, entity_type: str, entity_config: Optional[str],
...
+
+        cls._set(submission)
+
+        if _current_submission_status != submission._submission_status:
+            event = _make_event(
+                submission,
+                EventOperation.UPDATE,
+                "submission_status",
+                submission._submission_status,
+                job_triggered_submission_status_changed=job.id,
```
jrobinAV, line 130: This parameter name is exposed to the user through the event. It becomes an attribute of the event. We must find a better name. What about: "job_id"?

toan-quach: a bit vague imho, how about `trigger_job_id`?

taipy/gui_core/_context.py
```diff
@@ -174,7 +174,10 @@ def submission_status_callback(self, submission_id: t.Optional[str]):
                             self.gui._call_user_callback(
                                 client_id,
                                 submission_name,
-                                [core_get(submission.entity_id), {"submission_status": new_status.name}],
+                                [
+                                    core_get(submission.entity_id),
```
jrobinAV: `core_get(submission.entity_id)` is a scenario. I guess we want to pass the submission instead: `core_get(submission.id)`.

@FredLL-Avaiga @toan-quach What do you think?

FredLL-Avaiga: the doc indicates : 1st param is a `submittable` (scenario or sequence)
We could add the `submission` to the detail but I think the main object is rightly the `submittable`.

toan-quach: well, I'm not sure either so I don't have any comments on it

FlorianJacta: This is my comment on this: I want two things generally. I want to know the Status of the job that has changed and the job itself. I also want to know if my whole submission has started, is canceled, or is completed. I want to see the progress of the jobs and how many of them are completed.

toan-quach: @FlorianJacta Flo
in that case, I can add the submission entity as a metadata, what do you think?

FlorianJacta: That could do the trick

taipy/core/_entity/_reload.py
jrobinAV: We should remove it, then.

taipy/core/submission/_submission_manager.py
jrobinAV: it is a submission status change. I can't see any possible confusion using `job_id` instead of `trigger_job_id`. Plus, it is a technical consideration for me to add a prefix like `trigger`.

The event expresses a submission_status change for a submission entity due to a job update. I don't see any possible confusion.

What do you think? and @FlorianJacta?

taipy/gui_core/_context.py
jrobinAV: The event represents an update of `submission_status` attribute of a submission entity.

The main entity must be the submission object both for semantic and consistency reasons. It cannot be the submittable as the main object, otherwise we won't be able to differentiate with an update of an attribute of the submittable itself.

Of course we can add the submittable id in the metadata.

taipy/core/submission/_submission_manager.py
FlorianJacta: I agree

taipy/gui_core/_context.py
toan-quach: @jrobinAV @FredLL-Avaiga @FlorianJacta so do we all agree on:
- changing the `submittable` -> `submission` and  `submission.entity_id` -> `submission.id`? 
- adding `submission.entity_id` to the metadata under the key name `submittable_id`?

jrobinAV: I do. Looks great.

FredLL-Avaiga: I'm still not convinced but if everybody agrees...

taipy/gui_core/_context.py
```diff
@@ -174,7 +174,14 @@ def submission_status_callback(self, submission_id: t.Optional[str]):
                             self.gui._call_user_callback(
                                 client_id,
                                 submission_name,
-                                [core_get(submission.entity_id), {"submission_status": new_status.name}],
+                                [
+                                    core_get(submission.id),
+                                    {
+                                        "submission_status": new_status.name,
+                                        "submittable_entity_id": core_get(submission.entity_id),
+                                        **(event.metadata if event else {}),
```
FredLL-Avaiga, line 182: any possibility that event.metadata holds the other 2 keys ?
why not put it first and then override with the other 2 keys ?

FredLL-Avaiga: I suppose you need to propagate the change to the [viselements.json](https://vscode.dev/github/Avaiga/taipy/blob/develop/taipy/gui_core/viselements.json#L195)

FredLL-Avaiga (dismissed):

taipy/gui_core/_context.py
toan-quach: "any possibility that event.metadata holds the other 2 keys ?" -> hm I don't think so :/

FredLL-Avaiga: 👍

FredLL-Avaiga (approved):

jrobinAV (approved):

PR #1281, opened 2024-05-13
Handle Ngrok better (#1268)

### LINKED ISSUE #1268: BUG-gui.reload creates a new ngrok tunnel which fails when 3 tunnels are already active
**Description**
When working from a remote notebook like on colab or here on databricks, I use `gui.reload` to update my application after I edit the notebook. Unfortunately, `gui.reload` creates a new ngrok tunnel every time it is called. When 3 tunnels are created, ngrok refuses to create another one, forcing me to restart the cluster.

image

**How to reproduce**

On a Databricks notebook, create a Taipy application. You can use this file as an example:
[Industry Agglom Long Call.zip](https://github.com/Avaiga/taipy/files/15220789/Industry.Agglom.Long.Call.zip)
Run all the cells, then run the cell with `gui.reload()` 3 times. This will cause the error

**Expected behavior**
`gui.reload()` (and maybe also `gui.run()`) should create the app using the previously used ngrok tunnel, or create a new one and kill the previous one

**Runtime environment**
 - Taipy version: 3.1.1
 - Databricks

dinhlongviolin1:
Resolve #1268

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -2275,12 +2275,19 @@ class Gui:
 
     def __init_ngrok(self):
         app_config = self._config.config
-        if app_config["run_server"] and app_config["ngrok_token"]:  # pragma: no cover
+        if hasattr(self, "_ngrok"):
+            # Keep the ngrok instance if token has not changed
+            if app_config["ngrok_token"] == self._ngrok[1]:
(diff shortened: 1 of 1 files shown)


FredLL-Avaiga (approved):
Easy one for you

PR #1282, opened 2024-05-13
Remove hover_text from Indicator

### LINKED ISSUE #1147: BUG- hover_text not showing for indicator
**Description**
When adding a hover_text to an indicator, it does not show when hovering over it. (When I use hover_text on a selector instead of indicator, it works, though..)

**How to reproduce**

- A code fragment
    ```python
  
    Gui("""
    #test
    <|message|indicator|value=50|min=0|max=100|hover_text=This does not show|>
    """).run()

    ```

**Expected behavior**
Description of what would be the expected outcome.

**Screenshots**
When available and relevant, screenshots to better highlight the problem.

**Runtime environment**
 - Taipy==3.0.0:
 - OS: MacOS 14.4.1
 - Browser: Chrome

**Acceptance Criteria**

Satoshi-Sh:
## Related Issue 
Fixed #1147
Resolves #1147 

## Updates 
- Added `div` to the Indicator component to capture mouseOver events. 

## Screenshot 

tooltip-demo

## Reference 
I referred to [this link](https://stackoverflow.com/questions/57527896/material-ui-tooltip-doesnt-display-on-custom-component-despite-spreading-props#:~:text=Your%20Tooltip%20is%20not%20working%20properly%20because%20the%20child%20of%20a%20Material%2DUI%20Tooltip%20must%20be%20able%20to%20hold%20a%20ref.)

### DIFF
--- a/frontend/taipy-gui/src/components/Taipy/Indicator.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Indicator.tsx
@@ -13,13 +13,12 @@
 
 import React, { useCallback, useMemo } from "react";
 import Slider from "@mui/material/Slider";
-import Tooltip from "@mui/material/Tooltip";
 import { sprintf } from "sprintf-js";
 
-import { TaipyBaseProps, TaipyHoverProps } from "./utils";
-import { useClassNames, useDynamicProperty } from "../../utils/hooks";
+import { TaipyBaseProps } from "./utils";
+import { useClassNames } from "../../utils/hooks";
 
-interface IndicatorProps extends TaipyBaseProps, TaipyHoverProps {
+interface IndicatorProps extends TaipyBaseProps {
     min?: number;
     max?: number;
     value?: number;
@@ -43,7 +42,6 @@ const Indicator = (props: IndicatorProps) => {
 
     const horizontalOrientation = props.orientation ? props.orientation.charAt(0).toLowerCase() !== "v" : true;
     const className = useClassNames(props.libClassName, props.dynamicClassName, props.className);
-    const hover = useDynamicProperty(props.hoverText, props.defaultHoverText, undefined);
 
     const getLabel = useCallback(() => {
         const dsp = display === undefined ? (defaultDisplay === undefined ? "" : defaultDisplay) : display;
@@ -95,7 +93,6 @@ const Indicator = (props: IndicatorProps) => {
     );
 
     return (
-        <Tooltip title={hover || ""}>
             <Slider
                 id={props.id}
                 className={className}
@@ -109,7 +106,6 @@ const Indicator = (props: IndicatorProps) => {
                 orientation={horizontalOrientation ? undefined : "vertical"}
                 sx={sliderSx}
             ></Slider>
-        </Tooltip>
     );
 };
 
--- a/taipy/gui/viselements.json
+++ b/taipy/gui/viselements.json
@@ -833,6 +833,10 @@
                         "type": "str",
                         "default_value": "None",
                         "doc": "The height, in CSS units, of the indicator (used when orientation is vertical)."
+                    },
+                    {
+                        "name": "hover_text",
+                        "doc": "TODO not implemented"
                     }
                 ]
             }
--- a/taipy/gui/viselements.json
+++ b/taipy/gui/viselements.json
@@ -836,7 +836,8 @@
                     },
                     {
                         "name": "hover_text",
-                        "doc": "TODO not implemented"
+                        "doc": "TODO not implemented",
+                        "hide": true
                     }
                 ]
             }
--- a/contributors.txt
+++ b/contributors.txt
@@ -15,3 +15,4 @@ bobbyshermi
 Forchapeatl
 yarikoptic
 Luke-0162
+Satoshi-Sh
--- a/taipy/gui/_renderers/factory.py
+++ b/taipy/gui/_renderers/factory.py
@@ -57,7 +57,7 @@ class _Factory:
         "text": "value",
         "toggle": "value",
         "tree": "value",
-        "metric": "value"
+        "metric": "value",
     }
 
     _TEXT_ATTRIBUTES = ["format", "id", "hover_text", "raw"]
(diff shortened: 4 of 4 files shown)


frontend/taipy-gui/src/components/Taipy/Indicator.tsx
```diff
@@ -96,19 +96,21 @@ const Indicator = (props: IndicatorProps) => {
...
-                max={100}
-                value={getValue(value === undefined ? defaultValue : value, min, max)}
-                disabled={true}
-                valueLabelDisplay="on"
-                valueLabelFormat={getLabel}
-                marks={marks}
-                orientation={horizontalOrientation ? undefined : "vertical"}
-                sx={sliderSx}
-            ></Slider>
+            <div>
```
FredLL-Avaiga: I think a span would have less impact of the display ?

FredLL-Avaiga (changes requested):
I'm afraid that a simple div would force display block ...

FredLL-Avaiga:
please update your branch
I guess Indicator.spec.tsx needs change ?

frontend/taipy-gui/src/components/Taipy/Indicator.tsx
Satoshi-Sh: I tried with a span and an empy <> but the tooltip doesn't work. 

Do you mean a new test needs to be added to `Indicator.spec.tsx`?

FredLL-Avaiga: a simple div like this is not acceptable as it would change the display behavior of the indicator to block: can't put 2 in the same line.
It looks like a tooltip directly on a Mui Slider (as used in indicator) doesn't work because the slider uses the same technique as the tooltip to show it's own different tooltips.
A div with `inline` display behaves the same as a span ie shows the hover text only below the indicator.
I suppose then that a span is better than nothing but still not great ...
I wonder if we should keep the hover_text all together.
thoughts @FabienLelaquais ?

PS: it looks like the test in Indicator.spec.tsx don't pass anymore ?

FredLL-Avaiga:
We're having a discussion about this with @FabienLelaquais 
We still need to find the right meaning of hover_text for indicator ...
We'll get back here on the beginning of next week

FredLL-Avaiga:
Hi @Satoshi-Sh 
we've had quite a bit of discussion on this with @FabienLelaquais and we came to the conclusion that hover_text doesn't make sense in the context of indicator.
I'm sorry to say that we want to remove it completely.
Are you still willing to perform this task ?

Satoshi-Sh:
Sure, I can work on the task of removing the hover_text.

Satoshi-Sh:
I tried to delete the hover_text from the Indicator component. Could you take a look at it? @FredLL-Avaiga
 
#1147

FredLL-Avaiga (dismissed):
Looks good to me 
Could you also remove it in the [factory.py](https://vscode.dev/github/Avaiga/taipy/blob/feature/%231301-sort-scenario-selector/taipy/gui/_renderers/factory.py#L283) file ?
and also add it in the [viselements.json](https://vscode.dev/github/Avaiga/taipy/blob/feature/%231301-sort-scenario-selector/taipy/gui/viselements.json#L836) file so that it overrides the default description which is in `shared` ?
I suppose you'll also have to check the indicator.spec.tsx ?

Satoshi-Sh:
>and also add it in the [viselements.json](https://vscode.dev/github/Avaiga/taipy/blob/feature/%231301-sort-scenario-selector/taipy/gui/viselements.json#L836) file so that it overrides the default description which is in shared ?

I'm not sure about this instruction. What kind of description should I add to hover_text for the indicator?

FredLL-Avaiga:
Something like : TODO not implemented

Satoshi-Sh:
I just ran `npm test src/components/Taipy/Indicator` and it passed. Let me know if there is a proper way to test it.

FredLL-Avaiga:
> I just ran `npm test src/components/Taipy/Indicator` and it passed. Let me know if there is a proper way to test it. 

Excellent

FredLL-Avaiga (dismissed):
Well done

FredLL-Avaiga:
> > I just ran `npm test src/components/Taipy/Indicator` and it passed. Let me know if there is a proper way to test it.
> 
> Excellent

It looks like the test cannot pass when it is PR submitted from outside the taipy repository.
So I guess you're good to go.
Shall I merge this PR ?

FredLL-Avaiga:
PS: can you add an attribute hide=true to the viselements.json ?

Satoshi-Sh:
> PS: can you add an attribute hide=true to the viselements.json ?

{
                       "name": "hide",
                       "type": "boolean",
                       "default_value": "true"
                    }
 Do I need to add a description for this?

FredLL-Avaiga:
> > PS: can you add an attribute hide=true to the viselements.json ?
> 
> { "name": "hide", "type": "boolean", "default_value": "true" } Do I need to add a description for this?

oh sorry, it's not a new property for the component but a way for our doc system to ignore the property marked with "hide": true

taipy/gui/viselements.json
```diff
@@ -833,6 +833,11 @@
                         "type": "str",
                         "default_value": "None",
                         "doc": "The height, in CSS units, of the indicator (used when orientation is vertical)."
+                    },
+                    {
+                        "name": "hover_text",
+                        "doc": "TODO not implemented",
+                        "hide": true
```
FredLL-Avaiga, line 840: 👍

FredLL-Avaiga (approved):
tx @Satoshi-Sh

FredLL-Avaiga:
I guess this PR is ready to merge. What do you think @Satoshi-Sh ?
May I ?

FabienLelaquais (approved):
👌

Satoshi-Sh:
Yes, it's ready. Please merge it, @FredLL-Avaiga

FredLL-Avaiga:
tx a lot @Satoshi-Sh

PR #1295, opened 2024-05-21
Use lov as int and float for Python API (#1254)

### LINKED ISSUE #1254: BUG-lov of types different from strings not working with Python API
**Description**
Using the Python API without binding the lov variable, the lov can create issues. It creates an issue when the lov is not of type strings like the example below.

image

**How to reproduce**

```python
from taipy.gui import Gui


selected = [2019]
years_list = [2019, 2020, 2021, 2022, 2023]

with tgb.Page() as page:
    tgb.selector(
            value="{selected}",
            lov=years_list,
            dropdown=True,
            multiple=True,)


Gui(page).run()
```

**Expected behavior**
The behavior when the variable is bound and not bound should be the same. The code above should work by using a default adapter.

**Runtime environment**
 - Taipy: 3.1.1

**Acceptance Criteria**

dinhlongviolin1:
Resolve #1254

### DIFF
--- a/taipy/gui/utils/_adapter.py
+++ b/taipy/gui/utils/_adapter.py
@@ -119,8 +119,8 @@ class _Adapter:
                 result = adapter(result)
                 if result is None:
                     return result
-            elif isinstance(result, str):
-                return result
+            elif isinstance(result, (str, float, int)):
+                return str(result)
             tpl_res = self._get_valid_result(result, id_only)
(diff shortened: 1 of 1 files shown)


FredLL-Avaiga (approved):
👍

FabienLelaquais (approved):
if str(str) is free...

PR #1304, opened 2024-05-23
Feature/#405 exposing notification mechanism to user

### LINKED ISSUE #405: Notification: Expose notification mechanism to users
- [ ] Expose register and unregister methods and add Reference manual documentation

The documentation is covered by ticket [doc-#416](https://github.com/Avaiga/taipy-doc/issues/416).

toan-quach:
#405

### DIFF
--- a/taipy/core/notification/core_event_consumer.py
+++ b/taipy/core/notification/core_event_consumer.py
@@ -32,10 +32,18 @@ class CoreEventConsumerBase(threading.Thread):
             print(f"Received event created at : {event.creation_date}")
             pass
 
-    consumer = MyEventConsumer("consumer_1", event_queue)
+    registration_id, registered_queue = Notifier.unregister(
+        entity_type=EventEntityType.CYCLE,
+        entity_id="CYCLE_cycle_1",
+        operation=EventOperation.CREATION
+    )
+
+    consumer = MyEventConsumer("consumer_1", registered_queue)
     consumer.start()
     # ...
     consumer.stop()
+
+    Notifier.unregister(registration_id)
     ```
 
     Subclasses should implement the `process_event` method to define their specific event handling behavior.
--- a/taipy/core/notification/core_event_consumer.py
+++ b/taipy/core/notification/core_event_consumer.py
@@ -23,6 +23,8 @@ class CoreEventConsumerBase(threading.Thread):
     It should be subclassed, and the `process_event` method should be implemented to define
     the custom logic for handling incoming events.
 
+    Subclasses should implement the `process_event` method to define their specific event handling behavior.
+
     Example usage:
 
     ```python
@@ -32,13 +34,12 @@ class CoreEventConsumerBase(threading.Thread):
             print(f"Received event created at : {event.creation_date}")
             pass
 
-    registration_id, registered_queue = Notifier.unregister(
-        entity_type=EventEntityType.CYCLE,
-        entity_id="CYCLE_cycle_1",
+    registration_id, registered_queue = Notifier.register(
+        entity_type=EventEntityType.SCENARIO,
         operation=EventOperation.CREATION
     )
 
-    consumer = MyEventConsumer("consumer_1", registered_queue)
+    consumer = MyEventConsumer(registration_id, registered_queue)
     consumer.start()
     # ...
     consumer.stop()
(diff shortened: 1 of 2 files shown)


taipy/core/notification/core_event_consumer.py
```diff
@@ -32,10 +32,18 @@ def process_event(self, event: Event):
             print(f"Received event created at : {event.creation_date}")
             pass
 
-    consumer = MyEventConsumer("consumer_1", event_queue)
+    registration_id, registered_queue = Notifier.unregister(
```
jrobinAV:
```suggestion
    registration_id, registered_queue = Notifier.register(
```

taipy/core/notification/core_event_consumer.py
```diff
@@ -32,10 +32,18 @@ def process_event(self, event: Event):
...
+
+    consumer = MyEventConsumer("consumer_1", registered_queue)
     consumer.start()
     # ...
     consumer.stop()
+
+    Notifier.unregister(registration_id)
     ```
 
     Subclasses should implement the `process_event` method to define their specific event handling behavior.
```
jrobinAV: This line should be placed before the example.

taipy/core/notification/core_event_consumer.py
```diff
@@ -32,10 +32,18 @@ def process_event(self, event: Event):
             print(f"Received event created at : {event.creation_date}")
             pass
 
-    consumer = MyEventConsumer("consumer_1", event_queue)
+    registration_id, registered_queue = Notifier.unregister(
+        entity_type=EventEntityType.CYCLE,
+        entity_id="CYCLE_cycle_1",
+        operation=EventOperation.CREATION
```
jrobinAV: This topic will never generate an event. We are registering to "the creation of a cycle whose id is CYCLE_cycle_1". Either the cycle already exists, and the event won't be produced, or it will not, and the randomly generated unique ID won't be the one provided.

I would use a more common use case. Why not registering to the creation of any scenario?

taipy/core/notification/core_event_consumer.py
```diff
@@ -32,10 +32,18 @@ def process_event(self, event: Event):
...
+    )
+
+    consumer = MyEventConsumer("consumer_1", registered_queue)
```
jrobinAV:
```suggestion
    consumer = MyEventConsumer(registration_id, registered_queue)
```

taipy/core/notification/notifier.py
```diff
@@ -118,6 +128,18 @@ def register(
...
+
+        ```python
+        registration_id, registered_queue = Notifier.unregister(
+            entity_type=EventEntityType.CYCLE,
+            entity_id="CYCLE_cycle_1",
+            operation=EventOperation.CREATION
+        )
+
+        Notifier.unregister(registration_id)
+        ```
```
jrobinAV: This is not correct. we only provide the registration_id to unregister.

taipy/core/notification/core_event_consumer.py
```diff
@@ -32,10 +32,18 @@ def process_event(self, event: Event):
...
     consumer.start()
     # ...
     consumer.stop()
+
+    Notifier.unregister(registration_id)
     ```
```
jrobinAV, line 49: After the example I would add a short paragraph to explain the code, and particularly
what are the events we are registering. We can also briefly explain the principle of the topic registration and redirect readers to the register function refman page to make it easier to understand how to build the topic.

taipy/core/notification/notifier.py
```diff
@@ -68,6 +68,16 @@ def register(
         The topic is defined by the combination of the entity type, the entity id,
         the operation and the attribute name.
```
jrobinAV:
```suggestion
        The topic is defined by the combination of an optional entity type, an optional 
        entity id, an optional operation, and an optional attribute name. The purpose is
        to be as flexible as possible. For example, we can register to:
        
        - All scenario creations
        - A specific data node update
        - A sequence submission
        - A Scenario deletion
        - Job failures
```

taipy/core/notification/notifier.py
```diff
@@ -68,6 +68,16 @@ def register(
...
+        Example usage:
+
+        ```python
+        registration_id, registered_queue = Notifier.register(
+            entity_type=EventEntityType.CYCLE,
+            entity_id="CYCLE_cycle_1",
+            operation=EventOperation.CREATION
+        )
+        ```
+
```
jrobinAV: I would use a more realistic use case.

taipy/core/notification/core_event_consumer.py
toan-quach: AH yes I totally forgot that! I have changed it to:
```
registration_id, registered_queue = Notifier.register(
        entity_type=EventEntityType.SCENARIO,
        operation=EventOperation.CREATION
    )
```

taipy/core/notification/notifier.py
toan-quach: I believe that's the example no?

jrobinAV: On line 141, it should be `register`. That's why.

toan-quach: ahh fixed! thanks!

jrobinAV (approved):

PR #1336, opened 2024-05-29
navigate and download only for the browser window that requests it

### LINKED ISSUE #1145: BUG- If user has two windows open on the same browser, callbacks (`navigate`, `download`) happen on both windows
**Description**
It's a funny one and pretty sure it happens because we call `navigate` to the same user, and we end up calling it twice
Does not happen if you have two different browsers, because then you'll have two different users ID.

**How to reproduce**
Example code:
```python
from taipy.gui import navigate, Gui

def go_to_google(state):
    navigate(state, to="https://www.google.com/")


page = """
<|button|label=navigate to google|on_action=go_to_google|>
"""

Gui(
    page=page
).run(debug=True, use_reloader=True)

```
Open a browser (in my case it was Firefox) to your Taipy app, open another window to it. Make sure you have TWO separate windows accessing the app **ON THE SAME BROWSER.**

If you click the button while having two windows open on the same browser (Therefore same user session), you'll have two tabs opening to google in each browser window

**Expected behavior**
Clicking on the `navigate to google` button should only open a tab in the window which the button was clicked

**Runtime environment**
 - Taipy 3.1.0
 - Browser: Firefox

FredLL-Avaiga:
resolves #1145

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -1167,14 +1167,14 @@ class Gui:
             }
         )
 
-    def __send_ws(self, payload: dict, allow_grouping=True) -> None:
+    def __send_ws(self, payload: dict, allow_grouping=True, send_back_only=False) -> None:
         grouping_message = self.__get_message_grouping() if allow_grouping else None
         if grouping_message is None:
             try:
                 self._server._ws.emit(
                     "message",
                     payload,
-                    to=self.__get_ws_receiver(),
+                    to=self.__get_ws_receiver(send_back_only),
                 )
                 time.sleep(0.001)
             except Exception as e:  # pragma: no cover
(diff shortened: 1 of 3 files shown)


FabienLelaquais (commented):
🤔

FabienLelaquais (approved):
🤯

taipy/gui/gui.py
```diff
@@ -1193,7 +1193,11 @@ def __broadcast_ws(self, payload: dict, client_id: t.Optional[str] = None):
     def __send_ack(self, ack_id: t.Optional[str]) -> None:
         if ack_id:
             try:
-                self._server._ws.emit("message", {"type": _WsType.ACKNOWLEDGEMENT.value, "id": ack_id})
+                self._server._ws.emit(
+                    "message",
+                    {"type": _WsType.ACKNOWLEDGEMENT.value, "id": ack_id},
+                    to=self.__get_ws_receiver(True),
```
FredLL-Avaiga, line 1199: @dinhlongviolin1 I'd like your input on this ...

dinhlongviolin1: since the ack_id is unique between clients and unique between each call, I dont think that we need to set the `to` parameter right here.

dinhlongviolin1 (approved):
LGTM

namnguyen20999 (approved):

taipy/gui/gui.py
FredLL-Avaiga: But isn't it better to send the message only to the emitter ?

dinhlongviolin1: You are right! if no to parameter is provided it is sent to all clients. That should settle it.

PR #1337, opened 2024-05-30
Cache dependencies on github actions (#1296)

### LINKED ISSUE #1296: Enhance github action speed by caching dependencies
**Description**
Per [github](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows), we can try to cache the dependencies to increase the speed of each execution. 

**Acceptance Criteria**

dinhlongviolin1:
Resolve #1296

Caching for both npm and pipenv (for tests only). Deployment/build will still do everything from scratch.
Caching was also added for frontend build.

For npm, the dependency was declared and those will be used to hash as the cache name. 

For pipenv, the recommended file is `Pipfile.lock`. But since we don't use `Pipfile.lock` (mainly due to multiple os support), `Pipfile` will be used. We might have to manually delete the cache from [here](https://github.com/Avaiga/taipy/actions/caches) if any dev packages get a major/breaking update. Pipenv caching will work both for the min Pipfile and max Pipfile as those will be produce 2 different caches due to the different file hashes.

The cache will be removed automatically after 7 days if not used in any action. There is no limit on the number of cache, but there is a hard limit of 10 GB of cache per repo.


FredLL-Avaiga (dismissed):
I'm learning everyday 👍

FabienLelaquais (dismissed):
So that was impressive! 🤯

.github/workflows/partial-tests.yml
```diff
@@ -64,9 +66,17 @@ jobs:
         if: matrix.os == 'macos-13'
         run: brew install libmagic
 
-      - uses: actions/setup-node@v4
+      - name: Setup Node
+        if: steps.changes.outputs.gui == 'true' || steps.changes.outputs.gui-core == 'true'
```
jrobinAV, line 70: 👍🏻

.github/workflows/partial-tests.yml
```diff
@@ -128,7 +138,17 @@ jobs:
...
+        with:
+          python-version: ${{matrix.python-version}}
+          cache: 'pipenv'
+          cache-dependency-path: '**/Pipfile'
+
+      - name: Setup Python without cache
+        if: steps.changes.outputs.core != 'true'
+        uses: actions/setup-python@v5
         with:
           python-version: ${{matrix.python-version}}
```
jrobinAV: The `submit_tests` job is only run if some core files are touched. This step is run when core files are not touched. I don't get its purpose.

.github/workflows/partial-tests.yml
```diff
@@ -167,7 +187,17 @@ jobs:
...
```
jrobinAV: same here.

.github/workflows/partial-tests.yml
```diff
@@ -128,7 +138,17 @@ jobs:
             core:
               - 'taipy/core/**'
 
-      - uses: actions/setup-python@v5
+      - name: Setup Python with cache
+        if: steps.changes.outputs.core == 'true'
```
jrobinAV, line 157: Good catch. the filter was missing.

jrobinAV (commented):
This PR only covers partial tests. Why not the overall tests as well?

FabienLelaquais (dismissed):
🤯🤯🤯

dinhlongviolin1:
> This PR only covers partial tests. Why not the overall tests as well?

The overall tests uses the install/action.yml file for installation of the dependencies. The cache is already implemented in that file.

.github/workflows/partial-tests.yml
dinhlongviolin1: Thanks!

.github/workflows/partial-tests.yml
dinhlongviolin1: I will remove it. Thanks!

dinhlongviolin1:
> This PR only covers partial tests. Why not the overall tests as well?

We can also remove the python setup within overall-tests.yml file since it will be setup in the install/action.yml file. @florian-vuillemot can you confirm this

jrobinAV (dismissed):

FabienLelaquais (dismissed):
👍👍

florian-vuillemot (dismissed):

dinhlongviolin1:
Caching the frontend build is also added!

FredLL-Avaiga (dismissed):

tools/frontend/hash_source.py
```diff
@@ -0,0 +1,77 @@
...
+
+    combined_hash = combined_hasher.hexdigest()
+    return file_hashes, combined_hash
+
+
+if __name__ == "__main__":
+    frontend_folder = "frontend"
+    file_hashes, combined_hash = hash_files_in_frontend_folder(frontend_folder)
+    # for path, file_hash in sorted(file_hashes.items()):
+    #     print(f"{path}: {file_hash}")
```
jrobinAV, line 71: To remove

jrobinAV (approved):

tools/frontend/hash_source.py
dinhlongviolin1: I will keep that one for local debugging when needed. Thanks for spotting that JR!

tools/frontend/hash_source.py
```diff
@@ -0,0 +1,77 @@
...
+            if any(ignore in root for ignore in ignore_folder):
+                continue
+            # Sort before looping to ensure consistent cache key
+            for file in sorted(files):
+                for ignore in ignore_file_name:
+                    if ignore in file:
+                        continue
+                for ignore in ignore_file_extension:
+                    if file.endswith(ignore):
+                        continue
```
florian-vuillemot: What is the purpose of these two conditions? If these conditions are met, the `continue` is applied to the first `for` and not to both. Consequently, they do nothing. Do you agree?

jrobinAV: Indeed

jrobinAV (dismissed):

joaoandre-avaiga:
LGTM! Great job @dinhlongviolin1

joaoandre-avaiga (dismissed):

jrobinAV:
@dinhlongviolin1 Let's merge it if that's fine for everyone so we can finally benefit from it.

tools/frontend/hash_source.py
dinhlongviolin1: Thanks for the finding @florian-vuillemot. I will refactor it

florian-vuillemot (approved):

FredLL-Avaiga (approved):

jrobinAV (approved):

florian-vuillemot (approved):

joaoandre-avaiga (approved):

PR #1354, opened 2024-06-04
manage multi trace selection when on selected variable is binded

### LINKED ISSUE #1061: Selecting points on a line chart only works for single trace
## Issue
The documentation currently provides a great example of how to get the mean value of selected points on a line chart:
https://docs.taipy.io/en/latest/manuals/gui/viselements/charts/advanced/#tracking-selection
image

Unfortunately, this only works with line charts with one trace. When you have multiple traces, this approach does not work:
image
Here, the user selected points from trace B (red line). When selecting these points, Taipy returns a new value for `selected={selected_indices}.` The dev then has to pre-define which trace he wants to calculate the mean on using:

```python
def on_change(state, var, val):
    if var == "selected_indices":
        state.mean_value = np.mean([data["A"][idx] for idx in val]) if len(val) else 0
```
Here, the dev chose to calculate on trace A, which leads to `mean_value` returning 12.03 instead of 3.5, i.e., returning the mean of points from trace A that were not selected.

## Feature Improvement

Ideally, the `selected` property should not only return indices in the dataset. It should return the actual points that were selected (indices and columns)

## Workaround

The current workaround is to ask the user which traces he wants to select using a selector visual element, but it is not ideal:
image

## Code example
```python
from taipy.gui import Gui
import pandas as pd
import numpy as np

mean_value = 0
selected_indices = []

data = pd.DataFrame(
    {
        "Time": np.arange(0, 10, 0.1),
        "A": np.arange(0, 10, 0.1) + np.random.randn(100) + 10,
        "B": np.arange(0, 10, 0.1) + np.random.randn(100),
    }
)

layout = {
    # Force the Box select tool
    "dragmode": "select",
}


def on_change(state, var, val):
    if var == "selected_indices":
        state.mean_value = np.mean([data["A"][idx] for idx in val]) if len(val) else 0


page = """
## Mean of <|{len(selected_indices)}|raw|> selected points: <|{mean_value}|format=%.2f|raw|>

<|{data}|chart|selected={selected_indices}|x=Time|y[1]=A|y[2]=B|layout={layout}|>
"""

Gui(page).run()
```

FredLL-Avaiga:
if there are several traces and only one bound variable for selected property, the value is an array of the different selections

resolves #1061

```
import numpy as np
import pandas as pd

from taipy.gui import Gui

mean_values = []
selected_indices = []

data = pd.DataFrame(
    {
        "Time": np.arange(0, 10, 0.1),
        "A": np.arange(0, 10, 0.1) + np.random.randn(100) + 10,
        "B": np.arange(0, 10, 0.1) + np.random.randn(100),
    }
)

layout = {
    # Force the Box select tool
    "dragmode": "select",
}


def on_change(state, var: str, val):
    if var.startswith("selected_indices"):
        if val and isinstance(val[0], list):
            means = []
            for i, sel in enumerate(val):
                arr = [data["A" if i == 0 else "B"][idx] for idx in sel]
                means.append(
                    f"Trace {i} Mean of {len(arr)} selected points:"
                    + "{:10.4f}".format(np.mean(arr) if len(sel) else 0)
                )
            state.mean_values = means
        else:
            state.mean_values = [
                f"Trace 0 ({var}): Mean of {len(val)} selected points:"
                + "{:10.4f}".format(np.mean([data["A"][idx] for idx in val]) if len(val) else 0)
            ]


page = """
## Select points to see the mean
<|selector|lov={mean_values}|not active|>

<|{data}|chart|selected={selected_indices}|x=Time|y[1]=A|y[2]=B|layout={layout}|>
"""

Gui(page).run()
```

### DIFF
--- a/frontend/taipy-gui/src/components/Taipy/Chart.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Chart.tsx
@@ -236,6 +236,14 @@ const TaipyPlotlyButtons: ModeBarButtonAny[] = [
     },
 ];
 
+const updateArrays = (sel: number[][], val: number[], idx: number) => {
+    if (idx >= sel.length || val.length !== sel[idx].length || val.some((v, i) => sel[idx][i] != v)) {
+        sel = sel.concat();
+        sel[idx] = val;
+    }
+    return sel;
+};
+
 const Chart = (props: ChartProp) => {
     const {
         title = "",
@@ -286,13 +294,12 @@ const Chart = (props: ChartProp) => {
                         if (!Array.isArray(val)) {
                             val = [];
                         }
-                        if (
-                            idx >= sel.length ||
-                            val.length !== sel[idx].length ||
-                            val.some((v, i) => sel[idx][i] != v)
-                        ) {
-                            sel = sel.concat();
-                            sel[idx] = val;
+                        if (idx === 0 && val.length && Array.isArray(val[0])) {
+                            for (let i = 0; i < val.length; i++) {
+                                sel = updateArrays(sel, val[i] as unknown as number[], i);
+                            }
+                        } else {
+                            sel = updateArrays(sel, val, idx);
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -236,6 +236,14 @@ const TaipyPlotlyButtons: ModeBarButtonAny[] = [
     },
 ];
 
+const updateArrays = (sel: number[][], val: number[], idx: number) => {
+    if (idx >= sel.length || val.length !== sel[idx].length || val.some((v, i) => sel[idx][i] != v)) {
+        sel = sel.concat();
```
FabienLelaquais: Please add // Shallow copy
That concat() syntax is really misleading.

FabienLelaquais (dismissed):
💪👍💪

frontend/taipy-gui/src/components/Taipy/Chart.tsx
FredLL-Avaiga: I could write sel = [...sel] if you'd rather ?

dinhlongviolin1 (dismissed):
LGTM

FabienLelaquais (approved):
👍

frontend/taipy-gui/src/components/Taipy/Chart.tsx
FabienLelaquais: With the comment (and why don't you keep the capital 'S', I don't know), everything is clear! Thanks.

PR #1360, opened 2024-06-05
When no cycle are used in the scenario configs ...

### LINKED ISSUE #1343: Scenario Selector - Default filters should be consistent with config
### Description

The goal of this issue is to improve the UX of Scenario Selector filters. Some filters are for example not useful when not using cycles:

- primary
- cycle label
- cycle start
- cycle end

These filters should not be shown as they were not used in this case.

### Acceptance Criteria

### Code of Conduct

FredLL-Avaiga:
do not show Cycle* and primary as column for filtering 
resolves #1343

### DIFF
--- a/taipy/gui_core/_adapters.py
+++ b/taipy/gui_core/_adapters.py
@@ -257,28 +257,21 @@ def _get_datanode_property(attr: str):
 
 
 class _GuiCoreScenarioProperties(_TaipyBase):
-    _SC_TYPES: t.Dict[str, t.Dict[str, t.Union[str, bool]]] = {
-        "Config id": {"type": "string"},
-        "Label": {"type": "string"},
-        "Creation date": {"type": "date"},
-        "Cycle label": {"type": "string", "cycle": True},
-        "Cycle start": {"type": "date", "cycle": True},
-        "Cycle end": {"type": "date", "cycle": True},
-        "Primary": {"type": "boolean", "cycle": True},
-        "Tags": {"type": "string"},
+    _SC_PROPS: t.Dict[str, t.Dict[str, t.Union[str, bool]]] = {
+        "Config id": {"attr": "config_id", "type": "string"},
+        "Label": {"attr": "name", "type": "string"},
+        "Creation date": {"attr": "creation_date", "type": "date"},
+        "Cycle label": {"attr": "cycle.name", "type": "string", "for_cycle": True},
+        "Cycle start": {"attr": "cycle.start", "type": "date", "for_cycle": True},
+        "Cycle end": {"attr": "cycle.end", "type": "date", "for_cycle": True},
+        "Primary": {"attr": "is_primary", "type": "boolean", "for_cycle": True},
+        "Tags": {"attr": "tags", "type": "string"},
     }
-    __SC_LABELS: t.Dict[str, str] = {
-        "Config id": "config_id",
-        "Creation date": "creation_date",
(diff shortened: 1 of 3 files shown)


taipy/gui_core/_adapters.py
```diff
@@ -257,17 +257,17 @@ def _get_datanode_property(attr: str):
...
-    _SC_TYPES = {
-        "Config id": "string",
-        "Label": "string",
-        "Creation date": "date",
-        "Cycle label": "string",
-        "Cycle start": "date",
-        "Cycle end": "date",
-        "Primary": "boolean",
-        "Tags": "string",
+    _SC_TYPES: t.Dict[str, t.Dict[str, t.Union[str, bool]]] = {
```
FabienLelaquais: Now that I think again about this, I think it makes far more sense to have a single dict:
```
_SC_PROPS: t.Dict[str, t.Dict[str, t.Union[str, bool]]] = {
    "Config id": {"attr": "config_id", "type": "string"},
    "Label": {"attr": "name", "type": "string"},
    "Creation date": {"attr": "creation_date", "type": "date"},
    ...
    "Cycle label": {"attr": "cycle.name", "type": "string", "for_cycle": True},
    ...
}
```

No need for _TYPES and _LABELS really, and less error prone.
Note that I renamed "cycle" to "for_cycle". Less ambiguous.

taipy/gui_core/_adapters.py
```diff
@@ -289,7 +289,7 @@ def get_hash():
     def get_type(attr: str):
         if prop := _get_datanode_property(attr):
             return _GuiCoreScenarioProperties.__DN_TYPES.get(prop, "any")
-        return _GuiCoreScenarioProperties._SC_TYPES.get(attr, "any")
+        return _GuiCoreScenarioProperties._SC_TYPES.get(attr, {"type": "any"}).get("type", "any")
```
FabienLelaquais: That reads really strange: if 'attr' is unknown, you carry on anyway?
With my renaming I would propose `_SC_PROPS.get(attr).get("type", "any")`

taipy/gui_core/_adapters.py
```diff
@@ -342,6 +342,9 @@ def get(self):
 
 class _GuiCoreScenarioFilter(_GuiCoreScenarioProperties):
     DEFAULT = list(_GuiCoreScenarioProperties._SC_TYPES.keys())
+    DEFAULT_NO_CYCLE = [
+        p[0] for p in filter(lambda prop: not prop[1].get("cycle", False), _GuiCoreScenarioProperties._SC_TYPES.items())
```
FabienLelaquais:
```suggestion
        p[0] for p in filter(lambda prop: not prop[1].get("for_cycle", False), _GuiCoreScenarioProperties._SC_TYPES.items())
```

FabienLelaquais (dismissed):
Let me know if you feel that my recommendation makes sense...

taipy/gui_core/_adapters.py
FredLL-Avaiga: Support for custom properties!

dinhlongviolin1 (approved):
LGTM

FabienLelaquais (approved):
👍

taipy/gui_core/_adapters.py
FabienLelaquais: Exxxxxxxxxxcellent point

PR #1390, opened 2024-06-11
Feature/enterprise#362 - Add can_create() api

trgiangdo:
Resolves https://github.com/Avaiga/taipy-enterprise/issues/362

This PR adds `taipy.can_create()` API to check if a DataNodeConfig or a ScenarioConfig can be created via `taipy.create_scenario()` or `taipy.create_global_data_node()` and return a Reasons object

### DIFF
--- a/taipy/config/_serializer/_base_serializer.py
+++ b/taipy/config/_serializer/_base_serializer.py
@@ -81,6 +81,8 @@ class _BaseSerializer(object):
             return [cls._stringify(val) for val in as_dict]
         if isinstance(as_dict, tuple):
             return [cls._stringify(val) for val in as_dict]
+        if isinstance(as_dict, set):
+            return [cls._stringify(val) for val in as_dict]
         return as_dict
 
     @staticmethod
--- a/taipy/core/data/_data_manager.py
+++ b/taipy/core/data/_data_manager.py
@@ -24,6 +24,8 @@ from ..config.data_node_config import DataNodeConfig
 from ..cycle.cycle_id import CycleId
 from ..exceptions.exceptions import InvalidDataNodeType
 from ..notification import Event, EventEntityType, EventOperation, Notifier, _make_event
+from ..reason._reason_factory import _build_config_can_not_create_reason, _build_not_global_datanode_config_reason
+from ..reason.reason import Reasons
 from ..scenario.scenario_id import ScenarioId
 from ..sequence.sequence_id import SequenceId
 from ._data_fs_repository import _DataFSRepository
@@ -67,6 +69,18 @@ class _DataManager(_Manager[DataNode], _VersionMixin):
             dn_config: data_nodes.get((dn_config, owner_id)) or cls._create_and_set(dn_config, owner_id, None)
             for dn_config, owner_id in dn_configs_and_owner_id
         }
+
+    @classmethod
+    def _can_create(cls, config: DataNodeConfig) -> Reasons:
+        config_id = getattr(config, "id", None) or str(config)
+        reason = Reasons(config_id)
+
+        if not isinstance(config, DataNodeConfig):
+            reason._add_reason(config_id, _build_config_can_not_create_reason(config_id))
+        elif config.scope is not Scope.GLOBAL:
+            reason._add_reason(config_id, _build_not_global_datanode_config_reason(config_id))
+
+        return reason
 
     @classmethod
     def _create_and_set(
--- a/taipy/core/data/_data_manager.py
+++ b/taipy/core/data/_data_manager.py
@@ -71,14 +71,15 @@ class _DataManager(_Manager[DataNode], _VersionMixin):
         }
 
     @classmethod
-    def _can_create(cls, config: DataNodeConfig) -> Reasons:
+    def _can_create(cls, config: Optional[DataNodeConfig] = None) -> Reasons:
         config_id = getattr(config, "id", None) or str(config)
         reason = Reasons(config_id)
 
-        if not isinstance(config, DataNodeConfig):
-            reason._add_reason(config_id, _build_wrong_config_type_reason(config_id))
-        elif config.scope is not Scope.GLOBAL:
-            reason._add_reason(config_id, _build_not_global_scope_reason(config_id))
+        if config is not None:
+            if not isinstance(config, DataNodeConfig):
+                reason._add_reason(config_id, _build_wrong_config_type_reason(config_id, "DataNodeConfig"))
+            elif config.scope is not Scope.GLOBAL:
+                reason._add_reason(config_id, _build_not_global_scope_reason(config_id))
 
         return reason
 
--- a/taipy/core/data/_data_manager.py
+++ b/taipy/core/data/_data_manager.py
@@ -24,7 +24,7 @@ from ..config.data_node_config import DataNodeConfig
 from ..cycle.cycle_id import CycleId
 from ..exceptions.exceptions import InvalidDataNodeType
 from ..notification import Event, EventEntityType, EventOperation, Notifier, _make_event
-from ..reason._reason_factory import _build_config_can_not_create_reason, _build_not_global_datanode_config_reason
+from ..reason._reason_factory import _build_not_global_scope_reason, _build_wrong_config_type_reason
 from ..reason.reason import Reasons
 from ..scenario.scenario_id import ScenarioId
 from ..sequence.sequence_id import SequenceId
@@ -76,9 +76,9 @@ class _DataManager(_Manager[DataNode], _VersionMixin):
         reason = Reasons(config_id)
 
         if not isinstance(config, DataNodeConfig):
-            reason._add_reason(config_id, _build_config_can_not_create_reason(config_id))
+            reason._add_reason(config_id, _build_wrong_config_type_reason(config_id))
         elif config.scope is not Scope.GLOBAL:
-            reason._add_reason(config_id, _build_not_global_datanode_config_reason(config_id))
+            reason._add_reason(config_id, _build_not_global_scope_reason(config_id))
 
         return reason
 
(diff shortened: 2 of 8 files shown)


taipy/config/_serializer/_base_serializer.py
```diff
@@ -81,6 +81,8 @@ def _stringify(cls, as_dict):
             return [cls._stringify(val) for val in as_dict]
         if isinstance(as_dict, tuple):
             return [cls._stringify(val) for val in as_dict]
+        if isinstance(as_dict, set):
+            return [cls._stringify(val) for val in as_dict]
```
jrobinAV, line 85: Ouch. We missed that?
Well spotted!

taipy/core/reason/_reason_factory.py
```diff
@@ -22,3 +22,11 @@ def _build_data_node_is_not_written(dn_id: DataNodeId) -> str:
 
 def _build_not_submittable_entity_reason(entity_id: str) -> str:
     return f"Entity {entity_id} is not a submittable entity"
+
+
+def _build_config_can_not_create_reason(config_id: str) -> str:
```
jrobinAV:
```suggestion
def _build_wrong_config_type_reason(config_id: str) -> str:
```

taipy/core/reason/_reason_factory.py
```diff
@@ -22,3 +22,11 @@ def _build_data_node_is_not_written(dn_id: DataNodeId) -> str:
...
+    return f'Object "{config_id}" is not a valid config to be created'
```
jrobinAV, line 33:
```suggestion
    return f'Object "{config_id}" is not a valid config.'
```

taipy/core/reason/_reason_factory.py
```diff
@@ -22,3 +22,11 @@ def _build_data_node_is_not_written(dn_id: DataNodeId) -> str:
...
+
+
+def _build_not_global_datanode_config_reason(config_id: str) -> str:
```
jrobinAV:
```suggestion
def _build_not_global_scope_reason(config_id: str) -> str:
```

taipy/core/taipy.py
```diff
@@ -873,6 +873,23 @@ def get_cycles() -> List[Cycle]:
     return _CycleManagerFactory._build_manager()._get_all()
 
 
+def can_create(config: Union[ScenarioConfig, DataNodeConfig]) -> Reasons:
```
jrobinAV: Is it possible to make the config optional?
If no config is passed, we just return True (an empty Reason). The false part will be implemented with the role management in Enterprise Edition.

So the API can be used for a specific config, or globally.

What do you think?

tests/core/data/test_data_manager.py
```diff
@@ -45,6 +45,24 @@ def test_create_data_node_and_modify_properties_does_not_modify_config(self):
         assert dn.properties.get("foo") == "bar"
         assert dn.properties.get("baz") == "qux"
 
+    def test_can_create(self):
+        dn_config = Config.configure_in_memory_data_node("dn", 10, scope=Scope.SCENARIO)
```
jrobinAV:
```suggestion
        dn_config = Config.configure_data_node("dn", 10, scope=Scope.SCENARIO)
```

tests/core/data/test_data_manager.py
```diff
@@ -45,6 +45,24 @@ def test_create_data_node_and_modify_properties_does_not_modify_config(self):
...
+        global_dn_config = Config.configure_data_node(
+            id="global_dn", storage_type="in_memory", scope=Scope.GLOBAL, data=10
+        )
+
+        reasons = _DataManager._can_create(global_dn_config)
+        assert bool(reasons) is True
```
jrobinAV, line 59: Do we really need to cast the Reasons object? What would the following do?
```
assert reasons is True
```
Does it work as well?

taipy/core/data/_data_manager.py
```diff
@@ -68,6 +70,18 @@ def _bulk_get_or_create(
             for dn_config, owner_id in dn_configs_and_owner_id
         }
 
+    @classmethod
+    def _can_create(cls, config: DataNodeConfig) -> Reasons:
+        config_id = getattr(config, "id", None) or str(config)
```
jrobinAV, line 75: I don't understand the way of getting the config id from a config.

taipy/core/taipy.py
```diff
@@ -873,6 +873,23 @@ def get_cycles() -> List[Cycle]:
...
+    Returns:
+        True if the given config can be created. False otherwise.
+    """
+    if isinstance(config, ScenarioConfig):
+        return _ScenarioManagerFactory._build_manager()._can_create(config)
+    elif isinstance(config, DataNodeConfig):
+        return _DataManagerFactory._build_manager()._can_create(config)
+    else:
+        config_id = getattr(config, "id", None) or str(config)
+        return Reasons(config_id)._add_reason(config_id, _build_config_can_not_create_reason(config_id))
```
jrobinAV: If it is not a `ScenarioConfig` or a `DataNodeConfig`, the reason to return is that `config` has an incorrect type. There is no need to look for a `config_id`. No?

```suggestion
        return Reasons(str(config))._add_reason(str(config), _build_config_can_not_create_reason(config_id))
```

What do you think?

tests/core/data/test_data_manager.py
trgiangdo: Unfortunately, no. It's a Reasons object

taipy/core/reason/_reason_factory.py
trgiangdo: This is to vague, no?

taipy/core/data/_data_manager.py
trgiangdo: If the object is not a proper Config with id, we will need to convert it to string to print it out

taipy/core/taipy.py
trgiangdo: If it's a Section object, `str(config)` will return its id?

We need the id here right?

taipy/core/taipy.py
trgiangdo: I'm not sure to understand this. So the user can check if he can create anything at all when using enterprise?

tests/core/data/test_data_manager.py
jrobinAV: and what about?
```
assert reasons==True
```

taipy/core/data/_data_manager.py
jrobinAV: If config is not a scenario config or a data node config, it is just a wrong object. I guess we can just say the object has an incorrect type. Do we need to see if it has an id or not? 
A message like `f"Type '{config.type}' of object '{str(config)}' is not correct." should be enough.

What do you think? In any case it is a minor remark.

trgiangdo: I think we can keep it this way to be consistent with the is_submittable api

tests/core/data/test_data_manager.py
trgiangdo: unfortunately, it won't work

taipy/core/data/_data_manager.py
jrobinAV: All right. 👍🏻

taipy/core/reason/_reason_factory.py
jrobinAV: Yes. This message will be displayed in the UI to the end user.
What about:
```f"'{config_id}' must be a valid {entity_type} configuration."```
 That means we need to pass the entity_type as a parameter.
 It will be instantiated as: 
```
"'toto' must be a valid Scenario configuration."
```
or
```
"'toto' must be a valid DataNode configuration."
```

What do you think?

taipy/core/taipy.py
jrobinAV: Yes. 
The purpose is deactivating the 'add scenario' button in the scenario selector.
At this stage, there is no configuration to pass since clicking on the button opens a dialog in which you can select your configuration and create a new scenario from it.
If I don't have the right to create a scenario (whatever the configuration), it is useless to let the user click on the 'add scenario' button.

To sum up, `can_create()` should return true if and only if I have the right to create a scenario or a data node.

taipy/core/taipy.py
jrobinAV: My point is that if the `config` variable is not a `ScenarioConfig `or a `DataNodeConfig`, it means the user created a bug somehow. It can be anything, like an integer, a list, or a more complex object like a Dataframe, or whatever. It looks strange to me to assume that it is maybe a `Section`, and so I don't get the point of trying to get an `id` from it. We just want a way to have a text representation of the object to let the end-user know it has an incorrect type.

But as I said on another comment, this is a minor remark. I am fine with your proposal.

tests/core/data/test_data_manager.py
jrobinAV: I thought the Reasons object would behave like a boolean in every situation. 
That is not so important. The most important is to be able to do: 
`if reasons:`

jrobinAV (dismissed):

taipy/core/reason/_reason_factory.py
trgiangdo: let me try something. This is pretty hard since we have a generic `taipy.can_create()` API, which accept both DataNodeConfig and ScenarioConfig

tests/core/data/test_data_manager.py
trgiangdo: yes we can do that, since the `if reasons:` is the same as `if bool(reasons):`

trgiangdo:
@FredLL-Avaiga @jrobinAV The `taipy.can_create()` now accepts the `config` as an optional parameter.

So with enterprise, `taipy.can_create()` without any parameter will check if the user has sufficient role to create entities

jrobinAV (approved):
The implementation is tricky, but I like it.

PR #1393, opened 2024-06-11
add submission indicator to scenario

FredLL-Avaiga:
refresh dag without on_submission

### DIFF
--- a/frontend/taipy/src/ChipStatus.tsx
+++ b/frontend/taipy/src/ChipStatus.tsx
@@ -1,0 +1,34 @@
+import React from "react";
+import { SxProps, Theme } from "@mui/material";
+import Chip from "@mui/material/Chip";
+
+export enum Status {
+    SUBMITTED = 1,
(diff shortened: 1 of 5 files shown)


frontend/taipy/src/ChipStatus.tsx
```diff
@@ -0,0 +1,34 @@
...
+    PENDING = 3,
+    RUNNING = 4,
+    CANCELED = 5,
+    FAILED = 6,
+    COMPLETED = 7,
+    SKIPPED = 8,
+    ABANDONED = 9,
+}
+
+const ChipStatus = ({ status, sx }: { status: number; sx?: SxProps<Theme> }) => {
```
FabienLelaquais: I would seriously consider renaming that to `StatusChip`: a chip that represents a status.
Don't you think?

FabienLelaquais (dismissed):
Really really nice!

FabienLelaquais (approved):
👌

Issue #1314, opened 2024-05-24
Chat control should support images


AlexandreSajus:
### Description

The chat control of #1092 is really cool! It would be a good addition to support images in the messages of the chat. This would be especially useful since now LLMs start supporting image vision and image generation. I could create cool demos around this chat control.

Florian:
People are also using chat demos to visualize tables inside their chat. This could also be a nice addition.

### Acceptance Criteria

- [ ] Ensure new code is unit tested, and check code coverage is at least 90%.
- [ ] Create related issue in taipy-doc for documentation and Release Notes.
- [ ] Check if a new demo could be provided based on this, or if legacy demos could be benefit from it.
- [ ] Ensure any change is well documented.

### Code of Conduct

- [X] I have checked the [existing issues](https://github.com/Avaiga/taipy/issues?q=is%3Aissue+).
- [ ] I am willing to work on this issue (optional)

ioannastantzou:
Hello there I think I could work on this issue!

FlorianJacta:
Great, I will assign it to you!

ioannastantzou:
hello again! can you specify where the chat code exists?

ioannastantzou:
> hello again! can you specify where the chat code exists?

is it taipy\frontend\taipy-gui\src\components\Taipy\chat.tsx ? @FabienLelaquais @FredLL-Avaiga

yaten2302:
Hey @FlorianJacta @AlexandreSajus , could you please assign this issue to me, if no one is working on this?

FlorianJacta:
@ioannastantzou is normally working on it. Could you confirm @ioannastantzou ?

yaten2302:
> @ioannastantzou is normally working on it. Could you confirm @ioannastantzou ?

It's ok if someone is working on this, but if not I'm in 👍

yaten2302:
Hey @FlorianJacta @AlexandreSajus , can this issue be assigned to me?

Also, I would like to ask that in here, do we've to add a files icon on the left side of the message box (similar to what is present in chatgpt)?<br>
![image](https://github.com/user-attachments/assets/a0fde0cf-65d9-496d-957d-dc7ecd906fcd)

<br>

![image](https://github.com/user-attachments/assets/8a86ec0c-e348-4a13-9001-1efcab3a4846)

FlorianJacta:
What do you mean by a file icon?

Can this be assigned? @FredLL-Avaiga @FabienLelaquais

[...]

PR #1399, opened 2024-06-12
Create chatwithimagesending

ioannastantzou:
onImageUpload() :handles the upload of image file. <input type='"file">:selection of image file for upload. handleImageUpload:is called when the user selects an image file. onChange:gets triggered when the user selects the image file.

### DIFF
--- a/chatwithimagesending
+++ b/chatwithimagesending
@@ -1,0 +1,383 @@
+import React, { useMemo, useCallback, KeyboardEvent, MouseEvent, useState, useRef, useEffect, ReactNode } from "react";
+import { SxProps, Theme, darken, lighten } from "@mui/material/styles";
+import Avatar from "@mui/material/Avatar";
+import Box from "@mui/material/Box";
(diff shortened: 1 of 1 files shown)


FlorianJacta:
Is this about https://github.com/Avaiga/taipy/issues/1314 ?

What is the goal of the code, and how does it interact with the chat control?

FredLL-Avaiga (changes requested):
This doesn't allow display of images in the chat
and does not handle images in the back-end.
the property onImageUpload cannot be generated by the back-end.

FabienLelaquais:
So this PR looks like you started by copying frontend/taipy-gui/src/components/Taipy/Chat.tsx then added the onImageUpload property to the Chat component. This would be invoked when the end user requests the upload of an image.

Now we don't quite understand how this articulates with the 'chat' control on the backend, which is what the application really sees, and certainly... how we would integrate that.

Is this PR like showing a concept that we might want to get into the product somehow?

PR #1405, opened 2024-06-13
feature/#1355 POC add polars to taipy

toan-quach:


jrobinAV (commented):
Is this a prototype? If yes, Can we turn it to a draft PR so we don't merge it by mistake?

Globally, it makes me confident we can add such feature.

trgiangdo:
This PR should be removed right?

toan-quach:
@trgiangdo yep 👍

PR #1425, opened 2024-06-18
call can_create to get the reason

FredLL-Avaiga:
resolves taipy-enterprise#362

### DIFF
--- a/taipy/gui_core/_context.py
+++ b/taipy/gui_core/_context.py
@@ -1107,6 +1107,4 @@ class _GuiCoreContext(CoreEventConsumerBase):
             _warn(f"dag.on_action(): Invalid function '{args[1]}()'.")
 
     def get_creation_reason(self):
-        if not (reason := can_create()):
-            return f"Cannot create scenario: {reason.reasons}"
-        return ""
+        return f"Cannot create scenario: {reason.reasons}" if not (reason := can_create()) else ""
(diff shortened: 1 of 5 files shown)


trgiangdo (dismissed):
LGTM

namnguyen20999 (dismissed):

namnguyen20999 (dismissed):
lgtm

taipy/gui_core/_context.py
```diff
@@ -1104,3 +1105,8 @@ def on_dag_select(self, state: State, id: str, payload: t.Dict[str, str]):
                     _warn(f"dag.on_action(): Exception raised in '{args[1]}()' with '{args[0]}'", e)
         elif args[1]:
             _warn(f"dag.on_action(): Invalid function '{args[1]}()'.")
+
+    def get_creation_reason(self):
+        if not (reason := can_create()):
```
FabienLelaquais: def get_creation_reason(self):
```suggestion
        f"Cannot create scenario: {reason.reasons}" if (reason := can_create()) else ""
```
?

FabienLelaquais (dismissed):
As long as `get_creation_reason()` is invoked at the right time... looks just fine!

taipy/gui_core/_context.py
FredLL-Avaiga: it would be f"Cannot ..." if not (reason :! ...)
I suppose it's more pythonic ...

FabienLelaquais (approved):
Even better!

FabienLelaquais (approved):
🫤

PR #1431, opened 2024-06-19
feature/#1344 refactored to integrate polars

toan-quach:
Avaiga/taipy-enterprise#411


taipy/core/data/csv.py
```diff
@@ -180,15 +180,19 @@ def _append(self, data: Any):
...
+
+        if columns and isinstance(data, pd.DataFrame):
+            data.columns = pd.Index(columns, dtype="object")
+
+        data.to_csv(
+            self._path,
+            index=False,
+            encoding=properties[self.__ENCODING_KEY],
+            header=properties[self._HAS_HEADER_PROPERTY]
             )
```
trgiangdo: This seems to be miss-aligned, right?
```suggestion
        )
```

taipy/core/data/parquet.py
```diff
@@ -157,7 +157,10 @@ def __init__(
         with _Reloader():
             self._write_default_data(default_value)
 
-        if not self._last_edit_date and (isfile(self._path) or isdir(self._path)):
+        if (
+            not self._last_edit_date  # type: ignore
+            and (isfile(self._path) or isdir(self._path[:-1] if self._path.endswith("*") else self._path))
```
trgiangdo, line 162: I don't get this change. Which case do we need the `if self._path.endswith("*") else self._path`?

tests/core/data/test_filter_sql_table_data_node.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, foo=None, bar=None, *args, **kwargs):
 
 
 class TestFilterSQLTableDataNode:
-    __pandas_properties = [
+    __sql_properties = [
```
trgiangdo: Oh, are we using the `__pandas_properties` or the `__sql_properties` in this file at all?
Is this a redundant variable?

trgiangdo (dismissed):
The improvements look good to me. I only have some minor comments

taipy/core/data/parquet.py
toan-quach: this is for polars actually, because to read a folder in polars, you need `folder_path/*` instead of just `folder_path/`

tests/core/data/test_filter_sql_table_data_node.py
toan-quach: ah good point, don't think we do! let me remove them

taipy/core/data/parquet.py
trgiangdo: Ah, I see. Thank you

trgiangdo (dismissed):

trgiangdo (dismissed):
LGTM

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -77,8 +77,9 @@ def path(self) -> Any:
     @path.setter
     def path(self, value):
         self._path = value
-        self.properties[self._PATH_KEY] = value
-        self.properties[self._IS_GENERATED_KEY] = False
+        properties = self.properties
+        properties[self._PATH_KEY] = value
+        properties[self._IS_GENERATED_KEY] = False
```
jrobinAV: I don't see the purpose for this change.

Data
jrobinAV, line 1: What is this file?

taipy/core/data/generic.py
```diff
@@ -108,7 +108,7 @@ def __init__(
             editor_expiration_date,
             **properties,
         )
-        if not self._last_edit_date:
+        if not self._last_edit_date:  # type: ignore
```
jrobinAV, line 111: WHy do we need to add an ignore here? Can't we avoid it?

tests/core/data/test_filter_sql_table_data_node.py
```diff
@@ -30,60 +29,6 @@ def __init__(self, foo=None, bar=None, *args, **kwargs):
 
 
 class TestFilterSQLTableDataNode:
-    __pandas_properties = [
```
jrobinAV, line 33: I don't get this. 
What are these properties about? Who use them?

tests/core/data/test_generic_data_node.py
```diff
@@ -53,9 +55,11 @@ class TestGenericDataNode:
     data = list(range(10))
 
     def test_create(self):
```
jrobinAV: This test function is too long. Can we take the opportunity to split it into 4 test functions?

jrobinAV (commented):
minor remarks

tests/core/data/test_filter_sql_table_data_node.py
trgiangdo: This is probably a copy-paste error. We have similar variable in other test files, but it's not needed here

jrobinAV: OK thx

taipy/core/data/_file_datanode_mixin.py
toan-quach: to reduce the amount of time we need to call `self.properties` as each time it's called, we are pulling from the repo again

toan-quach: ah I see what you meant! let me remove this change!

taipy/core/data/generic.py
toan-quach: mypy couldn't detect the type of `self._last_edit_date` so it throws an error :/

tests/core/data/test_generic_data_node.py
jrobinAV, line 1: 👍🏻

jrobinAV (dismissed):

trgiangdo (dismissed):

Data
jrobinAV, line 1: Can you delete this file? It should not be committed, right?

jrobinAV (commented):
1. Data should not be committed.

2. `excel.py `and `abstract_sql.py` have a low coverage. Can you check why and eventually increase the coverage?

Data
toan-quach: for excel.py -> this is due to we're skipping tests on append function as we're using pandas 1.3.5
image

toan-quach: For abstract_sql, it's scattered through a small number of lines not executed by tests with the highest concentration in this function 
image

jrobinAV (dismissed):

trgiangdo (approved):
I like the new tests on SQL connection and queries. Neat

jrobinAV (approved):
Great job

PR #1435, opened 2024-06-20
Update README.md

SohaibBaig1243:


SohaibBaig1243:
I have updated this file

FlorianJacta:
What is the goal of this PR? What issue do you want to address?

florian-vuillemot (changes requested):
IMHO this does not improve the readability of the file.

jrobinAV:
I agree with @florian-vuillemot . The file readability is degraded while the result does not change.

With any more motivation (or description) for this PR, I reject it.

PR #1437, opened 2024-06-21
chore: add extra step to mark taipy version as latest

### LINKED ISSUE #1266: Always mark Taipy main release as latest
**Description**
When releasing a subpackage from the Taipu repository, github will show the subpackage release as latest and this will also be reflected on Taipy Doc, which is not ideal. 

It is necessary to add a step on our release actions to mark the latest Taipy release with the latest tag to override the current behavior.

joaoandre-avaiga:
resolves #1266

### DIFF
--- a/.github/workflows/build-and-release-single-package.yml
+++ b/.github/workflows/build-and-release-single-package.yml
@@ -27,8 +27,7 @@ jobs:
         gui_VERSION: ${{ steps.version-setup.outputs.gui_VERSION }}
         rest_VERSION: ${{ steps.version-setup.outputs.rest_VERSION }}
         templates_VERSION: ${{ steps.version-setup.outputs.templates_VERSION }}
-        VERSION: ${{ steps.version-setup.outputs.VERSION }}
-        NEW_VERSION: ${{ steps.version-setup.outputs.NEW_VERSION }}
+        taipy_VERSION: ${{ steps.version-setup.outputs.taipy_VERSION }}
     steps:
       - uses: actions/checkout@v4
       - name: Extract branch name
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-release-single-package.yml
```diff
@@ -165,3 +165,10 @@ jobs:
         shell: bash
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Ensure Taipy release is marked as latest
+        run: |
+           gh release edit ${{needs.fetch-versions.outputs.config_VERSION}} --latest
```
trgiangdo: I believe we can use the `config_VERSION` or any other `{subpackage}_VERSION` right?

Anyway, the `needs.fetch-versions.outputs.VERSION` and the `needs.fetch-versions.outputs.NEW_VERSION` are redundant on line 30-31 right?

jrobinAV (dismissed):

jrobinAV:
Please add a description with the addressed issue #.

Thx.

.github/workflows/build-and-release-single-package.yml
joaoandre-avaiga: >I believe we can use the config_VERSION or any other {subpackage}_VERSION right?

It should be always Taipy Version, because thats the one we want to make latest. I made a typo.

>Anyway, the needs.fetch-versions.outputs.VERSION and the needs.fetch-versions.outputs.NEW_VERSION are redundant on line 30-31 right?
Right. Good catch.

.github/workflows/build-and-release-single-package.yml
```diff
@@ -165,3 +163,10 @@ jobs:
         shell: bash
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Ensure Taipy release is marked as latest
+        run: |
+           gh release edit ${{needs.fetch-versions.outputs.taipy_VERSION}} --latest
```
trgiangdo, line 170: I don't understand, where is the `taipy_VERSION` defined?

joaoandre-avaiga: Oh, it comes from the fech version step. A output variable was missing.

trgiangdo (approved):

jrobinAV (approved):

PR #1443, opened 2024-06-24
root page decouple

dinhlongviolin1:
Related to Avaiga/taipy-designer#128, Avaiga/taipy-designer#144

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -1111,10 +1111,13 @@ class Gui:
 
     def __handle_ws_get_module_context(self, payload: t.Any):
         if isinstance(payload, dict):
+            page_path = str(payload.get("path"))
+            if page_path in {"/", ""}:
+                page_path = Gui.__root_page_name
             # Get Module Context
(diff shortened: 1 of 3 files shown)


taipy/gui/gui.py
```diff
@@ -1111,10 +1111,13 @@ def __request_var_update(self, payload: t.Any):
 
     def __handle_ws_get_module_context(self, payload: t.Any):
         if isinstance(payload, dict):
+            page_path = str(payload.get("path"))
+            if page_path in {"/", ""}:
```
FabienLelaquais, line 1115:
```suggestion
            if page_path in ["/", ""]:
```
More efficient don't you think?

FabienLelaquais (approved):
🤔

taipy/gui/gui.py
dinhlongviolin1: set and list as iterable is similar i think. Set is also safer in most cases. we should move to set even in other cases where index is not important

FredLL-Avaiga (approved):

PR #1446, opened 2024-06-25
fix decouple metadata

dinhlongviolin1:
Resolve Avaiga/taipy-designer#129

### DIFF
--- a/frontend/taipy-gui/base/src/wsAdapter.ts
+++ b/frontend/taipy-gui/base/src/wsAdapter.ts
@@ -45,7 +45,11 @@ export class TaipyWsAdapter extends WsAdapter {
             } else if (message.type === "GMC") {
                 const payload = message.payload as Record<string, unknown>;
                 taipyApp.context = payload.context as string;
-                taipyApp.metadata = JSON.parse((payload.metadata as string) || "{}");
+                try {
+                    taipyApp.metadata = JSON.parse((payload.metadata as string) || "{}");
+                } catch (e) {
+                    console.error("Error parsing metadata from Taipy Designer", e);
+                }
             } else if (message.type === "GDT") {
                 const payload = message.payload as Record<string, ModuleData>;
                 const variableData = payload.variable;
--- a/frontend/taipy-gui/base/src/wsAdapter.ts
+++ b/frontend/taipy-gui/base/src/wsAdapter.ts
@@ -43,9 +43,9 @@ export class TaipyWsAdapter extends WsAdapter {
                 taipyApp.clientId = id;
                 taipyApp.updateContext(taipyApp.path);
             } else if (message.type === "GMC") {
-                const mc = (message.payload as Record<string, unknown>).data as string;
-                window.localStorage.setItem("ModuleContext", mc);
-                taipyApp.context = mc;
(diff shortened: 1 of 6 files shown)


frontend/taipy-gui/base/src/wsAdapter.ts
```diff
@@ -43,9 +43,9 @@ export class TaipyWsAdapter extends WsAdapter {
                 taipyApp.clientId = id;
                 taipyApp.updateContext(taipyApp.path);
             } else if (message.type === "GMC") {
-                const mc = (message.payload as Record<string, unknown>).data as string;
-                window.localStorage.setItem("ModuleContext", mc);
-                taipyApp.context = mc;
+                const payload = message.payload as Record<string, unknown>;
+                taipyApp.context = payload.context as string;
+                taipyApp.metadata = JSON.parse((payload.metadata as string) || "{}");
```
FredLL-Avaiga: is a surrounding try/catch worth it ?

FredLL-Avaiga (dismissed):

taipy/gui/gui.py
```diff
@@ -1116,10 +1116,15 @@ def __handle_ws_get_module_context(self, payload: t.Any):
...
+                # get metadata if there is one
+                metadata: t.Dict[str, t.Any] = {}
+                page = self._get_page(str(payload.get("path")))
+                if hasattr(page._renderer, "_metadata"):
+                    metadata = getattr(page._renderer, "_metadata", {})
                 self.__send_ws(
                     {
                         "type": _WsType.GET_MODULE_CONTEXT.value,
-                        "payload": {"data": mc},
+                        "payload": {"context": mc, "metadata": json.dumps(metadata, cls=_TaipyJsonEncoder)},
```
FredLL-Avaiga: "metadata": json.dumps(metadata, cls=_TaipyJsonEncoder)} if metadata else None

frontend/taipy-gui/base/src/wsAdapter.ts
dinhlongviolin1: Yes you are right!

taipy/gui/gui.py
dinhlongviolin1: metadata wouldnt be None since I set the default as an empty array

dinhlongviolin1: both the declaration and the default value of the getattr

FredLL-Avaiga: an empty dict would be falsy in python :-)

dinhlongviolin1: oh it is fine I want an empty dict anyway to ensure correct json parsing on the frontend side :D

FredLL-Avaiga:
That was my point: to avoid parsing if not necessary ;-)

taipy/gui/gui.py
dinhlongviolin1: Alright I will go with you :D

FredLL-Avaiga:
Don't feel obliged
No big deal

FabienLelaquais (approved):
How come this works, I don't have a clue.
Well done, very certainly.

FredLL-Avaiga (approved):

PR #1448, opened 2024-06-25
Updating the Readme

RymMichaut:
- Adding the new Taipy Designer banner
- Removing the old logo from the reame_img folder


RymMichaut:
image

this is how it would look like

jrobinAV (commented):
I have several remarks here.

1. The most important to me is that the README needs to talk to the right people. People coming to the GitHub page are looking for the source code. So they are contributors, or eventually strong developers that want to better understand how it works under the hood. They are not interested in other tools, ads, marketing speech, or communication channels. Plus, we have to understand that readme is specific to one branch and so specific to one version of Taipy.

2. I would remove the mentions of Taipy Designer and Taipy Studio. They are not part of the Taipy package or the Taipy repository. It is out of scope and confusing to me. If you want to communicate better, you can propose changes to the Taipy Studio repository README.
Eventually, I would add a one-line section just before contributing called Taipy Ecosystem to add links to other taipy tools like taipy Studio, taipy Designer, taipy Enterprise, Taipy Doc, and Discord.

3. I would remove the feature section. It does not provide useful information for the targeted reader.

4. If you want to keep mentioning Taipy Designer, the banner should be compatible with light mode:
image

5. If you want to keep the Taipy designer mention, the Request a demo button should be a real clickable button or should be removed.

6. The first screen takes too much space for no real information. As a developer, when I go to a repo. I want to get straight to the point and not have to scroll one screen to start reading what I want. I would reduce the Title, the subtitle, and the logo to the strict minimum. I would move 

I don't want to block the PR if I am the only one to share this opinion on the role of the README. So I don't block the PR, and let the responsibility to others to approve or reject it. (coward decision 🤣)

PR #1459, opened 2024-06-28
Reason class refactoring

### LINKED ISSUE #1471: Refactor Reason API

toan-quach:


jrobinAV (commented):
That's exactly what I had in mind. I love it.
As you can imagine, the documentation of the `Reasons` class, and every single `Reason`, will be greatly facilitated.

If you can provide some doc strings that would be great. Thx!

taipy/core/reason/reason.py
```diff
@@ -9,33 +9,114 @@
...
+    """
+    A reason holds the reason why specific actions cannot be performed.
 
-    def _add_reason(self, entity_id: str, reason: str) -> "Reasons":
-        if entity_id not in self._reasons:
-            self._reasons[entity_id] = set()
-        self._reasons[entity_id].add(reason)
-        return self
+    Attributes:
+        reason (str): The reason why the action cannot be performed.
```
jrobinAV:
```suggestion
        reason (str): The English representation of the reason why the action cannot be performed.
```

taipy/core/reason/reason.py
```diff
@@ -9,33 +9,114 @@
...
+        entity_id (str): The identifier of the `Entity^`.
+    """
+
+    def __init__(self, entity_id: str):
+        Reason.__init__(self, f"Entity {entity_id} is not a submittable entity")
+
+
+class WrongConfigType(Reason):
+    """
+    WrongConfigType holds the reason the provided config is not the valid config expected.
```
jrobinAV:
```suggestion
    A config id is not a valid expected config, which prevents specific actions from being performed.
```

taipy/core/reason/reason.py
```diff
@@ -9,33 +9,114 @@
...
+    def datanode(self):
+        from ..data._data_manager_factory import _DataManagerFactory
+
+        return _DataManagerFactory._build_manager()._get(self.datanode_id)
+
+
+class DataNodeEditInProgress(Reason, _DataNodeReasonMixin):
+    """
+    DataNodeEditInProgress holds the reason a `DataNode^` is being edited,
+    which prevents specific actions from being performed.
```
jrobinAV:
```suggestion
    A `DataNode^` is being edited, which prevents specific actions from being performed.
```

taipy/core/reason/reason.py
```diff
@@ -9,33 +9,114 @@
...
+
+    def __init__(self, datanode_id: str):
+        Reason.__init__(self, f"DataNode {datanode_id} is being edited")
+        _DataNodeReasonMixin.__init__(self, datanode_id)
+
+
+class DataNodeIsNotWritten(Reason, _DataNodeReasonMixin):
+    """
+    DataNodeIsNotWritten holds the reason a `DataNode^` is not yet written,
+    which prevents specific actions from being performed.
```
jrobinAV:
```suggestion
    A `DataNode^` has never been written, which prevents specific actions from being performed.
```

taipy/core/reason/reason.py
```diff
@@ -9,33 +9,114 @@
...
+    """
+
+    def __init__(self, datanode_id: str):
+        Reason.__init__(self, f"DataNode {datanode_id} is not written")
+        _DataNodeReasonMixin.__init__(self, datanode_id)
+
+
+class EntityIsNotSubmittableEntity(Reason):
+    """
+    EntityIsNotSubmittableEntity holds the reason an entity is not submittable.
```
jrobinAV:
```suggestion
    An entity is not a submittable entity, which prevents specific actions from being performed.
```

taipy/core/reason/reason.py
```diff
@@ -9,33 +9,114 @@
...
+        else:
+            reason = f'Object "{config_id}" is not a valid config to be created'
+
+        Reason.__init__(self, reason)
+
+
+class NotGlobalScope(Reason):
+    """
+    NotGlobalScope holds the reason the data node config does not have a GLOBAL scope,
+    which prevents certain actions from being performed.
```
jrobinAV:
```suggestion
    A data node config does not have a GLOBAL scope, which prevents specific actions from being performed.
```

taipy/core/reason/reason.py
```diff
@@ -9,33 +9,114 @@
...
+from typing import Any, Optional
 
 
-class Reasons:
-    def __init__(self, entity_id: str) -> None:
-        self.entity_id: str = entity_id
-        self._reasons: Dict[str, Set[str]] = {}
+class Reason:
+    """
+    A reason holds the reason why specific actions cannot be performed.
```
jrobinAV:
```suggestion
    A reason explains why a specific action cannot be performed.
    
    This is a parent class aiming at being implemented by specific sub-classes.
    
    Because Taipy applications are natively multiuser, asynchronous, and dynamic, 
    some functions might not be called in some specific contexts. You can protect 
    such calls by calling other methods that return `Reasons^` objects. A `Reasons^` 
    object acts like a boolean: True if the action can be performed and False otherwise. 
    If the action cannot be performed, the `Reasons^` object holds all the reasons why 
    the action cannot be performed as a list of `Reason` objects.
```

taipy/core/reason/reasons.py
```diff
@@ -0,0 +1,51 @@
...
+from typing import Dict, Set
+
+from .reason import Reason
+
+
+class Reasons:
+    """
+    This class is used to store all the reasons to explain why some Taipy operations are not allowed.
+
+    TODO - Add more details about the class.
```
jrobinAV:
```suggestion       
    Because Taipy applications are natively multiuser, asynchronous, and dynamic, 
    some functions might not be called in some specific contexts. You can protect 
    such calls by calling other methods that return a Reasons object. It acts like a 
    boolean: True if the operation can be performed and False otherwise. 
    If the action cannot be performed, the Reasons object holds all the `reasons as a list
    of `Reason` objects. Each `Reason` holds an explanation of why the operation cannot be
    performed.
```

taipy/core/reason/reasons.py
```diff
@@ -0,0 +1,51 @@
...
+        return self
+
+    def _entity_id_exists_in_reason(self, entity_id: str) -> bool:
+        return entity_id in self._reasons
+
+    def __bool__(self) -> bool:
+        return len(self._reasons) == 0
+
+    @property
+    def reasons(self) -> str:
```
jrobinAV: Missing docstring

taipy/core/reason/reason.py
```diff
@@ -9,33 +9,114 @@
...
-                del self._reasons[entity_id]
-        return self
+    def __init__(self, reason: str):
+        self._reason = reason
 
-    def _entity_id_exists_in_reason(self, entity_id: str) -> bool:
-        return entity_id in self._reasons
+    @property
+    def reason(self) -> str:
+        return self._reason
```
jrobinAV: Can't we remove this property? We already have the string representation.

jrobinAV (commented):
I proposed a few minor changes.

I don't have better names, but I believe we should change the wording.

`Reasons` and `Reason` are to close to each other. Any suggestion?

taipy/core/reason/reasons.py
```diff
@@ -0,0 +1,56 @@
...
+        """Retrieves a collections of reasons as a string that explains why the action cannot be performed.
```
jrobinAV:
```suggestion
        """Retrieves a collection of reasons as a string that explains why the action cannot be performed.
```

taipy/core/reason/reasons.py
jrobinAV: Because Taipy applications are natively multiuser, asynchronous, and dynamic, some functions
might not be called in some specific contexts. You can protect such calls by calling another 
method that tells you if the first method can be called. This second method returns a Reasons 
object that collects the explanations of why the operation cannot be performed. It acts like a 
boolean: True if the operation can be performed and False otherwise. If the action cannot be
performed, the Reasons object holds all the explanations' list. Each list element is a 
`Reason` and holds one explanation of why the operation cannot be performed.

taipy/core/reason/reasons.py
```diff
@@ -0,0 +1,62 @@
...
+    Because Taipy applications are natively multiuser, asynchronous, and dynamic,
+    some functions might not be called in some specific contexts. You can protect
+    such calls by calling other methods that return a Reasons object. It acts like a
+    boolean: True if the operation can be performed and False otherwise.
+    If the action cannot be performed, the Reasons object holds all the `reasons as a list
+    of `Reason` objects. Each `Reason` holds an explanation of why the operation cannot be
+    performed.
+    """
+
+    def __init__(self, entity_id: str) -> None:
```
jrobinAV: After rethinking the topic, I don't remember the need of an entity_id here.

toan-quach: probably to know which entity that reason belongs to?

taipy/core/data/_data_manager.py
```diff
@@ -75,9 +74,9 @@ def _can_create(cls, config: Optional[DataNodeConfig] = None) -> Reasons:
 
         if config is not None:
             if not isinstance(config, DataNodeConfig):
-                reason._add_reason(config_id, _build_wrong_config_type_reason(config_id, "DataNodeConfig"))
+                reason._add_reason(config_id, WrongConfigType(config_id, DataNodeConfig.__name__))
```
trgiangdo: Since the name difference between `Reasons` and `Reason` classes is only singular or plural, I would love it if the variables to match the class name to avoid future confusion.

```suggestion
                reasons._add_reason(config_id, WrongConfigType(config_id, DataNodeConfig.__name__))
```

taipy/core/_entity/submittable.py
```diff
@@ -94,9 +93,9 @@ def is_ready_to_run(self) -> Reasons:
 
         for node in self.get_inputs():
             if node._edit_in_progress:
-                reason._add_reason(node.id, _build_data_node_is_being_edited_reason(node.id))
+                reason._add_reason(node.id, DataNodeEditInProgress(node.id))
```
trgiangdo: Since the name difference between Reasons and Reason classes is only singular or plural, I would love it if the variables to match the class name to avoid future confusion.

```suggestion
                reasons._add_reason(node.id, DataNodeEditInProgress(node.id))
```

trgiangdo (dismissed):
LGTM

taipy/core/reason/reasons.py
trgiangdo: An action like `submit` may have multiple failures, on different entities, so I think the `entity_id` here is "reason"able

trgiangdo (dismissed):

taipy/core/reason/reasons.py
jrobinAV: A ReasonCollection object is not necessarily attached to an entity but to a boolean function call. Some individual Reason objects can be attached to an entity, though.

For example, `can_create(Scenario_cfg)` is not attached to any entity.
We can also imagine adding reasons to any function/method unrelated to an entity, like `is_standalone_mode_on()`, `is_core_running()`, etc. 

Because this ReasonCollection object is not persisted, keeping the entity_id is unnecessary, IMHO. Plus, I checked, and it is not read anywhere else but in the tests.

Am I missing something? What do you think?

jrobinAV (commented):
Did you have the chance to read the brainstorm we had on R&D Team channel about the naming?

If I recall, the outcome was to have a `ReasonCollection` object holding a collection of `Reason` objects. What do you think?

toan-quach:
@jrobinAV yep I agree but in context of behaving as a boolean, ReasonCollection doesn't sound right to me :/ 
like if I call `if ReasonCollection`, it just sounds kinda weird imo

trgiangdo:
I don't think we will ever call `if ReasonCollection` directly, because the `ReasonCollection` will be the return type of many boolean API. This naming is fine for me.

jrobinAV:
I agree with Giang. The type name ReasonColection won't necessarily appear in the user code. Just a variable name.

```
if reasons:= scenario.is_submittable():
    do_something_with(reasons.reasons)
```

It sounds fair to me.

taipy/core/reason/reasons.py
toan-quach: no you're right, let me remove that then

taipy/core/reason/reasons.py
jrobinAV, line 1: Please rename the module name as well.

taipy/core/reason/reasons.py
```diff
@@ -0,0 +1,61 @@
...
+from .reason import Reason
+
+
+class ReasonCollection:
+    """
+    This class is used to store all the reasons to explain why some Taipy operations are not allowed.
+
+    Because Taipy applications are natively multiuser, asynchronous, and dynamic,
+    some functions might not be called in some specific contexts. You can protect
+    such calls by calling other methods that return a Reasons object. It acts like a
```
jrobinAV:
```suggestion
    such calls by calling other methods that return a ReasonCollection. It acts like a
```

taipy/core/reason/reasons.py
```diff
@@ -0,0 +1,61 @@
...
+    boolean: True if the operation can be performed and False otherwise.
+    If the action cannot be performed, the Reasons object holds all the `reasons as a list
+    of `Reason` objects. Each `Reason` holds an explanation of why the operation cannot be
+    performed.
```
jrobinAV:
```suggestion
    If the action cannot be performed, the ReasonCollection holds all the individual reasons
    as a list of `Reason` objects. Each `Reason` explains why the operation cannot be
    performed.
```

taipy/core/reason/__init__.py
```diff
@@ -9,4 +9,13 @@
...
 
-from .reason import Reasons
+from .reason import (
+    DataNodeEditInProgress,
+    DataNodeIsNotWritten,
+    EntityIsNotSubmittableEntity,
+    NotGlobalScope,
+    Reason,
+    WrongConfigType,
+    _DataNodeReasonMixin,
```
jrobinAV: Why do we need to expose this `_DataNodeReasonMixin` class?

jrobinAV (dismissed):
I love the result. 
Last minor remarks...

taipy/core/reason/__init__.py
toan-quach: ah my bad didn't notice that!

trgiangdo (approved):

jrobinAV (approved):

PR #1474, opened 2024-07-01
Feature/#1196 - Add download and upload api for file-based datanodes

### LINKED ISSUE #1196: Data node Viewer - Add file download/upload capability for file based data nodes
**Description**
When a data node is stored as a file, it would be handy for the end user to add a way to download or upload the file from the data node viewer easily.

This should help the end user understand its data, particularly when data is corrupted or an unexpected behavior happens.

This capability could be enabled/disabled by the developer through a property. By default, it should be enabled.

**Proposal**
1. The download could be accessible through a "Download" button (eventually, an icon when the element width is too small).
2. The upload could be accessible through an "Upload" file selector (eventually, an icon when the element width is too small). When a file is passed to the Upload mechanism, Taipy should:
    - Check the file (a check property should let the developer provide a custom function).
    - Write the file at the data node path location. Eventually, overwrite it if a file already exists.
    - track the data node edition

**Usage**
```python
def custom_check(state, file_path) -> bool:
    # Apply the check logic
    return True


<| {selected_datanode} | datanode | file_download=false | file_upload=true | upload_check=custom_check |>
```

- [ ] Create a ticket for the Gui part (buttons into the datanode viewer)

trgiangdo:
Resolves #1196

In this PR, 2 new APIs are open for file-based DataNode:
- `_get_downloadable_path()` that returns the path to the downloadable data file
- `_upload()` that check the uploaded file and replace the current data file


taipy/core/data/_file_datanode_mixin.py
```diff
@@ -92,3 +92,50 @@ def _migrate_path(self, storage_type, old_path) -> str:
...
+        """
+        from ._data_manager_factory import _DataManagerFactory
+
+        upload_path = pathlib.Path(path)
+
+        if upload_checker is not None:
+            try:
+                upload_data = self._read_from_path(upload_path.name)
+            except Exception:
+                return False
```
trgiangdo: For now, we only return a bool if there is any Exception.

Should we instead:
- raise the Exception 
- log the error and return False

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -92,3 +92,50 @@ def _migrate_path(self, storage_type, old_path) -> str:
...
+
+            if not upload_checker(upload_path.name, upload_data):
+                return False
```
trgiangdo: For now, we only return a bool if the checker return False. Should we return a `Reasons`?

toan-quach: I think we should :D

taipy/core/data/csv.py
```diff
@@ -27,7 +27,7 @@
 from .data_node_id import DataNodeId, Edit
 
 
-class CSVDataNode(DataNode, _FileDataNodeMixin, _TabularDataNodeMixin):
+class CSVDataNode(_FileDataNodeMixin, DataNode, _TabularDataNodeMixin):
```
toan-quach: hmm? Don't we initiated them all manually? so the order shouldn't be an issue?

taipy/core/data/data_node.py
```diff
@@ -373,6 +373,28 @@ def _get_last_modified_datetime(cls, path: Optional[str] = None) -> Optional[dat
     def storage_type(cls) -> str:
         raise NotImplementedError
 
+    def _get_downloadable_path(self) -> str:
```
toan-quach: this has already defined in the FileDataNodeMixin no?

taipy/core/data/csv.py
trgiangdo: It does in this case.
So we have `_upload()` in both `DataNode` and `_FileDataNodeMixin`.
If we put DataNode first, when calling `CSVDataNode._upload()`, it will always go to `DataNode._upload()`, which is not what we want here.

toan-quach (dismissed):
lgtm but I have several questions

taipy/core/data/data_node.py
trgiangdo: Yes. The idea is that we want the method to only defined in `_ FileDataNodeMixin`.

`SQLDataNode. _get_downloadable_path()` should raise the NotImplementedError error.

toan-quach: ah ok, then I would suggest to move this code in to _abstract_sql.py instead, since DataNode and _FileDataNodeMixin are 2 different abstract class, I don't think they will have much connection in this sense.

trgiangdo: No I think I misrepresented my idea.

So, for all data nodes that are not File-based, calling to `_get_downloadable_path` should raise an error.
For file-based dn, it should work.

toan-quach: hmm in that case, I think it should still be fine if you put it in the _abstract_sql, since other DataNode are not extending from `_FileDataNodeMixin` so it should be fine for them no?

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -92,3 +92,50 @@ def _migrate_path(self, storage_type, old_path) -> str:
...
+        Returns:
+            The downloadable path of the file data of the data node if it exists, otherwise an empty string.
+        """
+        if os.path.exists(self.path) and isfile(self._path):
+            return self.path
+
+        return ""
+
+    def _upload(self, path: str, upload_checker: Optional[Callable[[str, Any], bool]] = None) -> bool:
+        """Upload a csv file data to the data node.
```
jrobinAV:
```suggestion
        """Upload a file data to the data node.
```

taipy/core/data/_file_datanode_mixin.py
jrobinAV: Yes, probably log the error and raise a custom exception.

taipy/core/data/_file_datanode_mixin.py
jrobinAV: I would say yes.

taipy/core/data/csv.py
jrobinAV: WHy don't you want to extend DataNode first?

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -92,3 +92,50 @@ def _migrate_path(self, storage_type, old_path) -> str:
         if os.path.exists(old_path):
             shutil.move(old_path, new_path)
         return new_path
+
+    def _get_downloadable_path(self) -> str:
```
jrobinAV, line 100: What is it used for?

taipy/core/data/data_node.py
trgiangdo: what about other classes like Mongo, S3, Generic, ...?
If we do that, we also need to add one for each of them, which is weird.

taipy/core/data/data_node.py
```diff
@@ -373,6 +373,28 @@ def _get_last_modified_datetime(cls, path: Optional[str] = None) -> Optional[dat
...
+        Parameters:
+            path (str): The path of the file to upload to the data node.
+            upload_checker (Optional[Callable[[str, Any], bool]]): A function to check if the upload is allowed.
+                The function takes the title of the upload data and the data itself as arguments and returns
+                True if the upload is allowed, otherwise False.
+
+        Returns:
+            True if the upload was successful, otherwise False.
+        """
+        raise NotImplementedError
```
jrobinAV: We don't want these to be part of a DataNode. They are only specific to the file data nodes, no?
No need to define the methods here. and so in each data node class (csv, excel, json, etc.) you can inherit from data node first.

taipy/core/data/json.py
```diff
@@ -25,7 +25,7 @@
 from .data_node_id import DataNodeId, Edit
 
 
-class JSONDataNode(DataNode, _FileDataNodeMixin):
+class JSONDataNode(_FileDataNodeMixin, DataNode):
```
jrobinAV: Same here, I would prefer to inherit from data node first.

taipy/core/data/json.py
```diff
@@ -150,7 +150,13 @@ def decoder(self, decoder: json.JSONDecoder):
         self.properties[self._DECODER_KEY] = decoder
 
     def _read(self):
-        with open(self._path, "r", encoding=self.properties[self.__ENCODING_KEY]) as f:
+        return self._read_from_path()
+
+    def _read_from_path(self, path: Optional[str] = None, **read_kwargs) -> Any:
+        if not path:
+            path = self._path
+
```
jrobinAV, line 158:
```suggestion
```

taipy/core/data/parquet.py
```diff
@@ -28,7 +28,7 @@
 from .data_node_id import DataNodeId, Edit
 
 
-class ParquetDataNode(DataNode, _FileDataNodeMixin, _TabularDataNodeMixin):
+class ParquetDataNode(_FileDataNodeMixin, DataNode, _TabularDataNodeMixin):
```
jrobinAV: Same here abt the order. I prefer to inherit from the data node first.

taipy/core/data/pickle.py
```diff
@@ -22,7 +22,7 @@
 from .data_node_id import DataNodeId, Edit
 
 
-class PickleDataNode(DataNode, _FileDataNodeMixin):
+class PickleDataNode(_FileDataNodeMixin, DataNode):
```
jrobinAV: same

taipy/core/data/data_node.py
toan-quach: hmmm I don't think they are extending the class `_FileDataNodeMixin` :/

taipy/core/data/_file_datanode_mixin.py
trgiangdo: For getting the download path so that the GUI can provide a download interface

jrobinAV (dismissed):

jrobinAV (commented):
Instead of sleeping, can we use frozen time instead?

trgiangdo:
Oh, great idea. Let me do that

taipy/core/reason/reason.py
```diff
@@ -123,3 +123,33 @@ class NotGlobalScope(Reason):
 
     def __init__(self, config_id: str):
         Reason.__init__(self, f'Data node config "{config_id}" does not have GLOBAL scope')
+
+
+class UploadFileCanNotBeRead(Reason):
```
jrobinAV:
```suggestion
class UploadFileCanNotBeRead(Reason, _DataNodeReasonMixin):
```

taipy/core/reason/reason.py
```diff
@@ -123,3 +123,33 @@ class NotGlobalScope(Reason):
...
+
+    def __init__(self, file_name: str, datanode_id: str):
+        Reason.__init__(
+            self,
+            f"The uploaded file {file_name} can not be read, "
+            f'therefore is not a valid data file for data node "{datanode_id}"',
+        )
+
+
+class InvalidUploadFile(Reason):
```
jrobinAV:
```suggestion
class InvalidUploadFile(Reason, _DataNodeReasonMixin):
```

taipy/core/reason/reason.py
```diff
@@ -123,3 +123,33 @@ class NotGlobalScope(Reason):
...
+        datanode_id (str): The datanode id that the file is intended to upload to.
+    """
+
+    def __init__(self, file_name: str, datanode_id: str):
+        Reason.__init__(
+            self,
+            f"The uploaded file {file_name} can not be read, "
+            f'therefore is not a valid data file for data node "{datanode_id}"',
+        )
+
```
jrobinAV:
```suggestion
        Reason.__init__(
            self,
            f"The uploaded file {file_name} can not be read, "
            f'therefore is not a valid data file for data node "{datanode_id}"',
        )
        _DataNodeReasonMixin.__init__(self, datanode_id)

```

taipy/core/reason/reason.py
```diff
@@ -123,3 +123,33 @@ class NotGlobalScope(Reason):
...
+    """
+    The uploaded file has invalid data, therefore is not a valid data file for the data node.
+
+    Attributes:
+        file_name (str): The name of the file that was uploaded.
+        datanode_id (str): The datanode id that the file is intended to upload to.
+    """
+
+    def __init__(self, file_name: str, datanode_id: str):
+        Reason.__init__(self, f'The uploaded file {file_name} has invalid data for data node "{datanode_id}"')
```
jrobinAV, line 156:
```suggestion
        Reason.__init__(self, f'The uploaded file {file_name} has invalid data for data node "{datanode_id}"')
        _DataNodeReasonMixin.__init__(self, datanode_id)
        
```

jrobinAV (dismissed):

toan-quach (dismissed):
looks good to me, might need to rebase though

jrobinAV (dismissed):

toan-quach (approved):

PR #1480, opened 2024-07-05
Feature/#393 - Creation time filtering for scenario getters

### LINKED ISSUE #393: Time filtering for scenario getters
- Propose a way to filter scenarios by time period (start time - end time).

trgiangdo:
Resolves #393 

The user can filter scenarios by creation date when calling `taipy.get_scenarios()`, for example:
```python
from datetime import datetime
import taipy

jan_scenarios = taipy.get_scenarios(
  created_start_time=datetime(2024, 1, 1),
  created_end_time=datetime(2024, 1, 31)
)
```

Let me know what you think.

### DIFF
--- a/taipy/core/scenario/_scenario_manager.py
+++ b/taipy/core/scenario/_scenario_manager.py
@@ -288,9 +288,8 @@ class _ScenarioManager(_Manager[Scenario], _VersionMixin):
     def _get_primary_scenarios(cls) -> List[Scenario]:
         return [scenario for scenario in cls._get_all() if scenario.is_primary]
 
-    @classmethod
+    @staticmethod
     def _sort_scenarios(
-        cls,
         scenarios: List[Scenario],
         descending: bool = False,
         sort_key: Literal["name", "id", "config_id", "creation_date", "tags"] = "name",
@@ -306,9 +305,8 @@ class _ScenarioManager(_Manager[Scenario], _VersionMixin):
             scenarios.sort(key=lambda x: (x.name, x.id), reverse=descending)
         return scenarios
 
-    @classmethod
+    @staticmethod
     def _filter_by_creation_time(
-        cls,
         scenarios: List[Scenario],
         created_start_time: Optional[datetime] = None,
         created_end_time: Optional[datetime] = None,
--- a/taipy/core/scenario/_scenario_manager.py
+++ b/taipy/core/scenario/_scenario_manager.py
@@ -9,7 +9,7 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
-import datetime
+from datetime import datetime
 from functools import partial
 from typing import Any, Callable, Dict, List, Literal, Optional, Union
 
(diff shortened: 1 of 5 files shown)


taipy/core/scenario/_scenario_manager.py
```diff
@@ -468,3 +468,25 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
...
+    def _get_by_creation_time(
+        cls, start_time: datetime, end_time: datetime, version_number: Optional[str] = None
+    ) -> List[Scenario]:
+        """
+        Get all scenarios by a given creation time period.
+        The time period is inclusive.
+
+        Parameters:
+            start_time (datetime): Start time of the period.
+            end_time (datetime): End time of the period.
```
jrobinAV: I would call them `from` and `to`.

jrobinAV (commented):
We can add two optional parameters to the `get_scenarios` API: `from` and `to`. That's easier to use than having a separate function.

taipy/core/scenario/_scenario_manager.py
trgiangdo: `from` is a Python keyword though, so we can't use it

jrobinAV: Sure. Let's keep your proposal, then.

taipy/core/scenario/_scenario_manager.py
```diff
@@ -306,6 +306,35 @@ def _sort_scenarios(
             scenarios.sort(key=lambda x: (x.name, x.id), reverse=descending)
         return scenarios
 
+    @classmethod
```
jrobinAV: This can be static. cls is never used. Right?

taipy/core/taipy.py
```diff
@@ -548,6 +552,8 @@ def get_scenarios(
     else:
         scenarios = []
 
+    if created_start_time or created_end_time:
+        scenarios = scenario_manager._filter_by_creation_time(scenarios, created_start_time, created_end_time)
     if is_sorted:
         scenario_manager._sort_scenarios(scenarios, descending, sort_key)
     return scenarios
```
jrobinAV, line 559: Does it make sense to completely delegate the logic of this method `get_scenarios` to the Scenario_manager?

taipy/core/taipy.py
```diff
@@ -589,6 +599,9 @@ def get_primary_scenarios(
     """
     scenario_manager = _ScenarioManagerFactory._build_manager()
     scenarios = scenario_manager._get_primary_scenarios()
+
+    if created_start_time or created_end_time:
+        scenarios = scenario_manager._filter_by_creation_time(scenarios, created_start_time, created_end_time)
     if is_sorted:
         scenario_manager._sort_scenarios(scenarios, descending, sort_key)
```
jrobinAV, line 606: Same here. This logic could be delegated to the ScenarioManager.

taipy/core/taipy.py
trgiangdo: I don't think so. We already have manager methods to handle different options, no complicated logic is handled here

taipy/core/taipy.py
trgiangdo: I don't think so. We already have manager methods to handle different options, no complicated logic is handled here

taipy/core/taipy.py
jrobinAV: All right.

jrobinAV (approved):
neat

PR #1485, opened 2024-07-08
Feature/#848 adding min max picker for date selector

namnguyen20999:


doc/gui/examples/controls/date-min-max-range.py
```diff
@@ -0,0 +1,28 @@
...
+import datetime
+
+from taipy.gui import Gui
+
+date = datetime.date(2024, 6, 15)
+min_date = datetime.date(2024, 5, 15)
+max_date = datetime.date(2024, 7, 15)
+
+page = """
+<|{date}|date|min={min_date}|max={max_date}|>
```
FabienLelaquais: Please rename the *min_date* property to *min* and the *max_date* property to *max*.

doc/gui/examples/controls/date-min-max-range.py
```diff
@@ -0,0 +1,28 @@
...
+# To execute this script, make sure that the taipy-gui package is installed in your
+# Python environment and run:
+#     python <script>
+# -----------------------------------------------------------------------------------------
+import datetime
+
+from taipy.gui import Gui
+
+date = datetime.date(2024, 6, 15)
+min_date = datetime.date(2024, 5, 15)
```
FabienLelaquais: Please rename *min_date* to *start* and *max_date* to *end*.

taipy/gui/types.py
```diff
@@ -82,6 +82,10 @@ class PropertyType(Enum):
     """
     The property holds a dictionary.
     """
+    dynamic_date = "dynamicdate"
+    """
+    The property holds a dynamic date.
```
FabienLelaquais: The proper wording should be: "The property is dynamic and holds a  date."
Can you please change the surrounding property types as well?

doc/gui/examples/controls/date-min-max-range.py
FabienLelaquais, line 1: Rename to date-min-max.py

FabienLelaquais (dismissed):
Minor changes... and a mismatch between the example's properties and the element's.

doc/gui/examples/controls/date-min-max-range.py
namnguyen20999: Min_date & Max_date; These are the variables, not the attribute properties. I would not name the variable to min & max because it might conflict with shadow's built-in name. :D

taipy/gui/types.py
namnguyen20999: I've changed based on your suggestion

doc/gui/examples/controls/date-min-max-range.py
FabienLelaquais: That's exactly what I said, and what you did:
properties = min & max (they appeared as min_date and max_date in the fragment I commented)
variables = start & end

FabienLelaquais (approved):
👌

PR #1501, opened 2024-07-11
Expose websocket decouple frontend

dinhlongviolin1:


frontend/taipy-gui/base/src/app.ts
```diff
@@ -101,6 +103,16 @@ export class TaipyApp {
...
     }
 
+    get onWsMessage() {
+        return this._onWsMessage;
+    }
+    set onWsMessage(handler: OnWsMessage | undefined) {
+        if (handler !== undefined && handler?.length !== 3) {
+            throw new Error("onWsMessage() requires three parameters");
+        }
+        this._onWsMessage = handler;
```
FredLL-Avaiga, line 136: I wonder if you should `bind` the function so that this will be defined and you don't have to pass taipyApp as first parameter ?

FabienLelaquais (dismissed):
👌

frontend/taipy-gui/base/src/app.ts
FredLL-Avaiga: You could also have a default implementation that does nothing and would allow you to not check the existence before each call...

FredLL-Avaiga (dismissed):

frontend/taipy-gui/base/src/app.ts
dinhlongviolin1: i think I can replace the get method with a default function that can also check for its existence. the method will also bypass the passing of the app

dinhlongviolin1:
added a comment execution function for all exposed callbacks

frontend/taipy-gui/base/src/app.ts
```diff
@@ -101,6 +104,16 @@ export class TaipyApp {
         this._onReload = handler;
     }
 
+    get onWsMessage() {
```
FredLL-Avaiga, line 129: I thought you would only defined the setter and let the on&lt;whatever> (ie onWsMessage here) be a real method with parameter that would check the existence of this._onWsMessage and call it

ie 
```
onWsMessage = (event, payload) => {
  this._onWsMessage && this._onWsMessage(event, payload);
}
```
I think we're loosing params checking with a single onUserEvent

FredLL-Avaiga (dismissed):

frontend/taipy-gui/base/src/app.ts
dinhlongviolin1: since onWsMessage namespace was occupied by the set, you could not actually have a normal function. It is reserved already by the set method and I tried to remove the get function and definied a normal function and ts screamed at me :D

FabienLelaquais (dismissed):
👍

frontend/taipy-gui/base/src/app.ts
```diff
@@ -80,6 +93,10 @@ export class TaipyApp {
         this._onChange = handler;
```
FredLL-Avaiga, line 93: this._onChange = handler.bind(this)

frontend/taipy-gui/base/src/app.ts
```diff
@@ -80,6 +93,10 @@ export class TaipyApp {
...
     }
 
+    onChangeEvent(encodedName: string, value: unknown) {
+        this.onChange && this.onChange(this, encodedName, value);
```
FredLL-Avaiga, line 97: Then this.on change && this.onChange(encodedName, value)

FredLL-Avaiga (approved):

FabienLelaquais (approved):
👌

frontend/taipy-gui/base/src/app.ts
dinhlongviolin1: I tried this one but it doesnt work unfortunately. After testing I dont think that they work with classes.
```javacript
class A {
    a = undefined;
    constructor() {
        this.a = ((a) => {
            console.log(a)
        }).bind(this);
    }
}
```

FredLL-Avaiga: It does
But I think there's a difference between function and arrow function. 
No big deal anyway

PR #1538, opened 2024-07-17
Implement delta_color options

### LINKED ISSUE #1520: Possibility to swap colors for metric viz elements
### Description

Sometimes, if the delta is negative, it is good, and the associated color should be green, not red.

In this example, we want to decrease the cost and carbon footprint so the metric should indicate that this solution is good. The color associated with that should be green and not red.

image

### Acceptance Criteria

### Code of Conduct

Satoshi-Sh:
## Related Issue 
fix #1520 

## Updates 
- Implemented positive_delta_color and negative_delta_color 
- Added tests for the new features 

## Test Screenshot 

delta_color_test

```
from taipy.gui import Gui

value = 50
negative = -20
positive = 20
# format & delta_format are used to format the value and delta value respectively.
# They use the printf syntax.

page = """
<|{value}|metric|delta={positive}|positive_delta_color=#FF4136|format=%d km/h|delta_format=%d km/h|>
<|{value}|metric|delta={negative}|negative_delta_color=#3D9970|format=%d km/h|delta_format=%d km/h|>
"""


Gui(page).run()
```


taipy/gui/viselements.json
```diff
@@ -1189,6 +1189,19 @@
                         "type": "dynamic(int|float)",
                         "doc": "The delta value to display."
                     },
+                    {
+                        "name": "positive_delta_color",
+                        "type": "str",
+                        "default_value": "#3D9970",
```
FredLL-Avaiga: that should be "\"#3D9970\"" but we want no default value here

taipy/gui/viselements.json
```diff
@@ -1189,6 +1189,19 @@
...
+                        "doc": "Set the color of positive delta."
+                    },
+                    {
+                        "name": "negative_delta_color",
+                        "type": "str",
+                        "default_value": "#FF4136",
```
FredLL-Avaiga: "\"#FF4136\""  but we want no default value here

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -58,7 +60,9 @@ const Metric = (props: MetricProps) => {
     const {
         width = "100%",
         height,
-        showValue = true
+        showValue = true,
+        positiveDeltaColor = "#3D9970",
```
FredLL-Avaiga: I'm not sure we want a default value here

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -103,7 +107,16 @@ const Metric = (props: MetricProps) => {
                     reference: typeof value === 'number' && typeof delta === 'number' ? value - delta : undefined,
                     prefix: extractPrefix(props.deltaFormat),
                     suffix: extractSuffix(props.deltaFormat),
-                    valueformat: sprintfToD3Converter(props.deltaFormat)
+                    valueformat: sprintfToD3Converter(props.deltaFormat),
+                    increasing:{
```
FredLL-Avaiga: define only if the value is not undefined

taipy/gui/_renderers/factory.py
```diff
@@ -353,6 +353,8 @@ class _Factory:
                 ("min", PropertyType.number, 0),
                 ("max", PropertyType.number, 100),
                 ("delta", PropertyType.dynamic_number),
+                ("positive_delta_color", PropertyType.string, "#3D9970"),
```
FredLL-Avaiga: no default value here

FredLL-Avaiga (changes requested):
I think we don't want default values for those properties ie default behavior if not defined

Satoshi-Sh:
Does this update follow your request? @FredLL-Avaiga

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -88,6 +92,14 @@ const Metric = (props: MetricProps) => {
         const mode = (props.type === "none") ? [] : ["gauge"];
         showValue && mode.push("number");
         (delta !== undefined) && mode.push("delta");
+        const deltaIncreasing = positiveDeltaColor?{
+            symbol: "▲",
```
FredLL-Avaiga: Do you have to specify the symbol ?

Satoshi-Sh: I got a Typescript error at first.

I just checked it and it didn't show up anymore, so I deleted the symbols.

FredLL-Avaiga (dismissed):
@FabienLelaquais ?

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -103,7 +113,10 @@ const Metric = (props: MetricProps) => {
                     reference: typeof value === 'number' && typeof delta === 'number' ? value - delta : undefined,
                     prefix: extractPrefix(props.deltaFormat),
                     suffix: extractSuffix(props.deltaFormat),
-                    valueformat: sprintfToD3Converter(props.deltaFormat)
+                    valueformat: sprintfToD3Converter(props.deltaFormat),
+                    increasing:deltaIncreasing,
```
FabienLelaquais:
```suggestion
                    increasing: deltaIncreasing,
```

taipy/gui/viselements.json
```diff
@@ -1189,6 +1189,17 @@
                         "type": "dynamic(int|float)",
                         "doc": "The delta value to display."
                     },
+                    {
+                        "name": "positive_delta_color",
+                        "type": "str",
+                        "doc": "Set the color of positive delta."
```
```suggestion
                        "doc": "The color used to display positive values of <i>delta</i>."
```

taipy/gui/viselements.json
```diff
@@ -1189,6 +1189,17 @@
...
+                    },
+                    {
+                        "name": "negative_delta_color",
+                        "type": "str",
+                        "doc": "Set the color of negative delta."
```
```suggestion
                        "doc": "The color used to display negativee values of <i>delta</i>."
```

FabienLelaquais (changes requested):
That's really neat, thanks!

But... because we did not discuss enough before you jumped on this, I would recommend we switch gears a little bit.

As @FredLL-Avaiga mentioned, giving the ability to easily switch the positive/negative colors for the delta value makes sense: the user has a single property to set, and we're done.

Here is my (new, unfortunately) proposal:
We would add two properties (this would be my doc proposal, by the way):
- delta_color (str, no default): The color that is used to display the value of the delta property.
      If negative_delta_color is set, then this property applies for positive values of delta only.
      If this property is set to "invert", then delta values are represented with the color used for negative values if delta is positive. The value for delta is also represented with the color used for positive values if delta is negative.
- negative_delta_color (str, no default): if set, this represents the color to be used when the value of delta is negative (or positive if delta_color is set to 'invert').

What do you @Satoshi-Sh and you @FredLL-Avaiga think about this mess?

@FlorianJacta you are more than welcome to give your feedback on this as well!

FredLL-Avaiga:
Let's do it

FlorianJacta:
The API seems a bit strange to me; I have a hard time understanding it.

*delta_color' can be a color or the string: 'invert' which is not a color.

When is *delta_color* different from the color for positive value?

FabienLelaquais:
> The API seems a bit strange to me; I have a hard time understanding it.
> 
> *delta_color' can be a color or the string: 'invert' which is not a color.
> 
> When is _delta_color_ different from the color for positive value?

Well if you simply want to see green for positive values and red for negative values, you do nothing. Like today.
If you want to see red for positive values and green for negative values, set _delta_color_ to "invert".
In other cases, you can set _delta_color_ to any color you want represent positive values, and _negative_delta_color_ to any color you want represent negative values.
Clearer?

The case where you do not set _delta_color_, but *do* set _negative_delta_color_ would impact only the color used to display negative values.

Satoshi-Sh:
I implemented "invert" as Fabien suggested. 

I have one issue in the `tests/gui/e2e/test_metric_indicator.py`.

<|-50|metric|delta_color="invert"|delta=20|>

The test doesn't pass if I don't put "" around invert. The color works fine without ""(ex, #3D9970). Do you know how to fix it? 

Update
I found out the incorrect logic in the Metric.tsx. Will fix it soon.

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -88,6 +92,11 @@ const Metric = (props: MetricProps) => {
         const mode = (props.type === "none") ? [] : ["gauge"];
         showValue && mode.push("number");
         (delta !== undefined) && mode.push("delta");
+        const deltaIncreasing = deltaColor ? {
+            color: deltaColor == "invert" ? "#FF4136" : deltaColor } : undefined
```
FredLL-Avaiga, line 96: Where did you find this color? 
Is there anyway we can get it dynamically from plotly?
Should the invert value be composted in lower case?

FredLL-Avaiga (changes requested):

frontend/taipy-gui/src/components/Taipy/Metric.tsx
Satoshi-Sh: I found the default color in their [documentation](https://plotly.com/javascript/reference/indicator/). 

I checked the library and it's hardcoded. Is there any way to get these values?
https://github.com/plotly/plotly.js/blob/b6411b58357ecdc1c1518dede25060bac7d6cd41/src/constants/delta.js#L4

I solved the test problem. I just needed to make the delta value negative instead of value.

taipy/gui/viselements.json
```diff
@@ -1189,6 +1189,17 @@
...
                     },
+                    {
+                        "name": "delta_color",
+                        "type": "str",
+                        "doc": "The color that is used to display the value of the <i>delta</i> property. If negative_delta_color is set, then this property applies for positive values of delta only. If this property is set to \"invert\", then delta values are represented with the color used for negative values if delta is positive. The value for delta is also represented with the color used for positive values if delta is negative."
+                    },
+                    {
+                        "name": "negative_delta_color",
+                        "type": "str",
+                        "doc": "if set, this represents the color to be used when the value of <i>delta</i> is negative (or positive if delta_color is set to \"invert\")"
```
```suggestion
                        "doc": "If set, this represents the color to be used when the value of <i>delta</i> is negative (or positive if <i>delta_color</i> is set to \"invert\")"
```

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -130,6 +142,8 @@ const Metric = (props: MetricProps) => {
         props.type,
         value,
         showValue,
+        deltaColor,
```
FabienLelaquais, line 145: Isn't there are way *not* to indicate these colors if the user doesn't change them after all?
That would allow us to not go and find the Plotly's default values, but also to follow their changes when they occur...

FabienLelaquais (changes requested):
That's really good!
I have a tiny doc comment, and an implementation question though.
Thanks!

frontend/taipy-gui/src/components/Taipy/Metric.tsx
Satoshi-Sh: If users don't specify the colors or "invert", the code takes them as undefined so the delta will be displayed in the default color.

FabienLelaquais: Yes yes I understand.
My question really was about why we need to inject any color value in this code.
Pos and neg default color values should be left to Plotly to decide. See my point? No big deal by the way.
Great job!

FabienLelaquais (approved):
👍👍👍

FredLL-Avaiga (approved):
I can think of no easy solution to get the colors used by plotly by default :-(

FredLL-Avaiga:
If you agree @Satoshi-Sh, I'll merge this PR

FabienLelaquais:
> I can think of no easy solution to get the colors used by plotly by default :-(

Fair enough.

@FredLL-Avaiga go ahead and merge if you also agree.

Satoshi-Sh:
Yes, you can go ahead.

PR #1555, opened 2024-07-19
feat: remove SQL repository

joaoandre-avaiga:


toan-quach (dismissed):
lgtm! a friendly note that some of the changes should be applied to enterprise repo as well :D

toan-quach:
I forgot but do check taipy-config as well to see if there's any removals to be done

jrobinAV:
The code coverage of `taipy/core/_repository/_base_taipy_model.py` is slightly low. Can you check if we can improve the coverage of this class?

jrobinAV (dismissed):

joaoandre-avaiga:
> I forgot but do check taipy-config as well to see if there's any removals to be done

I've checked config, theres no direct reference to sql repo, only the repository attributes(type, properties), we should be good. I'll check enterprise if theres anything to be done.

joaoandre-avaiga:
> The code coverage of `taipy/core/_repository/_base_taipy_model.py` is slightly low. Can you check if we can improve the coverage of this class?

I've added specif tests for the base_model class. There are 2 abstract methods still not covered, but they don't have any logic to be tested. So this should be enough.

trgiangdo (dismissed):
This is looking good. Thank you Joao

jrobinAV (dismissed):
As people were saying in the 50's: 
"Rock-and-roll"...

jrobinAV (approved):

trgiangdo (approved):

PR #1580, opened 2024-07-24
Status, Status Link Controls update

### LINKED ISSUE #1157: Feature Improvement: Status controls could have a way to have icons instead of the default letters
**Description**
For the Status GUI control, instead of showing E, S, I and W, it would be nice to have a way to put icons in it instead.

**Acceptance Criteria**
- [ ] Status control from TaiPy gui can have Icons instead of the default letters

pravintargaryen:
Resolves: #1157

### DIFF
--- a/frontend/taipy-gui/src/components/Taipy/Status.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Status.tsx
@@ -14,6 +14,10 @@
 import React, { MouseEvent, ReactNode, useMemo } from "react";
 import Chip from "@mui/material/Chip";
 import Avatar from "@mui/material/Avatar";
+import CheckCircleIcon from "@mui/icons-material/CheckCircle";
+import ErrorIcon from "@mui/icons-material/Error";
+import WarningIcon from "@mui/icons-material/Warning";
+import InfoIcon from "@mui/icons-material/Info";
 
 import { getInitials } from "../../utils";
 import { TaipyBaseProps } from "./utils";
@@ -29,6 +33,21 @@ interface StatusProps extends TaipyBaseProps {
     onClose?: (evt: MouseEvent) => void;
     icon?: ReactNode;
 }
+
+export const getStatusIcon = (status: string) => {
+    switch (status) {
+        case "S":
+            return <CheckCircleIcon data-testid="CheckCircleIcon" />;
+        case "E":
+            return <ErrorIcon data-testid="ErrorIcon" />;
+        case "W":
+            return <WarningIcon data-testid="WarningIcon" />;
+        case "I":
+            return <InfoIcon data-testid="InfoIcon" />;
+        default:
+            return "❓";
+    }
+};
 
 const status2Color = (status: string): "error" | "info" | "success" | "warning" => {
     status = (status || "").toLowerCase();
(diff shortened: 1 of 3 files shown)


frontend/taipy-gui/src/components/Taipy/StatusList.spec.tsx
FredLL-Avaiga, line 1: No need to commit a file if it is only reformated

frontend/taipy-gui/src/components/Taipy/Status.tsx
FredLL-Avaiga, line 1: Should we allow the choice between letter and icon, for example add a property `with_icon` boolean True by default.
What do you think @FlorianJacta @FabienLelaquais ?

FredLL-Avaiga (changes requested):
please don't commit reformated files

frontend/taipy-gui/src/components/Taipy/Status.tsx
FlorianJacta: I guess, we should

frontend/taipy-gui/src/components/Taipy/StatusList.spec.tsx
pravintargaryen: please advise on the next steps

frontend/taipy-gui/src/components/Taipy/Status.tsx
pravintargaryen: please advise on the next steps

FredLL-Avaiga: What kind of advice are you looking for?

FredLL-Avaiga: - add a property to taipy\gui\_renderers\factory.py L 494 : `("with_icon", PropertyType.boolean, False),`
- add the corresponding property to frontend\taipy-gui\src\components\Taipy\StatusList.tsx L 88: `withIcon?: boolean;`
- declare a default value L92: `withIcon = false`
- pass the value to the Status Component L 170 and L 161 : withIcon={withIcon}
- in frontend\taipy-gui\src\components\Taipy\Status.tsx declare and use the `withIcon`
- add tests to StatusList and Status

frontend/taipy-gui/src/components/Taipy/StatusList.spec.tsx
FredLL-Avaiga: you'll have to modify it to test the new property

frontend/taipy-gui/src/components/Taipy/Status.tsx
FabienLelaquais: @pravintargaryen @FredLL-Avaiga @FlorianJacta 
Before a proposal for implementation is done, I would like to have some requirements.
I really don't like the `with_icon` proposal that addresses only the issue (which is good) but makes little sense for future improvements.

This 'all or nothing' approach sounds really bad to me. We should let users decide what to show for what status priority.
Maybe some application has a special need where a new status priority would be useful as well (say the application enters a disastrous mode, where users would want to create a status line like `status = ("disaster", "Out of memory.")`.
Each status priority could have an associated icon, accessed in a dictionary for example.

So the default would be a dict: 
```
    icons = { 
        "S": <success-icon>,
        "E": <error-icon>,
        "W":  <warning-icon>,
        "I":  <info-icon>
     }
```
and the "question mark" fallback, we would let users have the possibility of customizing the status priority representation by overloading existing ones or adding their own.

My first proposal would then be a new property called `icons`, holding something that would be a dict that merges with the predefined one.
- Each key would be a status priority.
  When a new status record arrives, we search for a corresponding key: one that, case-insensitively, would best match the recode status priority (today, the rule is 'starts with', which is fine — we can think of something better like 'most matching starting characters').
- Each value would be an Icon object, that we already have all over the place in Taipy.

The result is:
- The status control represents status priorities with a relevant icon by default
- If a user prefers the text version, then the property must be set to a dict that overloads all priority ids with None
- If an application requires a new status priority, it can be decorated using the property.

I'm still not clear on the 'icon selection' rule - but it can be addressed by really listing use cases.

My final point are:
- do not rush implementing features that are unclear or limiting.
- why invent a new type of Icon in Taipy? Or can Icon rely on MaterialUI icons with some sorcellery?

Comments are welcome.
Code, not just yet.

PR #1581, opened 2024-07-24
Prepare for Polars support

FredLL-Avaiga:
push csv production to data accessors

### DIFF
--- a/taipy/gui/.gitignore
+++ b/taipy/gui/.gitignore
@@ -56,3 +56,5 @@ gui/packaging/taipy-gui.gen.d.ts
 
 # GUI Build
 webapp
+
+!data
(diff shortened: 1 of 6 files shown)


namnguyen20999 (approved):
lgtm

taipy/gui/.gitignore
```diff
@@ -56,3 +56,5 @@ gui/packaging/taipy-gui.gen.d.ts
 
 # GUI Build
 webapp
+
+!data
```
FabienLelaquais, line 60: That is too broad a name at the top.
I suspect that for *your* data: *FLLdata*?
Anyway I would not commit that

FabienLelaquais (approved):
👍

taipy/gui/.gitignore
FredLL-Avaiga: I had to do it as someone committed a change to the root .ignore file with Data on it.
We don't want gui/data to be ignored!

PR #1582, opened 2024-07-25
Feature/doc#986 - Document best practice on main script

### LINKED ISSUE #986: Protect main script from being executed outside the main process
Taipy can start subprocesses using a 'spawn' method. Because of that, the main module is re-executed every time a sub-processed is spawned. In order NOT to re-execute some code that should not be re-executed, the good practice Taipy wants to enforce is to add the boilerplate
condition `if __name__ == "__main"` at the top of the main script.

- [ ] Ensure the `if __name__ == "__main"` condition is explicitly documented and explained in the standalone execution mode documentation (user, ref, tuto, etc.) The motivation should be explicit.
- [ ] Ensure every complete example includes the `if __name__ == "__main"` in all chapters

trgiangdo:
Resolves https://github.com/Avaiga/taipy-doc/issues/986

Please review the example in the `README.md` since it can be important.


jrobinAV (dismissed):
Looks good to me.

I just need @fabien to approve it before merging anything.

jrobinAV (approved):

PR #1624, opened 2024-07-31
prevent duplicated gui hooks during registration

dinhlongviolin1:


taipy/gui/hook.py
```diff
@@ -14,6 +14,10 @@ def __init__(self):
         self.__hooks: t.List[Hook] = []
 
     def _register_hook(self, hook: Hook):
+        # Prevent duplicated hooks
+        for h in self.__hooks:
+            if isinstance(hook, type(h)):
```
FredLL-Avaiga: what if hook is of a subclass of h ?
Is it possible ?
should we test type(hook) is type(h) ?
Should a warning message be added ?

FredLL-Avaiga (dismissed):

taipy/gui/hook.py
FabienLelaquais: Indeed I would suspect (although I really cannot see the use case at this point) that having two different hooks registered, one inheriting from the other... could happen.

FabienLelaquais (dismissed):
I'm good with that.

taipy/gui/hook.py
dinhlongviolin1: We can test the exact type by getting by comparing str(type(h)) == str(type(hook)). That would prevent the inheritance issue.

I will add a little warning as recommended

FredLL-Avaiga: why not `type(hook) is type(h)` ?

dinhlongviolin1: yes that is what I did :D

FabienLelaquais (commented):
👍

FabienLelaquais (approved):
👍👍👍

PR #1634, opened 2024-08-05
feat: replace setup.py with pyproject.toml

joaoandre-avaiga:


toan-quach (dismissed):
lgtm :D

taipy/gui/pyproject.toml
```diff
@@ -0,0 +1,58 @@
...
+Source = "https://github.com/Avaiga/taipy"
+Download = "https://pypi.org/project/taipy/#files"
+Tracker = "https://github.com/Avaiga/taipy/issues"
+Security = "https://github.com/Avaiga/taipy?tab=security-ov-file#readme"
+"Release notes" = "https://docs.taipy.io/en/release-0.0.0/relnotes/"  # will be set dynamically
+
+[tool.setuptools]
+zip-safe = false
+
+[tool.setuptools.package-data]
```
FabienLelaquais, line 53: I don't see the 'taipy/gui/viselements.json' file that is mandatory.
Maybe somewhere else?

pyproject.toml
```diff
@@ -1,3 +1,67 @@
...
+Source = "https://github.com/Avaiga/taipy"
+Download = "https://pypi.org/project/taipy/#files"
+Tracker = "https://github.com/Avaiga/taipy/issues"
+Security = "https://github.com/Avaiga/taipy?tab=security-ov-file#readme"
+"Release notes" = "https://docs.taipy.io/en/release-0.0.0/relnotes/"  # version will be dynamically set
+
+[tool.setuptools.packages.find]
+include = ["taipy", "taipy.*"]
+
+[tool.setuptools.package-data]
```
FabienLelaquais, line 56: I don't see the 'taipy/gui_core/viselements.json' file that is mandatory.
Maybe somewhere else?

FabienLelaquais (dismissed):
That's really cool!
Make sure my two comments are addressed though.

pyproject.toml
dinhlongviolin1: it should be included by the manifest.in file

dinhlongviolin1 (dismissed):
LGTM

taipy/gui/pyproject.toml
joaoandre-avaiga: I just put the same files that were referenced in setup.py. Which was only the version.json.
```
    data_files=[("version", ["version.json"])],
```
We do however have `viselements.json` from gui and gui_core on the respective MANIFEST.in files for taipy-gui and taipy.

PR #1635, opened 2024-08-05
Decouple add base url support

dinhlongviolin1:
Related to Avaiga/taipy-designer#143

### DIFF
--- a/frontend/taipy-gui/base/src/app.ts
+++ b/frontend/taipy-gui/base/src/app.ts
@@ -7,6 +7,7 @@ import { DataManager, ModuleData } from "./dataManager";
 import { initSocket } from "./socket";
 import { TaipyWsAdapter, WsAdapter } from "./wsAdapter";
 import { WsMessageType } from "../../src/context/wsUtils";
+import { getBase } from "./utils";
 
 export type OnInitHandler = (taipyApp: TaipyApp) => void;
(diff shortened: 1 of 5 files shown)


frontend/taipy-gui/base/src/app.ts
```diff
@@ -252,6 +253,10 @@ export class TaipyApp {
     getWsStatus() {
         return this._ackList;
     }
+
+    getBaseUrl() {
+        return getBase();
```
FabienLelaquais, line 258: Why not... rename `getBase()`  to `getBaseUrl()`?

frontend/taipy-gui/base/src/utils.ts
```diff
@@ -0,0 +1,3 @@
+export const getBase = () => {
```
FabienLelaquais, line 1: Same thing here. Use the same terminology for the same concept...

FabienLelaquais (dismissed):
Waiting to see Designer...

frontend/taipy-gui/base/src/app.ts
dinhlongviolin1: I dont want it to conflict with the original getBaseUrl from taipy-gui hence the shortened version. I can still change it back though

FabienLelaquais (approved):
👍

PR #1640, opened 2024-08-06
feat: add execution duration as a information field in Job and Submission entities

### LINKED ISSUE #1544: Add job duration as new information for jobs
### Description

The goal of this issue is to add new information to the job: the job duration. Between the beginning of the execution of the related task and the end of its execution.

This information can be helpful for developers and end users to follow their submission and to set expectation about how long a job will take.

This new information could then be displayed inside the job_selector. 

### Acceptance Criteria

### Code of Conduct

trgiangdo:
Resolves #1544 

Now, in Job and Submission entities, there are new information fields:
- `execution_started_at`
- `execution_ended_at`
- `execution_duration`

In the Job entity, only the `execution_started_at` and `execution_ended_at` are stored in the model. `execution_duration` is calculated from the other 2.

In the Submission entity, `execution_started_at`, `execution_ended_at`, and `execution_duration` are calculated from the `Submission.jobs`.

### DIFF
--- a/taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
@@ -9,6 +9,7 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
+import datetime
 from typing import Optional
 
 from ...job.job import Job
@@ -44,5 +45,7 @@ class _DevelopmentJobDispatcher(_JobDispatcher):
         Parameters:
             job (Job^): The job to submit on an executor with an available worker.
         """
+        job.execution_started_at = datetime.datetime.now()
         rs = _TaskFunctionWrapper(job.id, job.task).execute()
         self._update_job_status(job, rs)
+        job.execution_ended_at = datetime.datetime.now()
--- a/taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
+++ b/taipy/core/_orchestrator/_dispatcher/_standalone_job_dispatcher.py
@@ -9,6 +9,7 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
+import datetime
 import multiprocessing as mp
 from concurrent.futures import Executor, ProcessPoolExecutor
 from functools import partial
@@ -59,6 +60,8 @@ class _StandaloneJobDispatcher(_JobDispatcher):
             self._nb_available_workers -= 1
             self._logger.debug(f"Setting nb_available_workers to {self._nb_available_workers} in the dispatch method.")
         config_as_string = _TomlSerializer()._serialize(Config._applied_config)  # type: ignore[attr-defined]
+
+        job.execution_started_at = datetime.datetime.now()
         future = self._executor.submit(_TaskFunctionWrapper(job.id, job.task), config_as_string=config_as_string)
         future.add_done_callback(partial(self._update_job_status_from_future, job))
 
@@ -67,3 +70,4 @@ class _StandaloneJobDispatcher(_JobDispatcher):
             self._nb_available_workers += 1
             self._logger.debug(f"Setting nb_available_workers to {self._nb_available_workers} in the callback method.")
         self._update_job_status(job, ft.result())
+        job.execution_ended_at = datetime.datetime.now()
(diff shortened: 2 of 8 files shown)


FlorianJacta:
It may be unnecessary to have this inside the job `execution_duration` or not. In my mind, I want to have this information inside the job selector. Up to you!

FabienLelaquais:
> It may be unnecessary to have this inside the job `execution_duration` or not. In my mind, I want to have this information inside the job selector. Up to you!

I want to see use cases before we decide anything.
I don't think this duration should be stored anywhere.
If you "want it in the job selector", the questions will be how, what format, what for, and what if the job is still running. Things like that.

taipy/core/_orchestrator/_dispatcher/_development_job_dispatcher.py
```diff
@@ -44,5 +45,7 @@ def _dispatch(self, job: Job):
         Parameters:
             job (Job^): The job to submit on an executor with an available worker.
         """
+        job.execution_started_at = datetime.datetime.now()
```
joaoandre-avaiga, line 48: What do you think about creating a context manager to set execution_started_at and execution_end_at? Something like:

```
import datetime
from contextlib import contextmanager

@contextmanager
def track_execution(job):
    job.execution_started_at = datetime.datetime.now()
    try:
        yield
    finally:
        job.execution_ended_at = datetime.datetime.now()
   ```
 And them:
 ```
   def _dispatch(self, job: Job):
    """Dispatches the given `Job` on an available worker for execution.

    Parameters:
        job (Job): The job to submit on an executor with an available worker.
    """
    with track_execution(job):
        rs = _TaskFunctionWrapper(job.id, job.task).execute()
        self._update_job_status(job, rs)

**Comment by joaoandre-avaiga** [comment]
Does Taipy rest need to be updated as well to have execution_started_at and execution_ended_at?

**Comment by trgiangdo** [comment]
> Does Taipy rest need to be updated as well to have execution_started_at and execution_ended_at?

@joaoandre-avaiga You're right, REST needs to be updated as well, after we decide which fields to be added.

**Comment by trgiangdo** [comment]
@FabienLelaquais @FlorianJacta 
For now, the use-case is:

- If the job is not started, both the start and end time will be None.
- If the job is started and not finished, the end time will be None, and also the duration will be None.
- If the job is finished, the duration is calculated and returned in seconds.

**Review by toan-quach — DISMISSED** [review]
lgtm

**Review by FlorianJacta — DISMISSED** [review]
I created a new issue to encapsulate my need of having the job duration inside the job selector: https://github.com/Avaiga/taipy/issues/1646

Otherwise, this looks good to me.

**Review by toan-quach — APPROVED** [review]

**Review by jrobinAV — COMMENTED** [review]
I would try to be more generic and keep all the timestamps for all the job status changes: SUBMITTED, BLOCKED, PENDING, RUNNING, CANCELED, FAILED, COMPLETED, SKIPPED, ABANDONED.
We should not face any performance impact since a status change already triggers the job update.

With that, we got all the information for the user to compute everything,
like the total execution time, of course, but also the pending time, the running time, the blocked time.

We could have something similar for the submission.

@trgiangdo What do you think?

**Comment by trgiangdo** [comment]
It can be confusing since CANCELED, FAILED, COMPLETED, SKIPPED, ABANDONED are all counted as `is_finished`.

However, I believe we can do it.

For submission entity, we can calculate all information from the job list.

**Comment by jrobinAV** [comment]
> It can be confusing since CANCELED, FAILED, COMPLETED, SKIPPED, ABANDONED are all counted as `is_finished`.
> 
> However, I believe we can do it.
> 
> For submission entity, we can calculate all information from the job list.

Storing the information is one thing, but exposing it is different. We can keep all the job status changes timestamps, and we can expose good properties at the job level that compile the information properly. 
So we can use it like that: 
```
jobs._records == 
    {
        "SUBMITTED": 2022-09-27 18:00:00.000, 
        "RUNNING": 2022-09-27 18:00:01.000, ..., 
        "COMPLETED": 2022-09-27 18:00:02.000}
jobs.execution_time == 2
```

We could add some dynamic properties like the following

``` python
@property
def execution_time():
    if not "SUBMITTED" in self._records:
        return -1
    if "CANCELED" in self._records:
        return self._records["CANCELED"] - self._records["SUBMITTED"]
    elif "FAILED" in self._records:
        return self._records["FAILED"] - self._records["SUBMITTED"]
    elif "COMPLETED" in self._records:
        return self._records["COMPLETED"] - self._records["SUBMITTED"]
    elif "SKIPPED" in self._records:
        return self._records["SKIPPED"] - self._records["SUBMITTED"]
    elif "ABANDONED" in self._records:
        return self._records["ABANDONED"] - self._records["SUBMITTED"]
    else
        return datetime.now - self._records["SUBMITTED"]
```

trgiangdo:
New ticket for improvement has been created: https://github.com/Avaiga/taipy/issues/1704

Issue #1572, opened 2024-07-23
Protect scenario custom properties to contain key already used


jrobinAV:
# motivation:
When calling scenario.foo, there could be a conflict in the interpretation of foo.
It could be:
- an attribute or a property
- a user-defined property (implemented in the `_properties` dict)
- a sequence name
- a task config id
- a data node config id

The purpose of this issue is to avoid conflicts and to ensure property keys are valid python identifiers.

# Proposed solution:
Before setting a new property, check if the new key is already used as a config id, or a sequence name. If so raise a dedicate error.

The solution consists of defining two new methods in the Scenario class:
- The `__setattr__` method
- The setter of the `properties` property.

# Expected behavior

The following test should pass:
```python
def test_get_set_property_to_scenario():
    dn_cfg = Config.configure_data_node("bar")
    s_cfg = Config.configure_scenario("foo", additional_data_node_configs=[dn_cfg], properties={"key": "value"})
    scenario = taipy.create_scenario(s_cfg)

    assert scenario.properties == {"key": "value"}
    assert scenario.key == "value"

    scenario.properties["new_key"] = "new_value"
    scenario.another_key = "another_value"

    assert scenario.key == "value"
    assert scenario.new_key == "new_value"
    assert scenario.another_key == "another_value"
    assert scenario.properties == {"key": "value", "new_key": "new_value", "another_key": "another_value"}

    with pytest.raises(PropertyKeyAlreadyUsed):
        scenario.bar = "KeyAlreadyUsed"

    with pytest.raises(PropertyKeyAlreadyUsed):
        scenario.properties["bar"] = "KeyAlreadyUsed"

```

# Acceptance criteria
- [ ] Implement the change in the Scenario class and other entities with custom properties (data nodes, tasks)
- [ ] Ensure the new code is unit tested, and check that the code coverage is at least 90%.
- [ ] Propagate any change on the demos and run all of them to ensure no breaking change.
- [ ] Ensure any change is well documented and an item is in the release notes

PR #1642, opened 2024-08-07
feature/#1572 attempt to Protect scenario custom properties

toan-quach:
#1572

### DIFF
--- a/taipy/core/data/data_node.py
+++ b/taipy/core/data/data_node.py
@@ -354,12 +354,9 @@ class DataNode(_Entity, _Labeled):
             return super().__setattr__(name, value)
         else:
             protected_attribute_name = _validate_id(name)
-            try:
-                if protected_attribute_name not in self._properties:
-                    raise AttributeError
+            if protected_attribute_name in self._properties:
                 raise AttributeKeyAlreadyExisted(name)
-            except AttributeError:
-                return super().__setattr__(name, value)
+            return super().__setattr__(name, value)
 
     def _get_attributes(self, protected_attribute_name, attribute_name):
         raise AttributeError
--- a/taipy/core/data/data_node.py
+++ b/taipy/core/data/data_node.py
@@ -29,7 +29,7 @@ from .._entity._ready_to_run_property import _ReadyToRunProperty
 from .._entity._reload import _Reloader, _self_reload, _self_setter
 from .._version._version_manager_factory import _VersionManagerFactory
 from ..common._warnings import _warn_deprecated
-from ..exceptions.exceptions import DataNodeIsBeingEdited, NoData
+from ..exceptions.exceptions import AttributeKeyAlreadyExisted, DataNodeIsBeingEdited, NoData
 from ..job.job_id import JobId
 from ..notification.event import Event, EventEntityType, EventOperation, _make_event
 from ..reason import DataNodeEditInProgress, DataNodeIsNotWritten
@@ -139,6 +139,7 @@ class DataNode(_Entity, _Labeled):
     _MANAGER_NAME: str = "data"
     _PATH_KEY = "path"
     __EDIT_TIMEOUT = 30
+    __CHECK_INIT_DONE_ATTR_NAME = "_init_done"
 
     _TAIPY_PROPERTIES: Set[str] = set()
 
@@ -174,6 +175,7 @@ class DataNode(_Entity, _Labeled):
         self._edits: List[Edit] = edits or []
 
         self._properties: _Properties = _Properties(self, **kwargs)
+        self._init_done = True
 
     @staticmethod
     def _new_id(config_id: str) -> DataNodeId:
@@ -347,8 +349,17 @@ class DataNode(_Entity, _Labeled):
     def __setstate__(self, state):
         vars(self).update(state)
 
-    # def __setattr__(self, name: str, value: Any) -> None:
-    #     return super().__setattr__(name, value)
+    def __setattr__(self, name: str, value: Any) -> None:
+        if self.__CHECK_INIT_DONE_ATTR_NAME not in dir(self) or name in dir(self):
+            return super().__setattr__(name, value)
+        else:
+            protected_attribute_name = _validate_id(name)
+            try:
+                if protected_attribute_name not in self._properties:
+                    raise AttributeError
+                raise AttributeKeyAlreadyExisted(name)
+            except AttributeError:
+                return super().__setattr__(name, value)
 
     def _get_attributes(self, protected_attribute_name, attribute_name):
         raise AttributeError
--- a/taipy/core/data/data_node.py
+++ b/taipy/core/data/data_node.py
@@ -347,6 +347,12 @@ class DataNode(_Entity, _Labeled):
     def __setstate__(self, state):
         vars(self).update(state)
 
+    # def __setattr__(self, name: str, value: Any) -> None:
+    #     return super().__setattr__(name, value)
+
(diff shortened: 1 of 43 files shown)


taipy/core/scenario/scenario.py
```diff
@@ -175,11 +176,40 @@ def __hash__(self):
...
+            "_primary_scenario",
+            "_tags",
+            "_properties",
+            "_sequences",
+            "_version",
+            "_submittable_id",
+            "_subscribers",  # from Submittable class
+            "_in_context_attributes_changed_collector",
+            "_is_in_context"  # from _Entity class
+            "name",  # from def name setter
```
trgiangdo: What if we use `Scenario.__dict__.keys()` here?
Will that also cause a recursion error?

trgiangdo (commented):
I think if we can resolve the problem of hard-coding the `entity_taipy_attributes`, we can implement the `__setattr__` and `__getattr__` logic in the `_Entity` class to apply for all entity right?

taipy/core/scenario/scenario.py
toan-quach: well Scenario.__dict__ only contains pre-defined functions or attributes outside of __init__, while we need the ones within __init__ also and they are not set/available until __setattr__ is called

toan-quach: well Scenario.__dict__ only contains pre-defined functions or attributes outside of __init__, while we need the ones within __init__ also and they are not set/available until __setattr__ is called

trgiangdo: right, this is tough

taipy/core/data/data_node.py
```diff
@@ -346,6 +348,21 @@ def __getstate__(self):
...
+        if self.__CHECK_INIT_DONE_ATTR_NAME not in dir(self) or name in dir(self):
+            return super().__setattr__(name, value)
+        else:
+            protected_attribute_name = _validate_id(name)
+            try:
+                if protected_attribute_name not in self._properties:
+                    raise AttributeError
+                raise AttributeKeyAlreadyExisted(name)
+            except AttributeError:
+                return super().__setattr__(name, value)
```
trgiangdo: Does this work?
Raise and catch right away seems weird to me.
```suggestion
            if protected_attribute_name in self._properties:
                raise AttributeKeyAlreadyExisted(name)
            return super().__setattr__(name, value)
```

taipy/core/data/mongo.py
```diff
@@ -143,15 +143,15 @@ def __init__(
             properties.get(self.__COLLECTION_KEY, "")
         ]
 
-        self.custom_document = properties[self._CUSTOM_DOCUMENT_PROPERTY]
+        self.custom_mongo_document = properties[self._CUSTOM_DOCUMENT_PROPERTY]
```
jrobinAV: What is the motivation for this rename?
It is a breaking change. If we want to keep it, it has to be documented in the refman, in the userman and in the release notes.

taipy/core/scenario/scenario.py
```diff
@@ -175,22 +178,40 @@ def __hash__(self):
...
-
+    def __setattr__(self, name: str, value: Any) -> None:
+        if self.__CHECK_INIT_DONE_ATTR_NAME not in dir(self) or name in dir(self):
+            return super().__setattr__(name, value)
+        else:
+            protected_attribute_name = _validate_id(name)
+            try:
+                if protected_attribute_name not in self._properties and not self._get_attributes(
+                    protected_attribute_name, name
+                ):
```
trgiangdo:
```suggestion
            try:
                if protected_attribute_name not in self._properties and not self._get_attributes(name):
```

taipy/core/scenario/scenario.py
```diff
@@ -175,22 +178,40 @@ def __hash__(self):
...
+                    raise AttributeError
+                raise AttributeKeyAlreadyExisted(name)
+            except AttributeError:
+                return super().__setattr__(name, value)
+
+    def _get_attributes(self, protected_attribute_name, attribute_name) -> Union[Sequence, Task, DataNode]:
```
trgiangdo: The `protected_attribute_name` is not used outside of the `_get_attributes()` right?

We probably can move the `_validate_id()` inside the `_get_attributes()` to make it cleaner.
```suggestion
    def _get_attributes(self, attribute_name) -> Union[Sequence, Task, DataNode]:
        protected_attribute_name = _validate_id(name)
```

trgiangdo (commented):
I have a few suggestions.

These can be applied to all `__setattr__()` and `_get_attributes()` of all entity classes

taipy/core/scenario/scenario.py
toan-quach: well we need `_validate_id` in both __setattr__ and __getattr__ so it probably won't help much :/

tests/core/scenario/test_scenario.py
```diff
@@ -156,6 +163,29 @@ def test_create_scenario_and_add_sequences():
...
+
+    assert scenario.properties == {"key": "value"}
+    assert scenario.key == "value"
+
+    scenario.properties["new_key"] = "new_value"
+    scenario.another_key = "another_value"
+
+    assert scenario.key == "value"
+    assert scenario.new_key == "new_value"
+    assert scenario.another_key == "another_value"
```
jrobinAV: It feels strange to me that
You can read a property through `scenario.key == "value"`
but you cannot write it using `scenario.another_key  = "another_value"` 

I did not realize that it can be really confusing.

What do you think? Should we just not expose properties as attributes ?

taipy/core/data/mongo.py
toan-quach: ah haha it was because initially, we already have `custom_document` as a key in `properties`! so by setting it as an attribute is conflicting with what's available in the `properties` :D since we're removing getting and setting attribute from properties, we should be fine reverting this change

tests/core/scenario/test_scenario.py
toan-quach: yep, I think we should not expose `properties`'s keys as attributes, will remove it from setattr and getattr

taipy/core/data/data_node.py
```diff
@@ -174,6 +175,7 @@ def __init__(
         self._edits: List[Edit] = edits or []
 
         self._properties: _Properties = _Properties(self, **kwargs)
+        self._init_done = True
```
jrobinAV: I believe we don't need this anymore. Am I correct?

taipy/core/data/parquet.py
```diff
@@ -189,7 +189,7 @@ def _read_from_path(self, path: Optional[str] = None, **read_kwargs) -> Any:
 
         # return None if data was never written
         if not self.last_edit_date:
-            self._DataNode__logger.warning(
+            self._DataNode.__logger.warning(
```
trgiangdo: I think this change is wrong

toan-quach: yeah it is, but linter is complaining that it doesn't recognize _DataNode__logger :/

trgiangdo: Yes, because it doesn't exist. But `self._DataNode__logger` does. Can you revert this change?

jrobinAV: Why don't we set the logger as a hidden variable instead of private? That would be cleaner for me. No?

trgiangdo: Yes, I think we missed this change when reviewed the work from a contributor.

jrobinAV (approved):
LGTM

PR #1647, opened 2024-08-08
Feature/#1566 - Check children config_id to make sure not overlapping the properties

### LINKED ISSUE #1566: Add config checker to validate the id is not a reserved key
### Description

Because the entities (data nodes, tasks) are exposed as properties of parent entities using their `config_id`s, each config_id must be a valid python identifier, but it must also not be an existing attribute or property of the parent entity. For instance, if a data node config as "name" for id, there will be a conflict between the name property and the name datanode when calling `scenario.name`.

To avoid such conflicts add a checker for ScenarioConfig, TaskConfig, and DataNodeConfig to ensure the id is not a reserved keyword. The reserved keywords are all the attributes and properties of scenarios, tasks, data nodes, and parent classes.

### Solution Proposed

Add a new checker, `ReservedKeyWordChecker,` which returns an error when the ID of a section is among a list of reserved keywords.

### Acceptance Criteria

- [ ] Ensure the new code is unit tested, and check that the code coverage is at least 90%.
- [ ] Create related issues in taipy-doc for documentation and Release Notes.
- [ ] Ensure any change is well documented in the Ref man (in the `Config.configure_...` methods) and the userman.

### Code of Conduct

trgiangdo:
Resolves #1566

### DIFF
--- a/taipy/core/config/checkers/_scenario_config_checker.py
+++ b/taipy/core/config/checkers/_scenario_config_checker.py
@@ -46,22 +46,24 @@ class _ScenarioConfigChecker(_ConfigChecker):
     def _check_if_children_config_id_is_overlapping_with_properties(
         self, scenario_config_id: str, scenario_config: ScenarioConfig
     ):
-        for task in scenario_config.tasks:
-            if task.id in scenario_config.properties:
-                self._error(
-                    TaskConfig._ID_KEY,
-                    task.id,
-                    f"The id of the TaskConfig `{task.id}` is overlapping with the "
-                    f"property `{task.id}` of ScenarioConfig `{scenario_config_id}`.",
-                )
-        for data_node in scenario_config.data_nodes:
-            if data_node.id in scenario_config.properties:
-                self._error(
-                    DataNodeConfig._ID_KEY,
-                    data_node.id,
-                    f"The id of the DataNodeConfig `{data_node.id}` is overlapping with the "
-                    f"property `{data_node.id}` of ScenarioConfig `{scenario_config_id}`.",
-                )
+        if scenario_config.tasks:
+            for task in scenario_config.tasks:
+                if isinstance(task, TaskConfig) and task.id in scenario_config.properties:
+                    self._error(
+                        TaskConfig._ID_KEY,
+                        task.id,
+                        f"The id of the TaskConfig `{task.id}` is overlapping with the "
+                        f"property `{task.id}` of ScenarioConfig `{scenario_config_id}`.",
+                    )
+        if scenario_config.data_nodes:
+            for data_node in scenario_config.data_nodes:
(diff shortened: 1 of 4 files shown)


taipy/core/config/checkers/_scenario_config_checker.py
```diff
@@ -38,10 +38,33 @@ def _check(self) -> IssueCollector:
...
                 self._check_comparators(scenario_config_id, scenario_config)
 
         return self._collector
 
+    def _check_if_children_config_id_is_overlapping_with_properties(
+        self, scenario_config_id: str, scenario_config: ScenarioConfig
+    ):
+        if scenario_config.tasks:
+            for task in scenario_config.tasks:
+                if isinstance(task, TaskConfig) and task.id in scenario_config.properties:
```
toan-quach, line 51: I understand why we want to have `isinstance` here but is it really necessary?

toan-quach (approved):
lgtm, minor inquiry only

taipy/core/config/checkers/_scenario_config_checker.py
trgiangdo: Yes, because if the `task` is not a TaskConfig, the `task.id` will fail and raise an AttributeError.
In the checker, we only want to log the error, not raise anything directly here.

toan-quach: in that case, should the log.error show a separate message to make it more concise?

trgiangdo: The case where the `task` is not TaskConfig is handle by `_check_task_configs()` method, so there is a error log their.

toan-quach: hmm yeah okay then :D

taipy/core/config/checkers/_scenario_config_checker.py
```diff
@@ -38,10 +38,33 @@ def _check(self) -> IssueCollector:
...
+        if scenario_config.data_nodes:
+            for data_node in scenario_config.data_nodes:
+                if isinstance(data_node, DataNodeConfig) and data_node.id in scenario_config.properties:
+                    self._error(
+                        DataNodeConfig._ID_KEY,
+                        data_node.id,
+                        f"The id of the DataNodeConfig `{data_node.id}` is overlapping with the "
+                        f"property `{data_node.id}` of ScenarioConfig `{scenario_config_id}`.",
+                    )
+
```
jrobinAV, line 67: What about adding a similar check for sequences?
keys of the scenario_config.sequences dictionary should not overlap data node config ids or task config ids.
@trgiangdo what do you think?

trgiangdo: That's a good point. Let me add this

PR #1650, opened 2024-08-08
Scenario_selector select scenario after on_creation

### LINKED ISSUE #1619: [🐛 BUG] Creating a new scenario in the scenario selector with *on_creation* should select it
**Problem**: When I create a new scenario with the *on_creation* callback, I start editing its properties. Since the new scenario was not selected when I created it, I edited the properties of a previously selected scenario. 

**Expected behavior:**
Creating a new scenario through the scenario selector and with *on_creation* should automatically select the new scenario.

core_issues_1

Similar to: https://github.com/Avaiga/taipy/issues/1074 but with *on_creation* callback.

Quick workaround:

Having `state.scenario = scenario` in the callback.

FredLL-Avaiga:
resolves #1619

### DIFF
--- a/taipy/gui_core/_context.py
+++ b/taipy/gui_core/_context.py
@@ -414,6 +414,7 @@ class _GuiCoreContext(CoreEventConsumerBase):
         data = args[start_idx + 2]
         with_dialog = True if len(args) < start_idx + 4 else bool(args[start_idx + 3])
         scenario = None
+        user_scenario = None
 
         name = data.get(_GuiCoreContext.__PROP_ENTITY_NAME)
         if update:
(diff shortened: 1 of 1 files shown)


taipy/gui_core/_context.py
```diff
@@ -502,10 +504,10 @@ def crud_scenario(self, state: State, id: str, payload: t.Dict[str, str]):  # no
                 state.assign(error_var, f"Error creating Scenario. {e}")
             finally:
                 self.scenario_refresh(scenario_id)
-                if scenario and (sel_scenario_var := args[1] if isinstance(args[1], str) else None):
+                if (scenario or user_scenario) and (sel_scenario_var := args[1] if isinstance(args[1], str) else None):
                     try:
                         var_name, _ = gui._get_real_var_name(sel_scenario_var)
-                        state.assign(var_name, scenario)
+                        state.assign(var_name, scenario or user_scenario)
```
FabienLelaquais, line 510: Is there a risk this raises an exception?

FabienLelaquais (approved):
👍

taipy/gui_core/_context.py
FredLL-Avaiga: the risk is that the variable name is not found in the default context

Issue #1632, opened 2024-08-04
IDE autocompletion doc is not that great


FabienLelaquais:
### What went wrong? 🤔

When using the Page Builder API in IDEs, the autocompletion works pretty well. We get all available property names.
The doc shows the type more or less (there' s this strange *dynamic* thing).

The function signatures should display the property name, type, and default value for every function.
The default property should always appear first as a positional argument, and all the others should appear as keywork arguments.
The doc should have an introduction text ("Creates a &lt;element-name> element.")  and at least one line per property, and not show the type or default value (that appears in the signature).

The pyi file should for example look like:
```
class text(_Control):
    _ELEMENT_NAME: str
    def __init__(
        self,
        value:str="",
       *, # remaining parameters *must* be keyword parameters
        raw:bool=False,
        mode:str|None=None,
        format:str|None=None,
        id:str|None=None,
        properties:dict[str, any]|None=None,
        class_name:str|None=None,
        hover_text:str|None=None,
    ) -> None:
        """
        Creates a text element.

        ### Arguments
        value (dynamic): The value displayed as text by this control.

        raw: If set to True, the component renders as an HTML \<span\> element without any default style.

        mode: Define the way the text is processed:

        - "raw": synonym for setting the `raw` property to True
        - "pre": keeps spaces and new lines
        - "markdown" or "md": basic support for Markdown.

        format: The format to apply to the value.

        id: The identifier that is assigned to the rendered HTML component.

        properties: Bound to a dictionary that contains additional properties for this element.

        class_name (dynamic): The list of CSS class names that are associated with the generated HTML Element. These class names are added to the default `taipy-text` class name.

        hover_text (dynamic): The information that is displayed when the user hovers over this element.
        """
        ...
```

### Steps to Reproduce Issue

1. A code fragment
2. And/or configuration files or code
3. And/or Taipy GUI Markdown or HTML files

### Screenshots

![DESCRIPTION](LINK.png)

### Acceptance Criteria

- [ ] Ensure new code is unit tested, and check code coverage is at least 90%.
- [ ] Create related issue in taipy-doc for documentation and Release Notes.

### Code of Conduct

- [X] I have checked the [existing issues](https://github.com/Avaiga/taipy/issues?q=is%3Aissue+).
- [ ] I am willing to work on this issue (optional)

PR #1653, opened 2024-08-09
Improve content of generated pyi for the Page Builder API

FabienLelaquais:
Addresses #1632.

Change Page Builder API pyi generation script to provide nearly full function signatures.
Better rendering of element properties documentation.

+ Updated actions (script can run from taipy's repo root dir)
+ Details in element examples

### DIFF
--- a/.github/actions/gui-test/pyi/action.yml
+++ b/.github/actions/gui-test/pyi/action.yml
@@ -8,7 +8,7 @@ runs:
       run: pipenv run pip install mypy black isort
     - name: Generate pyi
       shell: bash
-      run: cp tools/gui/generate_pyi.py pyi_temp.py && pipenv run python pyi_temp.py && rm pyi_temp.py
+      run: pipenv run python tools/gui/generate_pyi.py
     - name: Cleanup any untracked files
       shell: bash
       run: git clean -f
--- a/tools/gui/generate_pyi.py
+++ b/tools/gui/generate_pyi.py
@@ -26,7 +26,7 @@ if not os.path.isdir(os.path.abspath(os.path.join(script_dir, "taipy"))):
 # ##################################################################################################
 gui_py_file = "./taipy/gui/gui.py"
 gui_pyi_file = f"{gui_py_file}i"
-from taipy.config import Config
+from taipy.config import Config  # noqa: E402
 
 # Generate Python interface definition files
 os.system(f"pipenv run stubgen {gui_py_file} --no-import --parse-only --export-less -o ./")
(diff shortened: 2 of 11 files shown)


.github/actions/gui-test/pyi/action.yml
```diff
@@ -8,7 +8,7 @@ runs:
       run: pipenv run pip install mypy black isort
     - name: Generate pyi
       shell: bash
-      run: cp tools/gui/generate_pyi.py pyi_temp.py && pipenv run python pyi_temp.py && rm pyi_temp.py
```
FredLL-Avaiga, line 11: why was it through a copy @dinhlongviolin1 ?

tools/gui/generate_pyi.py
```diff
@@ -44,123 +45,152 @@
...
-with open("./tools/gui/builder/block.txt", "r") as file:
-    block_template = file.read()
-with open("./tools/gui/builder/control.txt", "r") as file:
-    control_template = file.read()
-
 os.system(f"pipenv run stubgen {builder_py_file} --no-import --parse-only --export-less -o ./")
 
 with open(builder_pyi_file, "a") as file:
     file.write("from ._element import _Element, _Block, _Control\n")
+    file.write("from typing import Union")
```
FredLL-Avaiga: should this import be first to please the linter ?

tools/gui/generate_pyi.py
```diff
@@ -44,123 +45,152 @@
...
+        property["indexed"] = ""
+    if m := re.match(r"dynamic\((.*)\)", type):
+        type = m[1]
+        property["dynamic"] = " (dynamic)"
+    else:
+        property["dynamic"] = ""
+    if type == "Callback" or type == "Function":
+        type = ""
+    else:
+        type = re.sub(r"(int|float|str)\[\]", r"list[\1]", type)
```
FredLL-Avaiga: is it not better to change viselements.json ?

FredLL-Avaiga (dismissed):

tools/gui/generate_pyi.py
FabienLelaquais: This line is now obsolete.
You are right, and I did change viselements.

tools/gui/generate_pyi.py
FabienLelaquais: It is relocated properly because the entire file is formatted.
But it certainly would make sense, indeed.

.github/actions/gui-test/pyi/action.yml
FabienLelaquais: Because of the import of taipy itself.

FredLL-Avaiga (approved):
👍

dinhlongviolin1 (approved):
LGTM

PR #1661, opened 2024-08-12
fix typing in generate-pyi

FredLL-Avaiga:
and use stefanzweifel/git-auto-commit-action v5 (that uses node.js 20 ie no warning)


FabienLelaquais (approved):
Ok. This one I saw coming.

PR #1662, opened 2024-08-13
feat: add lru_cache decorator to manager build functions

joaoandre-avaiga:


toan-quach (approved):

trgiangdo (approved):
LGTM. We can also add the same decorator to the `_build_repository` methods on enterprise

joaoandre-avaiga:
> LGTM. We can also add the same decorator to the `_build_repository` methods on enterprise

I will do that. Thanks!

PR #1664, opened 2024-08-13
Adding logic to handle the issue when height/width too small

### LINKED ISSUE #1535: Improve style of metric control
### Description

Having two metrics takes all my page. Metric takes 100% of my screen. This is unusable by itself. 

image

The size must be configurable.

This should look like this by default:

image

### Acceptance Criteria

### Code of Conduct

namnguyen20999:
Resolve https://github.com/Avaiga/taipy/issues/1535

Use the `width` and `height` properties to define dimensions within the layout. However, these properties should only set the width and height if they haven't already been specified in the `layout` object. The layout's own definitions should take precedence over the `width` and `height` properties.

```
from taipy.gui import Gui

value = 50

layout = {
    "width": 500,
    "height": 500,
}

page = """
**The layout has priority over the width and height properties.**
<|{value}|metric|layout={layout}|width=300|height=300|> 

**The width and height properties will be used to define the width and height of the Metric's layout.**
<|{value}|metric|width=300|height=300|>
"""
```

image


frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -166,7 +166,11 @@ const Metric = (props: MetricProps) => {
     const skelStyle = useMemo(() => ({ ...style, minHeight: "7em" }), [style]);
 
     const layout = useMemo(() => {
-        const layout = { ...baseLayout };
+        const layout = {
+            ...baseLayout,
+            height: baseLayout.height !== undefined ? (baseLayout.height >= 10 ? baseLayout.height : 10) : null,
```
FredLL-Avaiga: Why null why not undefined ?

FredLL-Avaiga (changes requested):
Can you change your code example in the PR to show the new behavior?

FlorianJacta (commented):
Can we change width and height directly through the "width" and "height" props of the metric and not through *layout*?

frontend/taipy-gui/src/components/Taipy/Metric.tsx
namnguyen20999: updated

namnguyen20999:
@FlorianJacta  We can do that, but the width and height in the layout control the CSS applied to the Metric within the container, whereas width and height props control the CSS applied to the Metric's container. 

```
from taipy.gui import Gui

value = 50

layout = {
    "margin": {
        "t": 1,
        "b": 1,
        "l": 1,
        "r": 1
    },
    "height": 300,
    "width": 300
}



page = """
<|{value}|metric|layout={layout}|>
<|{value}|metric|height=300px|width=300px|>
"""
```

You could see the differences here:
Screenshot 2024-08-14 at 13 04 44

FlorianJacta:
> @FlorianJacta We can do that, but the width and height in the layout control the CSS applied to the Metric within the container, whereas width and height props control the CSS applied to the Metric's container.
> 
> ```
> from taipy.gui import Gui
> 
> value = 50
> 
> layout = {
>     "margin": {
>         "t": 1,
>         "b": 1,
>         "l": 1,
>         "r": 1
>     },
>     "height": 300,
>     "width": 300
> }
> 
> 
> 
> page = """
> <|{value}|metric|layout={layout}|>
> <|{value}|metric|height=300px|width=300px|>
> """
> ```
> 
> You could see the differences here: Screenshot 2024-08-14 at 13 04 44

Is this something we can improve? There is an issue to make the metric component more usable in terms of style

FredLL-Avaiga:
If you don't specify anything else in the layout, both example render exactly the same.
I propose that the width and height properties define the width and height values in the layout  (if not defined ie layout definition has precedence) and have no impact on the style.
What do you think @namnguyen20999 @FlorianJacta ?

FlorianJacta:
Why not!

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -191,12 +184,11 @@ const Metric = (props: MetricProps) => {
         return layout as Partial<Layout>;
     }, [props.title, props.template, props.template_Dark_, props.template_Light_, theme.palette.mode, baseLayout]);
 
-    const plotConfig = {displaylogo: false}
```
FredLL-Avaiga, line 194: why did you remove this ?

FredLL-Avaiga: by the way it should be declared outside the component

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -45,8 +45,6 @@ interface MetricProps extends TaipyBaseProps, TaipyHoverProps {
     testId?: string;
     layout?: string;
     defaultLayout?: string;
-    style?: string;
```
FredLL-Avaiga, line 48: Should this be removed from factory.py and viselements.json ?

FredLL-Avaiga (changes requested):
there was a decision to remove the plotly logo

frontend/taipy-gui/src/components/Taipy/Metric.tsx
namnguyen20999: I've added it back. For a moment I forgot this will display the plotly logo

frontend/taipy-gui/src/components/Taipy/Metric.tsx
namnguyen20999: We don't have it in factory.py and viselements.json at the moment.

FredLL-Avaiga (approved):
Is the issue for width < 10 not a problem anymore ?

namnguyen20999:
> Is the issue for width < 10 not a problem anymore ?

When the width and height are less than 10, the plotly will revert to its default height of 700px and width of 450px. We just need to indicate it in our docs

PR #1679, opened 2024-08-19
port auto ranges (#1519)

### LINKED ISSUE #1519: [🐛 BUG] port="auto" does not work everytime
### What went wrong? 🤔

Sometimes, when using `port="auto"`, an error arises and stops the application from being run.

```python
[2024-07-15 14:39:11.409][Taipy][INFO]  * Server starting on http://127.0.0.1:57990
 * Serving Flask app 'Taipy'
 * Debug mode: on
An attempt was made to access a socket in a way forbidden by its access permissions
```

### Expected Behavior

It would be best if it works everytime.

### Steps to Reproduce Issue

Put `port="auto"` on an app and test many times.

### Version of Taipy

develop - 7/15/24

### Acceptance Criteria

The problem may come from a backend configuration issue, which is NOT our responsibility.

Plan could be to extend the 'port' syntax to a comma-separated list of ports or port ranges - simplifying avoiding wrong firewall settings.

### Code of Conduct

dinhlongviolin1:
Resolve #1519 

port ranges are weighted on the length of the ranges


taipy/gui/server.py
```diff
@@ -267,17 +267,25 @@ def _apply_patch(self):
             if not patcher.is_monkey_patched("time"):
                 monkey_patch(time=True)
 
-    def _get_random_port(self):  # pragma: no cover
+    def _get_random_port(self, port_auto_ranges: t.Optional[t.List[t.Union[int, t.Tuple[int, int]]]] = None):  # pragma: no cover
+        port_auto_ranges = port_auto_ranges or [(49152, 65535)]
+        random_weights = [
+            1 if isinstance(r, int) else r[1] - r[0] + 1 for r in port_auto_ranges
```
FredLL-Avaiga: Are you supposing that the first element of the tuple is lesser than the second one ?

FredLL-Avaiga (dismissed):

taipy/gui/server.py
dinhlongviolin1: you are right on this one I guess, never know what the user might enter. I will get the abs value

taipy/gui/server.py
```diff
@@ -267,17 +267,35 @@ def _apply_patch(self):
             if not patcher.is_monkey_patched("time"):
                 monkey_patch(time=True)
 
-    def _get_random_port(self):  # pragma: no cover
+    def _get_random_port(
+        self, port_auto_ranges: t.Optional[t.List[t.Union[int, t.Tuple[int, int]]]] = None
+    ):  # pragma: no cover
+        port_auto_ranges = port_auto_ranges or [(49152, 65535)]
+        random_weights = [1 if isinstance(r, int) else abs(r[1] - r[0]) + 1 for r in port_auto_ranges]
```
FredLL-Avaiga, line 274: and hopefully the tuple will have 2 elements :-D

FredLL-Avaiga (dismissed):

taipy/gui/server.py
dinhlongviolin1: It will raise an error :D their fault not ours

FlorianJacta:
TBH, I don't really see the use of this feature. I have no idea what port range I should put. Is it not possible to catch when a port is used and use another port?

dinhlongviolin1:
The reason why sometimes it is not working is because of firewall. We cant do anything to control the user's firewall and the best thing to provide is to allow the user to declare a list of ports that they want to use for the demo. I did check for opening ports before choosing one that is not used. But checking whether a port can be opened is really not good since we should not have an overhead of having to start a full application up just to do it again because of the port not being available on the user side.

FabienLelaquais (commented):
A bit of doc indeed. Nice though

FabienLelaquais (dismissed):
👍

FredLL-Avaiga (approved):

PR #1684, opened 2024-08-20
fix pipe warning (#467)

### LINKED ISSUE #467: Missing pipe warning should be more explicit
When a user writes a visual element with a missing pipe at the end:

```python
page = "<|{data}|table|style[questions]=style_questions>"
```

The page fails to display the element (which is normal) but prompts the user with the following warning:
```
--- 1 warning(s) were found for page '/' in variable 'page' ---
 - Warning 1: Invalid tag name 'table' in line 1.
```

This warning is confusing and should instead suggest that the syntax of the visual element might be incorrect

dinhlongviolin1:
Resolve #467


taipy/gui/_renderers/_markdown/preproc.py
```diff
@@ -100,7 +111,7 @@ def run(self, lines: List[str]) -> List[str]:
                         line += f' {property[0]}="{prop_value}"'
                     line += _MarkdownFactory._TAIPY_END + new_line_delimeter
                 else:
-                    _warn(f"Invalid tag name '{tag}' in line {line_count}.")
+                    _warn(f"Failed to recognized block tag '{tag}' in line {line_count}. Verify that you are closing the tag properly with '|>' if it is a control element.")  # noqa: E501
```
FabienLelaquais:
```suggestion
                    _warn(f"Failed to recognized block tag '{tag}' in line {line_count}. Check that you are closing the tag properly with '|>' if it is a control element.")  # noqa: E501
```

FabienLelaquais (dismissed):
👍

FabienLelaquais (approved):
👍

PR #1692, opened 2024-08-22
Fix - Reload the cache of _build_manager when needed

trgiangdo:
In this PR:
- Reload the cache of all `_build_manager()` method
- Fix an issue where we build the _VersionManager at class level, which build the cache from importing


toan-quach (approved):
ahh ok that makes sense to me! thanks!

Issue #1487, opened 2024-07-08
Data node Viewer - Add file download/upload capability for file based data nodes


jrobinAV:
**Description**
When a data node is stored as a file, it would be handy for the end user to add a way to download or upload the file easily from the data node viewer.

This should help the end user understand its data, particularly when data is corrupted or an unexpected behavior happens.

The developer could enable/disable this capability through a property. By default, it should be enabled.

This issue is the GUI part of ticket #1196 

**Proposal**
1. The download could be accessible through a "Download" button (eventually, an icon when the element width is too small).
2. The upload could be accessible through an "Upload" file selector (eventually, an icon when the element width is too small). 
When a file is passed to the Upload mechanism, Taipy should:
    - Check the file (a check property should let the developer provide a custom function).
    - Write the file at the data node path location. Eventually, overwrite it if a file already exists.
    - track the data node edition

**Usage**
```python
def custom_check(state, file_path) -> bool:
    # Apply the check logic
    return True


<| {selected_datanode} | datanode | file_download=false | file_upload=true | upload_check=custom_check |>
```

trgiangdo:
We now have new APIs on Filebased DataNode for download and upload.
 
- `_get_downloadable_path()` that returns the path to the downloadable data file, and an empty string if the path doesn't exist (in case the data node has no data).
This is so that the user can download the file. I'm not sure if this's enough for a Download button to work.

- `_upload(path, upload_checker)` that check the uploaded file and replace the current data file.
The idea is that when the user uploads a file, it will get stored somewhere. This `_upload()` method receives the uploaded path, try to read it and check it using the optional `upload_checker` method. If passes, the uploaded file will replace the file of the data node. If fails, a ReasonCollection is returned.

The `upload_checker` should be a Python method that provided by the developer, that takes the file name of the upload data and the data itself as arguments, and return a boolean indicates whether the uploaded file is valid or not.

FredLL-Avaiga:
is there a Core API to know if a datanode is file based ? @jrobinAV @trgiangdo ?

jrobinAV:
@FredLL-Avaiga All data nodes based on a file are children of the `_FileDataNodeMixin`
FYI, today we have pickle, csv, excel, json, and parquet.

FredLL-Avaiga:
I would propose that the call to the user checker function would be surrounded by a try catch @jrobinAV @trgiangdo 

[...]

PR #1732, opened 2024-08-31
Feature - Add try catch around the upload_checker call

trgiangdo:
Resolves the comment in https://github.com/Avaiga/taipy/issues/1487?notification_referrer_id=NT_kwDOAj2Ni7QxMTQwOTgzODI3NDozNzU4ODM2Mw#issuecomment-2321588642 by Fred.

If any Exception raised by the `upload_checker`, we log an error and return `InvalidUploadFile` reason


taipy/core/data/_file_datanode_mixin.py
```diff
@@ -135,7 +135,16 @@ def _upload(self, path: str, upload_checker: Optional[Callable[[str, Any], bool]
             return reason_collection
 
         if upload_checker is not None:
-            if not upload_checker(upload_path.name, upload_data):
+            try:
+                can_upload = upload_checker(upload_path.name, upload_data)
+            except Exception as err:
```
FredLL-Avaiga, line 162: Should the exception be added to the reason?
Especially useful on user function

FredLL-Avaiga (approved):
Tx

taipy/core/data/_file_datanode_mixin.py
jrobinAV: We could add an optional exception attached to a reason. It is useful information for developers. It is already in the logs, but making it available in taipy so they can use it in the code makes sense to me.

However, by default, I would not display it in the data node viewer to end-users. 

What do you think?

jrobinAV (approved):

taipy/core/data/_file_datanode_mixin.py
trgiangdo: There is only 1 way of returning a Reason, so I think it's the default.

I'm not sure if we can give the user multiple choices

PR #1766, opened 2024-09-09
make gui hooks private (#1736)

dinhlongviolin1:
Resolve #1736


FabienLelaquais (approved):
👌
So I expect changes in Enterprise+Designer right?

Issue #1756, opened 2024-09-06
[Refactor] Shorten Taipy Templates path


FlorianJacta:
### Description

`taipy\templates\scenario-management\{{cookiecutter.__root_folder_name}}\pages\scenario_page\data_node_management.py'`

We could shorten:
- scenario-management
- __root_folder_name

and maybe other elements.

This will fix :https://github.com/Avaiga/taipy-enterprise/issues/446 https://github.com/Avaiga/taipy/issues/808

### Acceptance Criteria

- [ ] Ensure new code is unit tested, and check code coverage is at least 90%.
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change.
- [ ] Ensure any change is well documented.

### Code of Conduct

- [X] I have checked the [existing issues](https://github.com/Avaiga/taipy/issues?q=is%3Aissue+).
- [ ] I am willing to work on this issue (optional)

jrobinAV:
We can add a warning on taipy doc.

PR #1783, opened 2024-09-12
Shorten the longest path 

jrobinAV:
adress #1756 
Shorten the longest path  by reducing the sdm folder name and the __ro…ot_folder name


FlorianJacta (approved):
LGTM. What about "data_node_management.py" > "data_node_page.py"?

jrobinAV:
> LGTM. What about "data_node_management.py" > "data_node_page.py"?

Since it is not a page, I believe it is confusing. No?
 what about `data_node_mgt.py`

FlorianJacta:
Yes, why not. Maybe, in the future, we will be able to delete this script, as you will be able to upload data with the data node viewer

jrobinAV:
Oh, you are right. And we can already do it in development, right? Did you test it? 
If it works, let's create an issue to remove this partial file.

FlorianJacta:
I did not test it; on my way!

trgiangdo (approved):
LGTM.

FYI, we can actually shorten it even more.
`__root_folder` --> `__root`
`__main_file` --> `__main`

Those are private variables so it won't affect the user. But I understand that it's probably not needed

PR #1809, opened 2024-09-19
Added examples for progress control

FabienLelaquais:
+ removed 'color' property from progress control
+ added CSS class names for value and title in the control
+ minor refman issues


taipy/gui/_renderers/factory.py
```diff
@@ -444,6 +444,24 @@ class _Factory:
...
+            [
+                ("linear", PropertyType.boolean, False),
+                ("show_value", PropertyType.boolean, False),
+                ("title", PropertyType.dynamic_string),
+                ("title_anchor", PropertyType.string, "bottom"),
+                ("render", PropertyType.dynamic_boolean, True),
+                ("width", PropertyType.string_or_number),
+            ]
+        )
+        ._set_propagate(),
```
FredLL-Avaiga: don't think `propagate` is needed here as there's no variable modification from the frontend

FredLL-Avaiga (dismissed):

namnguyen20999 (dismissed):
lgtm

taipy/gui/_renderers/factory.py
FabienLelaquais: Let's discuss this rapidly. I'm sure you're right.

FabienLelaquais: Done.

FredLL-Avaiga (dismissed):

FredLL-Avaiga (approved):

PR #1811, opened 2024-09-19
export taipy gui base as a proper package

dinhlongviolin1:


FabienLelaquais (approved):
🤔

frontend/taipy-gui/base/src/exports.ts
```diff
@@ -1,7 +1,7 @@
...
 import { WsAdapter } from "./wsAdapter";
-// import { TaipyApp } from "./app";
+import { ModuleData } from "./dataManager";
 
-export {
-    WsAdapter,
-    // TaipyApp,
-};
+export default TaipyApp;
+export { TaipyApp, createApp, WsAdapter };
```
FredLL-Avaiga, line 6: why export TaipyApp twice ?

FredLL-Avaiga (approved):
👍

frontend/taipy-gui/base/src/exports.ts
dinhlongviolin1: I want to have a default export in case they want to use it :D

PR #1815, opened 2024-09-21
Fix - load_workbook with read_only option when do reading

trgiangdo:
This PR fixes the failing test on taipy-integration-testing.
The test results can be found in https://github.com/Avaiga/taipy-integration-testing/pull/67.

The reason is that the `openpyxl.load_workbook()` can only be closed when the loaded file is read/write only.
On MacOS and Linux, it's fine. However, on Windows, the file is not cleaned automatically, which makes the next test to fail.

### DIFF
--- a/taipy/core/data/excel.py
+++ b/taipy/core/data/excel.py
@@ -175,7 +175,7 @@ class ExcelDataNode(DataNode, _FileDataNodeMixin, _TabularDataNodeMixin):
     def _read_as(self, path: str):
         try:
             properties = self.properties
-            excel_file = load_workbook(path)
+            excel_file = load_workbook(path, read_only=True)
             exposed_type = properties[self._EXPOSED_TYPE_PROPERTY]
             work_books = {}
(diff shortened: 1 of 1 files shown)


jrobinAV:
How did you figure it out? Is there any reason for this?

In any case, that's excellent!

jrobinAV (approved):
Great job!

trgiangdo:
> How did you figure it out? Is there any reason for this?

From [the docstring of the `openpyxl.close()` method](https://openpyxl.readthedocs.io/en/stable/api/openpyxl.workbook.workbook.html#openpyxl.workbook.workbook.Workbook.close)

jrobinAV:
> > How did you figure it out? Is there any reason for this?
> 
> From [the docstring of the `openpyxl.close()` method](https://openpyxl.readthedocs.io/en/stable/api/openpyxl.workbook.workbook.html#openpyxl.workbook.workbook.Workbook.close)

Yep. In other words: "I should have read the fucking doc!" 🤣

PR #1818, opened 2024-09-23
Doc-related updates focusing on the table edition features.

FabienLelaquais:
Addresses the 'taipy' part of [taipy-doc#1036](https://github.com/Avaiga/taipy-doc/issues/1036).

- Builder API generated with None default value when needed
- All __init__ files fixed in doc examples directories
- Added dynamic styling example
- Added doc for style parameter in page classes


doc/gui/examples/blocks/pane_simple_lambda/__init__.py
```diff
@@ -1 +1 @@
-# This file makes this directory a module on its own, mandatody for mypy.
```
FredLL-Avaiga, line 1: good catch (multiple)

FredLL-Avaiga (approved):

doc/gui/examples/blocks/pane_simple_lambda/__init__.py
FabienLelaquais: 👍

PR #1822, opened 2024-09-23
Fix/#680 - S3ObjectDtaNode.read should return the binary form of the data

### LINKED ISSUE #680: S3Object Data Node - possibility to change encoding
# Failed Test : 🔥 

It does n't read on images , video  and audio data .
This is because we are reading `uft-8 `encodings [here](https://github.com/Forchapeatl/taipy/blob/65ad882f4a9a8df90d9c5bc617a7f1903faa968b/taipy/core/data/aws_s3.py#L152) which is fair :)

removing `.decode('utf-8')` will mean the `s3ObjectDataNode.read() ` will have to be called as  `s3ObjectDataNode.read().decode('utf-8')` when trying to access text files.

I don't think this `s3ObjectDataNode.read().decode('utf-8')`  fuction call is the expected utility of the Datanode class `read()` method. Besides,  this is an overkill !

I would like to propose that the write **class method should enable one more property ** to the `S3ObjectDataNode`. This  [ACL](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html) - Permision management parameter.
Adding a new property to the s3ObjectDataNode will solve this media object reading trouble. **It will allow users to read s3 object  media files by Url**.

```diff
    __AWS_ACCESS_KEY_ID = "aws_access_key"
    __AWS_SECRET_ACCESS_KEY = "aws_secret_access_key"
    __AWS_STORAGE_BUCKET_NAME = "aws_s3_bucket_name"
    __AWS_S3_OBJECT_KEY = "aws_s3_object_key"
    __AWS_REGION = "aws_region"
+   __AWS_ACL= " ACL='private'|'public-read'|'public-read-write'|'authenticated-read'|'aws-exec-read'|'bucket-owner-read'|'bucket-owner-full-control"
    __AWS_S3_OBJECT_PARAMETERS = "aws_s3_object_parameters"
```

```diff
    def _write(self, data: Any):
        self._s3_client.put_object(
            Bucket=self.properties[self.__AWS_STORAGE_BUCKET_NAME],
            Key=self.properties[self.__AWS_S3_OBJECT_KEY],
            Body=data,
+           ACL= self.properties[self.__AWS_ACL],
        )
```

With the new property set to` public-read` _( ACL = public_read )_ the user will be able to access media files (image, vide , audio) files via url . 

```

page ="""
data = "https://bealoabucket.s3.us-west-2.amazonaws.com/taipy.png"
 <|{data}|image|> 
"""

```
This means the read method will handle ASCII data only and the user will be free to read Media files via url. 

I will like to leave this open for discussion. I just didn't want to forget , we can transfer this to a new issue when this PR is handled.

_Originally posted by @Forchapeatl in https://github.com/Avaiga/taipy/issues/585#issuecomment-1849066118_

trgiangdo:
Resolves #680

The `S3ObjectDtaNode.read()` method now returns the binary form of the data.

The developer will need to handle the decoding of the file them self.

### DIFF
--- a/taipy/core/data/aws_s3.py
+++ b/taipy/core/data/aws_s3.py
@@ -152,7 +152,7 @@ class S3ObjectDataNode(DataNode):
             Bucket=properties[self.__AWS_STORAGE_BUCKET_NAME],
             Key=properties[self.__AWS_S3_OBJECT_KEY],
         )
-        return aws_s3_object["Body"].read().decode("utf-8")
+        return aws_s3_object["Body"].read()
 
     def _write(self, data: Any):
         properties = self.properties
--- a/tests/core/data/test_aws_s3_data_node.py
+++ b/tests/core/data/test_aws_s3_data_node.py
@@ -9,9 +9,16 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
+import os
+import pathlib
+import pickle
+from io import BytesIO
+
 import boto3
+import pandas as pd
 import pytest
 from moto import mock_s3
+from pandas.testing import assert_frame_equal
 
 from taipy.config import Config
 from taipy.config.common.scope import Scope
@@ -52,9 +59,8 @@ class TestS3ObjectDataNode:
         ],
     )
     @pytest.mark.parametrize("properties", __properties)
-    def test_write(self, properties, data):
+    def test_write_text(self, properties, data):
         bucket_name = properties["aws_s3_bucket_name"]
-        # Create an S3 client
         s3_client = boto3.client("s3")
         # Create a bucket
         s3_client.create_bucket(Bucket=bucket_name)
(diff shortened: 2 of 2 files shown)


toan-quach (approved):
lgtm

jrobinAV (approved):
We should merge it as it's better than what we had before.

However, the added value of exposing the s3_objects seems greater if the user receives file data, such as a CSV, Excel, etc., instead of raw binary data.

In the future, I propose to extend the `_FileDataNodeMixin` by adding a config property to specify the storage as follows: 

```python
class _FileDataNodeMixin(object):
    """Mixin class designed to handle file-based data nodes
    (CSVDataNode, ParquetDataNode, ExcelDataNode, PickleDataNode, JSONDataNode, etc.)."""

    __EXTENSION_MAP = {"csv": "csv", "excel": "xlsx", "parquet": "parquet", "pickle": "p", "json": "json"}

    _IS_GENERATED_KEY = "is_generated"
    _DEFAULT_DATA_KEY = "default_data"
    _PATH_KEY = "path"
    _DEFAULT_PATH_KEY = "default_path"

    _POSSIBLE_STORAGE = ["local", "google_cloud_storage", "aws_s3_object", "azure_blob_storage", "dropbox", "google_drive"]
    _DEFAULT_STORAGE = "local"
```

Taipy should be able to read/write/append/download/upload the data to/from the specified storage with the right storage_type ("csv", "xlsx", "pickle", etc.)

trgiangdo:
I agree with expanding the usage of `_FileDataNodeMixin`

jrobinAV:
Let me create an issue for that. Nothing is ready yet, ut we can use the issue to write precise requirements and specifications.

PR #1836, opened 2024-09-25
feature/#1805 added checking if entity exists

### LINKED ISSUE #1805: Protect the Reason APIs in case the entity or its ID does no exist
If the entity does not exist, the ReasonCollection should be False, and a specific Reason should be provided. 

Must be propagated to enterprise and documentation.

toan-quach:
#1805

### DIFF
--- a/taipy/core/_orchestrator/_orchestrator_factory.py
+++ b/taipy/core/_orchestrator/_orchestrator_factory.py
@@ -9,11 +9,11 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 import typing
-from importlib import util
 from typing import Optional, Type
 
 from taipy.config.config import Config
 
+from ..common._check_dependencies import _TAIPY_ENTERPRISE_MODULE, _using_enterprise
 from ..common._utils import _load_fct
 from ..exceptions.exceptions import ModeNotAvailable, OrchestratorNotBuilt
 from ._abstract_orchestrator import _AbstractOrchestrator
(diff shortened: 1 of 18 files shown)


taipy/core/job/_job_manager.py
```diff
@@ -92,9 +92,14 @@ def _is_deletable(cls, job: Union[Job, JobId]) -> ReasonCollection:
         reason_collector = ReasonCollection()
 
         if isinstance(job, str):
-            job = cls._get(job)
+            job_id = job
+            job = cls._get(job, None)
```
jrobinAV, line 96:
```suggestion
            if job:= cls._get(job, None) is None:
                reason_collector._add_reason(job_id, EntityDoesNotExist(job_id))
                return reason_collector
```

taipy/core/job/_job_manager.py
```diff
@@ -92,9 +92,14 @@ def _is_deletable(cls, job: Union[Job, JobId]) -> ReasonCollection:
...
+        else:
+            job_id = job.id
```
jrobinAV: I believe we don't need that.
```suggestion
```

taipy/core/job/_job_manager.py
```diff
@@ -92,9 +92,14 @@ def _is_deletable(cls, job: Union[Job, JobId]) -> ReasonCollection:
...
 
-        if job and not job.is_finished():
+        if job is None:
+            reason_collector._add_reason(job_id, EntityDoesNotExist(job_id))
+        elif not job.is_finished():
```
jrobinAV: From the previous comment, the case where the job can be None is already handled.
```suggestion
        if not job.is_finished():
```

jrobinAV (dismissed):
The logic is there but I just proposed a change in the way we handle conditions. What do you think?

jrobinAV (dismissed):
NEat! 
And the tests are great too.

jrobinAV:
Well the tests are failing.

taipy/core/_orchestrator/_orchestrator_factory.py
```diff
@@ -87,7 +87,13 @@ def __build_standalone_job_dispatcher(cls, force_restart=False):
     def __build_development_job_dispatcher(cls):
         if isinstance(cls._dispatcher, _StandaloneJobDispatcher):
             cls._dispatcher.stop()
-        cls._dispatcher = _DevelopmentJobDispatcher(typing.cast(_AbstractOrchestrator, cls._orchestrator))
+
+        if util.find_spec(cls._TAIPY_ENTERPRISE_MODULE) is not None:
```
jrobinAV: I guess you ll want to use the _using_enterprise function from common package.

jrobinAV (dismissed):

taipy/core/_orchestrator/_orchestrator_factory.py
toan-quach: ah yes, ohh I merged the other PR before this one 🤣

jrobinAV (approved):

PR #1837, opened 2024-09-25
Feature/#1704 - Store the status changes in a records

### LINKED ISSUE #1704: Storing the information about Job's status changes
### Description

Currently, when a job is executed, the start and finish time is logged.

It can be more useful if all the timestamps for all the job status changes are kept: SUBMITTED, BLOCKED, PENDING, RUNNING, CANCELED, FAILED, COMPLETED, SKIPPED, ABANDONED.

With that, we got all the information for the user to compute everything,
like the total execution time, of course, but also the pending time, the running time, the blocked time.

For the submission, all information can be calculated from the jobs.

### Solution Proposed

So we can use it like that:

```
jobs._records == 
    {
        "SUBMITTED": 2022-09-27 18:00:00.000, 
        "RUNNING": 2022-09-27 18:00:01.000, ..., 
        "COMPLETED": 2022-09-27 18:00:02.000
    }
jobs.execution_time == 2
```

We could add some dynamic properties like the following

```python
@property
def execution_time():
    if not "SUBMITTED" in self._records:
        return -1
    if "CANCELED" in self._records:
        return self._records["CANCELED"] - self._records["SUBMITTED"]
    elif "FAILED" in self._records:
        return self._records["FAILED"] - self._records["SUBMITTED"]
    elif "COMPLETED" in self._records:
        return self._records["COMPLETED"] - self._records["SUBMITTED"]
    elif "SKIPPED" in self._records:
        return self._records["SKIPPED"] - self._records["SUBMITTED"]
    elif "ABANDONED" in self._records:
        return self._records["ABANDONED"] - self._records["SUBMITTED"]
    else
        return datetime.now - self._records["SUBMITTED"]
```

### Impact of Solution

There should be no performance impact since a status change already triggers the job update.

### Additional Context

_No response_

### Acceptance Criteria

### Code of Conduct

trgiangdo:
Resolves #1704

All status changes of a job is now recorded in a dictionary.

### DIFF
--- a/taipy/core/job/job.py
+++ b/taipy/core/job/job.py
@@ -12,7 +12,7 @@
 __all__ = ["Job"]
 
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Callable, List, Optional
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional
 
 from taipy.logger._taipy_logger import _TaipyLogger
 
@@ -49,8 +49,8 @@ class Job(_Entity, _Labeled):
 
     Every time a task is submitted for execution, a new *Job* is created. A job represents a
     single execution of a task. It holds all the information related to the task execution,
-    including the **creation date**, the execution `Status^`, and the **stacktrace** of any
-    exception that may be raised by the user function.
+    including the **creation date**, the execution `Status^`, the timestamp of status changes,
+    and the **stacktrace** of any exception that may be raised by the user function.
 
     In addition, a job notifies scenario or sequence subscribers on its status change.
 
@@ -78,8 +78,7 @@ class Job(_Entity, _Labeled):
         self._creation_date = datetime.now()
         self._submit_id: str = submit_id
         self._submit_entity_id: str = submit_entity_id
-        self._execution_started_at: Optional[datetime] = None
-        self._execution_ended_at: Optional[datetime] = None
+        self._status_change_records: Dict[str, datetime] = {"SUBMITTED": self._creation_date}
         self._subscribers: List[Callable] = []
         self._stacktrace: List[str] = []
         self.__logger = _TaipyLogger._get_logger()
@@ -134,6 +133,7 @@ class Job(_Entity, _Labeled):
     @status.setter  # type: ignore
     @_self_setter(_MANAGER_NAME)
     def status(self, val):
+        self._status_change_records[val.name] = datetime.now()
         self._status = val
 
     @property  # type: ignore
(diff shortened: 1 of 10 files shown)


taipy/core/job/job.py
```diff
@@ -212,7 +212,7 @@ def execution_duration(self) -> Optional[float]:
 
     @property
     @_self_reload(_MANAGER_NAME)
-    def pending_time(self) -> Optional[float]:
+    def pending_duration(self) -> Optional[float]:
```
toan-quach, line 215: great naming!

toan-quach (approved):

toan-quach (dismissed):

taipy/core/job/job.py
```diff
@@ -148,36 +148,107 @@ def creation_date(self, val):
...
-            completed, None is returned.
+            Optional[float]: The duration of the job execution in seconds.
+                - If the job was not run, None is returned.
+                - If the job is not finished, the execution time is the duration
+                  from the running time to the current time.
         """
-        if self._execution_started_at and self._execution_ended_at:
-            return (self._execution_ended_at - self._execution_started_at).total_seconds()
-        return None
+        if "RUNNING" not in self._status_change_records:
```
toan-quach: can we replace hard-coded `"RUNNING"` with a constant instead?

taipy/core/job/job.py
```diff
@@ -148,36 +148,107 @@ def creation_date(self, val):
...
+        Returns:
+            Optional[float]: The duration of the job in the pending state in seconds.
+                - If the job is not running, None is returned.
+                - If the job is not pending, the pending time is the duration
+                  from the submission to the current time.
+        """
+        if "PENDING" not in self._status_change_records:
+            return None
+
+        if self.is_finished() or self.is_running():
```
toan-quach, line 227: hmmm this might cause alot of reload yeah?

toan-quach (dismissed):
only reviewed the code and not the test but lgtm

taipy/core/job/job.py
trgiangdo: Yes, good idea. We can do something like this
```suggestion
        if Status.RUNNING.name not in self._status_change_records:
```

Let me update it everywhere

taipy/core/job/job.py
trgiangdo: it's an attribute so I think it's reasonable

taipy/core/job/job.py
```diff
@@ -148,36 +148,111 @@ def creation_date(self, val):
 
     @property
     @_self_reload(_MANAGER_NAME)
-    def execution_started_at(self) -> Optional[datetime]:
-        return self._execution_started_at
+    def submitted_time(self) -> datetime:
```
jrobinAV: The wording time is confusing. Usually, the running time is the duration of a program run. HEre it is the instant. What about having APIs like: 

is_submitted -> bool: Returns True if the job has submitted status, false otherwise.
submitted() -> None: Sets the status to submitted and returns None.
submitted_at -> datetime: Returns the instant the job status changed to submitted.
submitted_time -> datetime: Returns the time spent on submitted status.

taipy/core/job/job.py
```diff
@@ -148,36 +148,111 @@ def creation_date(self, val):
...
+            return None
+
+        if self.is_finished():
+            return (self.finished_time - self._status_change_records[Status.BLOCKED.name]).total_seconds()
+        if self.is_running():
+            return (
+                self._status_change_records[Status.RUNNING.name] - self._status_change_records[Status.BLOCKED.name]
+            ).total_seconds()
+
+        return (datetime.now() - self._status_change_records[Status.BLOCKED.name]).total_seconds()
```
jrobinAV: Is it correct?

What if the job got blocked, unblocked(pending), running, and completed?
Your method returns the difference between the completion time and the blocked time. while we expect to return PENDING - BLOCKED.

Let me propose something like the following:
```suggestion
        if self.is_blocked():
            # Currently blocked
            return (datetime.now() - self._status_change_records[Status.BLOCKED.name]).total_seconds()
        if Status.PENDING.name in self._status_change_records:
            # has been pending at a point
            return (
                self._status_change_records[Status.PENDING.name] - self._status_change_records[Status.BLOCKED.name]
            ).total_seconds()
        if self.is_finished():
            # has never been pending but is finished
            return (self.finished_time - self._status_change_records[Status.BLOCKED.name]).total_seconds()
        # has never been pending, never been finished
        # I don't see any configuration where we should enter here. 
        # Am I correct?
        raise AppropriateError
```

jrobinAV (changes requested):

tests/core/_orchestrator/test_orchestrator__submit.py
```diff
@@ -562,19 +558,18 @@ def test_submit_duration_standalone_mode():
     scenario = Scenario("scenario", {task_1, task_2}, {})
     _ScenarioManager._set(scenario)
     submission = taipy.submit(scenario)
-    jobs = submission.jobs
 
-    assert_true_after_time(jobs[1].is_completed)
+    while not all(job is not None and job.is_completed() for job in submission.jobs):
+        sleep(1)  # Limit CPU usage
```
jrobinAV: Why not use the assert_true_after_time? It has been designed for that.

tests/core/_orchestrator/test_orchestrator__submit.py
```diff
@@ -535,17 +534,14 @@ def test_submit_duration_development_mode():
...
     assert jobs_1s.execution_duration >= 1
     assert jobs_2s.execution_duration >= 2
 
-    assert submission.execution_duration >= 3
-    assert submission.execution_started_at == min(jobs_1s.execution_started_at, jobs_2s.execution_started_at)
-    assert submission.execution_ended_at == max(jobs_1s.execution_ended_at, jobs_2s.execution_ended_at)
-
 
 @pytest.mark.standalone
 def test_submit_duration_standalone_mode():
```
jrobinAV, line 548: This test is not only about submit duration anylonger. It should be renamed.

trgiangdo: Welp, it will be, when the attributes of a submissions entity is added.

tests/core/_orchestrator/test_orchestrator__submit.py
trgiangdo: It's because the `assert_true_after_time` wants a function.
Here, we need to check for a clause `all()` only.

taipy/core/job/job.py
trgiangdo: I would rename this `submitted_time` --> `submitted_at`.
Similarly, `run_time` --> `run_at`.
           `finished_time` --> `finished_at`.

Then, I would keep the `execution_duration`, `pending_duration`, and `blocked_duration` naming.

What do you think?

taipy/core/job/job.py
trgiangdo: You're right. I missed the PENDING state. Let me update the logic as you suggested

tests/core/_orchestrator/test_orchestrator__submit.py
jrobinAV: As discussed, we can try with lambda functions. 
In any case, this is not blocking.

taipy/core/job/job.py
jrobinAV: Sounds good to me. execution_duration will prevent us from any confusion!
thx.

jrobinAV (approved):

toan-quach (approved):

PR #1839, opened 2024-09-25
Improve config doc and remove deprecated function.

jrobinAV:


taipy/config/config.py
```diff
@@ -74,43 +66,49 @@ def by_two(x: int):
...
-        - *Export the configuration*: Use the `Config.export()^` method to export as a TOML file
-            the Python code configuration.
-        - *Load the configuration*: Use the `Config.load()^` method to load a TOML configuration
-            file and replace the current Python configuration.
-        - *Override the configuration*: Use the `Config.override()^` method to load a TOML
-            configuration file and override the current Python configuration.
-
+        The Config class also provides a few methods to manage the configuration.
+
+        ??? example "Method to manage configuration"
```
trgiangdo:
```suggestion
        ??? example "Methods to manage the configuration"
```
or just
```suggestion
        ??? example "Manage the configuration"
```

taipy/config/config.py
```diff
@@ -74,43 +66,49 @@ def by_two(x: int):
...
+
+            - *Check the configuration for issues*: Use the `Config.check()^` method to check the
+                configuration. It returns an `IssueCollector^` containing all the `Issue^`s
+                found representing the issues with their severity.
```
trgiangdo:
```suggestion
                configuration. It returns an `IssueCollector^` containing all the
                 `Issue^`s found. The issues are logged to the console for debugging.
```
I don't get the final part of this sentence.
Maybe we can add "The issues are logged to the console for debugging".

taipy/config/config.py
```diff
@@ -127,17 +125,14 @@ def by_two(x: int):
 
     @_Classproperty
     def unique_sections(cls) -> Dict[str, UniqueSection]:
-        """Return all unique sections."""
```
trgiangdo, line 130: Why don't we want the docstring for the @_Classproperty anymore?

taipy/core/_core.py
```diff
@@ -16,7 +16,7 @@
 
 
 class Core:
-    """Deprecated. Use the `Orchestrator^` service class instead."""
+    """NOT DOCUMENTED"""
```
trgiangdo, line 19: The `Core` is not in the refman anymore?

trgiangdo (commented):
I have some questions

taipy/config/config.py
jrobinAV: That's a question we have to answer for the Reference manual generation of a class. 
Today, two sections are generated to list the attributes: one with the attributes declared in the class docstring and one with the attributes and properties that have a docstring in the class definition.
So, either we expose all attributes and properties in the class docstring or we remove the attribute section in the class docstring and add a docstring for all properties and all attributes. In the second case, we need to define the attributes outside the __init__ method.

Waiting for this to be discussed and decided in a global manner, I just made the quickest change not to have two sections with the same title and each one with part of the attributes.

taipy/core/_core.py
jrobinAV: Nop. We keep it for backward compatibility, but we don't want to expose it in the 4.0 version of the doc.

trgiangdo (approved):
LGTM

PR #1840, opened 2024-09-26
Feature - Add submission status change attributes

trgiangdo:
Split from #1837 

This PR adds the status timestamp attributes to the Submission entities.


jrobinAV (commented):
The naming should be aligned with the Job API.

jrobinAV (approved):

PR #1841, opened 2024-09-26
feature/enterprise/#398 load enterprise _TaskFunctionWrapper 

toan-quach:


jrobinAV (changes requested):
After discussion, I believe we don't need this.

jrobinAV (approved):
Neat !

PR #1842, opened 2024-09-26
Builder: Detect indexed property as callable

FredLL-Avaiga:
so that this works

```
from random import randrange

from taipy.gui import Gui

x_range = range(0, 11)
data = {"x": [randrange(10) for _ in x_range]}

with tgb.Page(
    style={
        ".col0": {"background-color": "red"},
        ".col1": {"background-color": "blue"},
        ".col2": {"background-color": "green"},
    }
) as page:
    tgb.table("{data}", style__x=lambda _, v: f"col{v}", show_all=True)

Gui(page=page).run(title="1832 indexed styles lambda")

```

### DIFF
--- a/taipy/gui/builder/_api_generator.py
+++ b/taipy/gui/builder/_api_generator.py
@@ -139,5 +139,9 @@ class _ElementApiGenerator(object, metaclass=_Singleton):
         return type(
             classname,
             (ElementBaseClass,),
-            {"_ELEMENT_NAME": element_name, "_DEFAULT_PROPERTY": default_property, "_TYPES": properties},
+            {
+                "_ELEMENT_NAME": element_name,
+                "_DEFAULT_PROPERTY": default_property,
+                "_TYPES": {f"{parts[0]}__" if len(parts := k.split("[")) > 1 else k: v for k, v in properties.items()},
+            },
         )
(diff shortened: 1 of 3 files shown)


taipy/gui/builder/_api_generator.py
```diff
@@ -139,5 +139,9 @@ def create_element_api(
         return type(
             classname,
             (ElementBaseClass,),
-            {"_ELEMENT_NAME": element_name, "_DEFAULT_PROPERTY": default_property, "_TYPES": properties},
+            {
+                "_ELEMENT_NAME": element_name,
+                "_DEFAULT_PROPERTY": default_property,
+                "_TYPES": {f"{parts[0]}__" if len(parts := k.split("[")) > 1 else k: v for k, v in properties.items()},
```
dinhlongviolin1, line 145: I dont understand why do we split by `[` here.

FredLL-Avaiga: So that we can access with a get the indexed property name and check if it's callable

taipy/gui/viselements.json
```diff
@@ -788,22 +788,22 @@
                     },
                     {
                         "name": "style[<i>column_name</i>]",
-                        "type": "str",
+                        "type": "Union[str, Callable]",
```
FabienLelaquais, line 791: I love that! I wanted to point this out!!

FabienLelaquais (approved):

dinhlongviolin1 (approved):

namnguyen20999 (approved):

PR #1847, opened 2024-09-27
Refactor

FredLL-Avaiga:


dinhlongviolin1 (approved):
Did you use any script to do the modification :D seems like a tiring one

FabienLelaquais (approved):
👌

FredLL-Avaiga:
> Did you use any script to do the modification :D seems like a tiring one

no, but it was not that bad

PR #1854, opened 2024-09-28
add sso hooks

dinhlongviolin1:
Support for sso


FabienLelaquais (dismissed):
I think I see this coming...

taipy/gui/gui.py
```diff
@@ -2609,6 +2609,8 @@ def __register_blueprint(self):
         # server URL Rule for flask rendered react-router
         pages_bp.add_url_rule(f"/{Gui.__INIT_URL}", view_func=self.__init_route)
 
+        _Hooks()._add_external_blueprint(self, __name__)
```
FredLL-Avaiga, line 2630: 😍

taipy/gui/_default_config.py
```diff
@@ -46,7 +46,7 @@
     "change_delay": None,
     "chart_dark_template": None,
     "base_url": "/",
-    "client_url": "http://localhost:5000",
+    "client_url": "http://127.0.0.1:5000",
```
FredLL-Avaiga: why not keep localhost everywhere ?

FredLL-Avaiga (approved):

FabienLelaquais (approved):
👌

taipy/gui/_default_config.py
dinhlongviolin1: I will want to keep the host as an Ip address for now. localhost will be used by the client_url due to the sso stuff only allowing localhost and not ip addresses in most sso protocols.

PR #1865, opened 2024-10-01
Bug/Fix failing issue on enterprise due to import manager

toan-quach:
As the initial purpose of this PR was different, the branch name is now deprecated and should be ignored. The title is correct as it has been fixed.

### DIFF
--- a/taipy/config/__init__.py
+++ b/taipy/config/__init__.py
@@ -103,3 +103,9 @@ def _inject_section(
 
     for exposed_configuration_method, configuration_method in configuration_methods:
         setattr(Config, exposed_configuration_method, configuration_method)
+
+
+@_config_doc
+def _inject_method(configuration_methods: List[tuple]):
+    for exposed_configuration_method, configuration_method in configuration_methods:
+        setattr(Config, exposed_configuration_method, configuration_method)
--- a/taipy/config/__init__.py
+++ b/taipy/config/__init__.py
@@ -103,9 +103,3 @@ def _inject_section(
 
     for exposed_configuration_method, configuration_method in configuration_methods:
         setattr(Config, exposed_configuration_method, configuration_method)
-
-
-@_config_doc
-def _inject_method(configuration_methods: List[tuple]):
-    for exposed_configuration_method, configuration_method in configuration_methods:
-        setattr(Config, exposed_configuration_method, configuration_method)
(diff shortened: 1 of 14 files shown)


taipy/config/__init__.py
```diff
@@ -105,3 +105,9 @@ def _inject_section(
 
     for exposed_configuration_method, configuration_method in configuration_methods:
         setattr(Config, exposed_configuration_method, configuration_method)
+
+
+@_config_doc
+def _inject_method(configuration_methods: List[tuple]):
+    for exposed_configuration_method, configuration_method in configuration_methods:
+        setattr(Config, exposed_configuration_method, configuration_method)
```
trgiangdo: I understand that it's for Databrick.

Can you give an example how can this method being used?

jrobinAV:
I turned the PR into a draft to avoid merging it by mistake before the 4.0 release

taipy/config/__init__.py
toan-quach: yep sure thing, so we want to add a `configure_databricks_task` method to the Config, only the method. But the inject_section config also expects the section and many other things which don't relate to what we want to do with a single method. And so i added the function :D

taipy/core/common/_check_dependencies.py
```diff
@@ -34,5 +31,10 @@ def _check_dependency_is_installed(module_name: str, package_name: str) -> None:
         )
 
 
-def _using_enterprise() -> bool:
-    return util.find_spec(_TAIPY_ENTERPRISE_MODULE) is not None
+class EnterpriseChecker:
```
jrobinAV: The official name is Enterprise Edition. Plus Checker is used in another context, to verify the consistency of config sections. I would propose the following names:
- EnterpriseEditionHelper
- EnterpriseEditionUtils
- EnterpriseEditionContext

What do you think?

jrobinAV (commented):
Is there any search for `taipy.enterprise` elsewhere in the code than in `core` and `config`? 

If so, I would move the new class EnterpriseCHecker (I would rename it BTW) to the new package `taipy.common` What do you think?

taipy/core/common/_check_dependencies.py
toan-quach: I will go with `EnterpriseEditionUtils` thanks! 😃

trgiangdo (approved):
LGTM

jrobinAV (approved):

Issue #555, opened 2022-12-12
Mapbox token management


FabienLelaquais:
**Description**

Mapbox is used by Plotly to create map charts.

Users may need to use a Mapbox Access token (see https://plotly.com/python/mapbox-layers/) is order to use the Mapbox service.
We need to find a way for users to provide Taipy with their specific Mapbox token.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Ensure any change is well documented

FredLL-Avaiga:
I guess via the Gui config and a property that defaults to the config value

BaunaBadmosh:
i would like to work on this

anupriyakkumari:
Hi, could this get assigned to me?

FlorianJacta:
Hi all, I will assign @PRIYANKjakharia as he is the first to ask. Thank you for your contribution.

@anupriyakkumari, many other issues are waiting for you!

ShootingStarD:
FYI the mapbox plotly API is deprecated and may be removed in future versions of plotly. I don't know if taipy already updated its code based on that : https://plotly.com/python/tile-map-layers/#mapbox

FredLL-Avaiga:
yes, we support the new mapbox

FredLL-Avaiga:
Should this be closed or Should we support the token while plotly still support mapbox ? @FabienLelaquais

FabienLelaquais:
We don't want to base ant code on deprecated libraries.

PR #1886, opened 2024-10-02
Added Mapbox token integration to Taipy

BaunaBadmosh:
This pull request updates the get_plotly_content function to use the get_mapbox_token function to retrieve the Mapbox token from the environment variable. The Mapbox token is then used to create a Plotly figure with a Mapbox map.
Issue #555 solved
The changes include:

Updated get_plotly_content function in taipy/gui/gui.py to use get_mapbox_token function
Added get_mapbox_token function to taipy/gui/utils/mapbox_token.py
Updated get_plotly_content function to use Mapbox token to create Plotly figure with Mapbox map
Note: The tests failed due to missing modules, which will need to be installed before the tests can be run successfully."
Thank You


taipy/gui/gui.py
```diff
@@ -10,7 +10,7 @@
 # specific language governing permissions and limitations under the License.
 
 from __future__ import annotations
-
+from taipy.gui.utils.mapbox_token import get_mapbox_token
```
FredLL-Avaiga: ruff will not agree with this

taipy/gui/gui.py
```diff
@@ -445,6 +445,10 @@ def __process_content_provider(self, state: State, path: str, query: t.Dict[str,
                     if isinstance(content, PlotlyFigure):
 
                         def get_plotly_content(figure: PlotlyFigure):
+                            mapbox_token = get_mapbox_token()
```
FredLL-Avaiga: I think we wanted to support map token *not only* for figure but also for plotly.js based charts

taipy/gui/utils/mapbox_token.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+def get_mapbox_token():
+    mapbox_token = os.environ.get('MAPBOX_TOKEN')
+    if mapbox_token is None:
+        raise ValueError("Mapbox Access token is not set as an environment variable")
```
FredLL-Avaiga: this should stay optional

FredLL-Avaiga (changes requested):

taipy/gui/gui.py
```diff
@@ -452,42 +451,58 @@ def __process_content_provider(self, state: State, path: str, query: t.Dict[str,
...
+                                import json
+
+                                fig = go.Figure(data=content['data'], layout=content['layout'])
+                                fig.update_layout(mapbox_style="mapbox://styles/mapbox/streets-v11",
+                                mapbox_center_lat=37.7749, mapbox_center_lon=-122.4194,
+                                mapbox_zoom=12, mapbox_access_token=mapbox_token)
+                                html = fig.to_html(include_plotlyjs='cdn')
+                                return html
+
+                        Gui.register_content_provider((PlotlyFigure, dict), get_plotly_content)
```
FredLL-Avaiga, line 475: that won't work register_content_provider doesn't take a tuple as parameter

FredLL-Avaiga (changes requested):

taipy/gui/gui.py
BaunaBadmosh: sorry for making such mistakes. Shall I register the content provider for each type separately?

FredLL-Avaiga: I'm afraid it is more complicated than that
some Map chart don"t use figure at all and are completely built on the frontend
there are not using the content provider mechanism.

FredLL-Avaiga: don't apologize for making mistakes, that's the best way to learn :-)
Thanks for your help

BaunaBadmosh: thank you for your kind words. what shall I do ahead

FredLL-Avaiga: I propose that you understand how the chart component handle plotly.js charts on the frontend and figure a way to provide the right token from the backend to the frontend

BaunaBadmosh: okay sir i am on it.

PR #1896, opened 2024-10-03
Possibility to duplicate scenarios Issue Solved

AnujSaha0111:
When creating a new scenario by duplicating an existing one, the code checks whether any Data Nodes (inputs) for the tasks have changed. If no changes are detected, it reuses the cached task output instead of re-running the task, improving performance. Finally, the new scenario is saved with the appropriate reused or re-executed tasks.


Data_Node_Changes_check.py
```diff
@@ -0,0 +1,17 @@
+class ScenarioManager:
```
trgiangdo, line 1: We already have a `_ScenarioManager` class

Scenario_Duplication_Method.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def duplicate_scenario(self, original_scenario_id):
+        # Fetch the original scenario by ID
+        original_scenario = self.get_scenario_by_id(original_scenario_id)
+
+        # Create a copy of the scenario structure
+        new_scenario = Scenario()
+        new_scenario.name = f"Duplicate of {original_scenario.name}"
+        
+        # Iterate over tasks in the original scenario
```
trgiangdo, line 11: There are a lot more properties in a Scenario than just `tasks`

We also need to duplicate those

Reusing_Cached_Task_Results.py
```diff
@@ -0,0 +1,13 @@
+class Scenario:
+
+    def add_task_with_cached_output(self, task):
+        # Instead of executing the task, add it with cached results
+        cached_result = self.get_cached_task_output(task.id)
+        self.tasks.append({
+            'task': task,
+            'output': cached_result
+        })
```
trgiangdo: This doesn't make any sense regarding the Taipy `Task` object. A scenario `tasks` is not just a list of dictionary with output

trgiangdo (changes requested):
Thank you for interested in contributing to Taipy.

Duplicating a scenario is a complicated topic, which requires duplicating every properties of a Scenario object, including children objects as well. Please take a look at the `_ScenarioManager` and `Scenario` classes in Taipy for more detail.

There are a lot of placeholder methods in your PR, and it's not working yet, so I believe this is not ready to be merged

Data_Node_Changes_check.py
AnujSaha0111: For this file, I needed a classs, so I added

Scenario_Duplication_Method.py
AnujSaha0111: I found this helpful

Reusing_Cached_Task_Results.py
AnujSaha0111: deleted it

PR #1900, opened 2024-10-03
chore: update test_dispatcher__execute_job.py

eltociear:
disptach -> dispatch


trgiangdo (approved):
We have an action that check for spelling error, but somehow failed to spot this error.

Thank you @eltociear for the fix.

jrobinAV (approved):

PR #1921, opened 2024-10-04
Update index.ts

Yash-2707:
solving Refresh value of selector list on first render


FredLL-Avaiga (changes requested):
What issue does this PR fix ?
Index shouldn't have any code.
You can not add code that is only used for testing

FabienLelaquais:
@Yash-2707
I guess the question is: "What problem are solving here?"
Thank you

PR #1928, opened 2024-10-04
Update CONTRIBUTING.md

kushal34712:
added contributors list to the contributing.md file


jrobinAV (commented):
The contributor list already appears in GitHub; We don't want to duplicate the information.

PR #1956, opened 2024-10-07
use taipy resource handler route to resolve cookies

dinhlongviolin1:


FredLL-Avaiga (dismissed):
Good start @dinhlongviolin1

FabienLelaquais (dismissed):
🤐

frontend/taipy-gui/base/src/cookieHandler.ts
```diff
@@ -0,0 +1,66 @@
...
+    async init(socket: Socket, taipyApp: TaipyApp) {
+        const hasValidCookies = await this.verifyCookieStatus();
+        if (!hasValidCookies) {
+            await this.deleteCookie();
+            localStorage.removeItem(TAIPY_RESOURCE_HANDLER);
+            window.location.reload();
+            return;
+        }
+        this.addBeforeUnloadListener();
+        initSocket(socket, taipyApp);
```
FredLL-Avaiga, line 23: Not sure about the init Socket inside the CookieHandler ...

frontend/taipy-gui/base/src/cookieHandler.ts
```diff
@@ -0,0 +1,66 @@
...
+import { TaipyApp } from "./app";
+import { initSocket } from "./socket";
+import axios from "axios";
+import { getLocalStorageValue } from "../../src/context/utils";
+
+export const TAIPY_RESOURCE_HANDLER = "tprh";
+
+declare global {
+    interface Window {
+        deleteCookie: () => void;
```
FredLL-Avaiga: is this declaration not available by default ?

frontend/taipy-gui/base/src/cookieHandler.ts
FredLL-Avaiga: especially if hasValidCookies is false
Then initSocket is not called at all !

frontend/taipy-gui/base/src/cookieHandler.ts
```diff
@@ -0,0 +1,66 @@
...
+        // If not, then some part of the code must have removed the cookie
+        // or wants to remove the cookie by removing the local storage
+        if (!this.resourceHandlerId) {
+            return new Promise((resolve) => resolve(false));
+        }
+        try {
+            // call to get cookie status
+            const { data } = await axios.get("taipy-rh");
+            // validate cookie status
+            if (!data || !data.rh_id || data.rh_id !== this.resourceHandlerId) {
```
FredLL-Avaiga: could this be if (data?.rh_id !== this.resourceHandlerId)  ?

frontend/taipy-gui/base/src/cookieHandler.ts
```diff
@@ -0,0 +1,66 @@
...
+            }
+        } catch (error) {
+            console.error("Error while validating cookie:", error);
+            return new Promise((resolve) => resolve(false));
+        }
+        return new Promise((resolve) => resolve(true));
+    }
+
+    addBeforeUnloadListener() {
+        window.deleteCookie = () => {
```
FredLL-Avaiga: why do you need to make it an window attribute ?

frontend/taipy-gui/base/src/cookieHandler.ts
```diff
@@ -0,0 +1,66 @@
...
+            axios.delete("taipy-rh");
+        };
+        window.addEventListener("beforeunload", () => {
+            localStorage.removeItem(TAIPY_RESOURCE_HANDLER);
+            window.deleteCookie();
```
FredLL-Avaiga: why not use `self.deleteCookie()`
with `const self = this;` before the addEventListener

FredLL-Avaiga (changes requested):
tehre's a case when initSocket will not be called

frontend/taipy-gui/base/src/cookieHandler.ts
dinhlongviolin1: we have to make this declaration as I will be using it during the before unload

frontend/taipy-gui/base/src/cookieHandler.ts
dinhlongviolin1: We dont need to call init socket in this case as the cookie is invalid. The key is that we will get out of this instance and go back to taipy-gui to get a new cookie. I put it here and do things asynchronously as I dont want to start the websocket connection until i am sure the cookie is valid.

frontend/taipy-gui/base/src/cookieHandler.ts
dinhlongviolin1: make sense to me thanks!

frontend/taipy-gui/base/src/cookieHandler.ts
dinhlongviolin1: this.deleteCookie().bind(this) should do the trick but I just want to make sure that the window does not have any dependency on the teardown function

FabienLelaquais (approved):
👍

frontend/taipy-gui/base/src/cookieHandler.ts
```diff
@@ -51,12 +45,9 @@ export class CookieHandler {
     }
 
     addBeforeUnloadListener() {
-        window.deleteCookie = () => {
-            axios.delete("taipy-rh");
-        };
         window.addEventListener("beforeunload", () => {
             localStorage.removeItem(TAIPY_RESOURCE_HANDLER);
-            window.deleteCookie();
+            this.deleteCookie();
```
FredLL-Avaiga, line 50: are you sure `this` is valid in this context ?
I would have saved it in the scope with `const self = this;`

FredLL-Avaiga (approved):

frontend/taipy-gui/base/src/cookieHandler.ts
FredLL-Avaiga: I think arrow function (method) are automatically bound ?

dinhlongviolin1: It is still working per my testing. I will try your method later on if it breaks somehow and I like your proposal very much too!

Issue #1592, opened 2024-07-26
Open ScenarioConfig API to export a scenario config graph as a picture.


jrobinAV:
The purpose of this issue is to let developers draw a scenario config DAG using matplotlib (or an alternative) and networkx to visualize the graph he just created, helping him understand if it is correct.

### Description

In the `ScenarioConfig` class (located at `taipy/core/config/scenario_config.py`), we want a new method named `draw()` that exports the scenario configuration graph (a DAG) as an image file.

The graph comprises the input `DataNodeConfigs`, the output `DataNodeConfigs`, and the `TaskConfigs` as nodes. 
There is a directed edge from a `DatanodeConfig` node to a `TaskConfig` node if and only if the `DatanodeConfig` is a `TaskConfig` input. 
A directed edge is from a `TaskConfig` node to a `DatanodeConfig` node if and only if the `DatanodeConfig` is a `TaskConfig` input.

### Implementation constraint.

If a new package is required (matplotlib?, marmaid?, etc.), it must be made optional. The `draw` function should check the presence of the package before using it. If the package is not installed, the `draw` function should log a warning message and return doing nothing else.

https://networkx.org/documentation/stable/reference/drawing.html

FabienLelaquais:
I certainly don't like this idea, involving both the import of another (large) package but also impacts the global product homogeneity, as Taipy GUI and Taipy Studio already provide ways of representing DAGs.
Why chose a technology that we don't already use?

HimangshuYadav:
@quest-bot loot #1932

HimangshuYadav:
@quest-bot embark

xyfer17:
@jrobinAV could you please assign this issue to me, I would like to take this up.

satyyam11:
@quest-bot loot #1592

jrobinAV:
Yep sure, you are assigned, @xyfer17 .

jrobinAV:
### Description

In the `ScenarioConfig` class, we want a new method named `draw()` that exports the scenario configuration graph as an image file.

The graph comprises the input `DataNodeConfigs`, the output `DataNodeConfigs`, and the `TaskConfigs` as nodes. There is a directed edge from a `DatanodeConfig` node to a `TaskConfig` node if and only if the `DatanodeConfig` is a `TaskConfig` input. There is a directed edge from a `TaskConfig` node to a `DatanodeConfig` node if and only if the `DatanodeConfig` is a `TaskConfig` input.

[...]

PR #1976, opened 2024-10-08
Add graph visualization feature

satyyam11:
This pull request addresses issue #1592 by introducing a new graph visualization feature to the project. 
The changes include a new graph_visualizer.py file with a function to visualize scenario configuration graphs using networkx and matplotlib. 
Additionally, the config/__init__.py file was updated to load the configuration and call the visualization function. This enhancement provides a visual representation of nodes and edges in the configuration, making it easier to understand and debug complex scenarios. 
Dependencies for networkx and matplotlib are required, which can be installed via pip. Please review and merge this PR to integrate the new feature into the project.


jrobinAV (changes requested):
The requirements are not satisfied.

In the ScenarioConfig class, we want a new method named draw() that exports the scenario configuration graph as an image file.

The graph comprises the input DataNodeConfigs, the output DataNodeConfigs, and the TaskConfigs as nodes. 
A directed edge exists from a `DatanodeConfig` node to a `TaskConfig` node if and only if the `DatanodeConfig` is a `TaskConfig`'s input. 
A directed edge exists from a `TaskConfig` node to a `DatanodeConfig` node if and only if the `DatanodeConfig` is a `TaskConfig` input.


If a new package is required, it must be made optional. The draw function should check the presence of the package before using it. If the package is not installed, the draw function should log a warning message and return doing nothing else.

taipy/common/config/ScenarioConfig.py
```diff
@@ -0,0 +1,46 @@
+import matplotlib.pyplot as plt
```
jrobinAV, line 1: This must be an optional dependency. Please verify it is installed before using it.

taipy/common/config/ScenarioConfig.py
```diff
@@ -0,0 +1,46 @@
...
+import networkx as nx
+
+
+class ScenarioConfig:
```
jrobinAV, line 5: We already have a Scenario Config class that already contains a DAG. It must be re-used.
The purpose is to add a `draw` method to this class.

jrobinAV (changes requested):
Several remarks: 
- The main code should not be committed.
- The `ScenarioConfig` class already exists. You should use it.
- The creation of the graph is not integrated with Taipy. The graph already exists as a set of task configs. It must be converted into a networkx DAG. A dedicated method could be exposed in the `ScenarioConfig` class for building the dag as a nx object.
- The feature should be generic. In particular, the size of the graph may vary a lot. 
- Many unit tests are expected in this PR. Several use cases must be tested with various graph sizes and shapes.
- No print is allowed. Please use a Taipy Logger.
- The pictures should not be committed.

jrobinAV:
@satyyam11 any news? Do my comments make sense to you?

Manideep-Maddileti:
i want to work on this issue please assig to me @jrobinAV

jrobinAV:
@Manideep-Maddileti This is a Pull request. Please feel free to contribute by reviewing the code, and add any comment.

If you want, you can also be assigned to the related issue #1592 and propose your own Pull Request.

jrobinAV:
@satyyam11 Any news ?

jrobinAV:
This PR is abandoned. The issue as been solved by another PR.

Issue #439, opened 2023-10-24
Improve input with phone number mask


FlorianJacta:
**Description**
The goal would be to add a phone number mask if needed.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
we could extend the mode property to support phone numbers

ioannastantzou:
hello! I am a first time contributor do you think I could be assigned for this issue?

FabienLelaquais:
Hi @ioannastantzou, and thank you for volunteering to make Taipy better. 
We wanted to have some more time to think about how to handle this issue.

There are several problems to handle for that feature:
- There is no such thing as a generic 'phone number' mask.
- I see no reason why only phone number input would be guided. That should apply to anything that has a format (social security number, zipcode...)
- Because the feature really shines on the front-end (with an instant validation of the input), this requires an implementation in JavaScript. However, the mask itself (or a validation callback function) must be defined on the back-end in Python.
  We need to come up with a way to inter-operate easily.
  
If you have any thoughts on these rabbit holes, please let us know. I'm afraid there's nothing that can be done until these are solved. That impacts both requirements and implementation considerations.

Manav958:
Hi, @FabienLelaquais , @jrobinAV can we use Geolocation in the frontend to apply specific mask for a person
or we can let the user decide the country and then mask it?

If you feel its a good approach can you assign me this?
i have some ideas for other problems also

FredLL-Avaiga:
> Hi, @FabienLelaquais , @jrobinAV can we use Geolocation in the frontend to apply specific mask for a person
> or we can let the user decide the country and then mask it?
> 
> If you feel its a good approach can you assign me this?
> i have some ideas for other problems also
> 
> 
> 

I guess you can use the browser locale specific functions

Manav958:
Yes True we can use Navigator function i guess
Can i give it a shot?
Please Tell me the full requirements 
and if you find me suitable please assign it to me.

ahmadnadeem6236:
I want to work on this issue. Please assign to me and provide the further details.
Thank You!

jrobinAV:
@Manav958 You are assigned. Thank you.

Manav958:
Thankyou  @jrobinAV for assigning me.

just a few question for starting

[...]

PR #1984, opened 2024-10-09
issue #439 - added a script for detecting the country of the phone number 

Manav958:
I have added the feature to detect the country from the given phone numbers
using pycountry and phonenumbers i have tested it with some numbers from various countries it works good.
Please accept the PR or suggest the required changes.
Thankyou

To Run it :-
configure the GUI part and run the python script named number_mask.py

** I had to upgrade the version of networkx from exactly 2.6 to >= 2.6


Pipfile
```diff
@@ -18,7 +18,7 @@ gitignore-parser = "==0.1.1"
 kthread = "==0.2.3"
 markdown = "==3.4.4"
 marshmallow = "==3.20.1"
-networkx = "==2.6"
+networkx = ">=2.6"
```
FredLL-Avaiga, line 21: this should be discussed by the team

frontend/taipy-gui/package.json
FredLL-Avaiga, line 1: You  should have to modify this file

frontend/taipy/package.json
FredLL-Avaiga, line 1: you shouldn't commit this file

taipy/_entrypoint.py
FredLL-Avaiga, line 1: no need to commit this file

Pipfile
Manav958: Ok
I did it because it was producing error while installing packages.

FredLL-Avaiga (changes requested):
You are not proposing any change to the code base.
this issue should have an impact on Input.tsx

Pipfile
FredLL-Avaiga: then raise an issue

frontend/taipy-gui/package.json
Manav958: Alright , I am doing the necessary changes.

Pipfile
Manav958: Hi a Quick question
should i modify Input.tsx to have a field for phone number and then trigger the py script for validation
cause i currently dont see that

please guide me through this . 
if i am wrong please tell exactly what is required.
Thankyou

** pardon me for asking a lot of questions as this is my first time contributing to a big organization.

FredLL-Avaiga: the aim of the game is indeed to modify Input.tsx to support masks (phone number but maybe something else ...).
Once your modification is done and you've wrote frontend tests in Input.spec.tsx, you build the bundle with `npm run build:dev`
then you create a script that will allow you to test the new features you've added
then you create a few examples in doc/gui/examples/

FredLL-Avaiga: PS: And you should update your branch

Manav958: Hi ,
I had to create a backend URL in order to expose the function of phone number masking
where should i keep that file current its in the root.

also i have updated the Input.tsx file's props for calling the backend

FredLL-Avaiga: I would really prefer a frontend only solution

Manav958: Oh Okay 
I will do that

Manav958: Hi sir
Sorry for delayed reply i fell sick

i have a few questions
I implemented the logic in front but not sure how to test that
The logic takes the phone number and slice the last four digits and mask the remaining ones as "*" 

Could help me? Understanding what steps must be taken

Manav958: Also while running npm run build:dev it throws error

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

FredLL-Avaiga: we use `NODE_OPTIONS: --max-old-space-size=4096`

FredLL-Avaiga:
Are you still working on this PR @Manav958  ?
If you are, please resolve the conflicts, update your branch and answer the questions
If you are not, please close the PR

Issue #1553, opened 2024-07-18
[🐛 BUG] Stylekit is not applied for text with mode="md" 


FlorianJacta:
### What went wrong? 🤔

Some styles in the Stylekit (through the class_name) are not applied when mode="md" is set for text visual element. This prevents the user from changing the text element like he wants.

The issue comes from a user: https://github.com/Avaiga/taipy/discussions/1547

### Expected Behavior

The class_name/stylekit should be applied for all styles even with mode="md".

### Steps to Reproduce Issue

Style don't work:

```python
import taipy.gui.builder as tgb 
from taipy import Gui 

x = {"a": {"b": {"": 1}}}

with tgb.Page() as page:
    with tgb.layout('5*1', gap='2.0rem'):
        card_style = 'card pb0 m0'
        heading_style = 'text-right h5 color-primary pb0'
        number_style = 'text-right h1 pb0'
        with tgb.part(card_style):
            tgb.text("**Headline**", class_name=heading_style, mode="md")
            tgb.text("{x['a']['b']['']}", format='%.0f', class_name=number_style, "mode="md")

Gui(page).run()
```
![image](https://github.com/user-attachments/assets/c8443d75-cc27-4e64-aee7-03dbce770f6e)

Works (but without the Markdown format:
```python
import taipy.gui.builder as tgb 
from taipy import Gui 

x = {"a": {"b": {"": 1}}}

with tgb.Page() as page:
    with tgb.layout('5*1', gap='2.0rem'):
        card_style = 'card pb0 m0'
        heading_style = 'text-right h5 color-primary pb0'
        number_style = 'text-right h1 pb0'
        with tgb.part(card_style):
            tgb.text("**Headline**", class_name=heading_style)
            tgb.text("{x['a']['b']['']}", format='%.0f', class_name=number_style)

Gui(page).run()
```
![image](https://github.com/user-attachments/assets/2ea37083-1ca5-4e05-81c6-2f0bd0b7eb96)

### Solution Proposed

This seems to come from a *<p>* tag surrounding the text and not being a *span* like the normal visual element.

### Version of Taipy

3.1.1 / Develop

### Acceptance Criteria

- [ ] Ensure new code is unit tested, and check code coverage is at least 90%.
- [ ] Create related issue in taipy-doc for documentation and Release Notes.

### Code of Conduct

- [X] I have checked the [existing issues](https://github.com/Avaiga/taipy/issues?q=is%3Aissue+).
[...]

PR #1872, opened 2024-10-01
Component level Styling

FredLL-Avaiga:
I didn't find an issue
the doc or viselements.json is *not* modified

```
from taipy.gui import Gui

x_range = range(-10, 11, 4)

data = {"x": x_range, "y": [x * x for x in x_range]}

table_style = {"td": {"color": "green !important", "background-color": "silver !important"}, "tr": {"toto": "tata"}}
button_style = {"color": "green", "background-color": "silver"}

page = """<|{data}|table|show_all|style={table_style}|style[a]=tata|>

<|Hello Button|button|style={button_style}|>
"""

if __name__ == "__main__":
    Gui(page).run(title="Styling component")

```

```
import taipy.gui.builder as tgb

from taipy.gui import Gui

x_range = range(-10, 11, 4)

data = {"x": x_range, "y": [x * x for x in x_range]}

with tgb.Page(frame=None) as page:
        tgb.button("label", style={"td": { "color": "blue"  }})

if __name__ == "__main__":
    Gui(page).run(title="Styling component")


```

PR #1990, opened 2024-10-09
Fix Md mode styling

Satoshi-Sh:
## Related Issue
#1553

## Problem 
Md mode uses the react-markdown library to render markup language. 
It creates a different structure like
```
<div className=className>
    <p>value</p>
</div>
```
 This is why some styling doesn't work in non-md<div><span className></span></div> mode. It creates
```
<div>
  <span className=className>
  value
 </span>
</div>
``` 

## Solution
The library has mapping attribute. [source Appendix B: Components](https://github.com/remarkjs/react-markdown?tab=readme-ov-file#appendix-b-components:~:text=again%20contains%20markdown!-,Appendix%20B%3A%20Components,-You%20can%20also)
 
I used it to make the same structure for md-mode. 

##   Screenshot
Screenshot from 2024-10-09 09-00-37

### DIFF
--- a/frontend/taipy-gui/src/components/Taipy/Field.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Field.tsx
@@ -14,6 +14,7 @@
 import React, { lazy, useMemo } from "react";
 import Typography from "@mui/material/Typography";
 import Tooltip from "@mui/material/Tooltip";
+import { Components } from "react-markdown";
 
 import { formatWSValue } from "../../utils";
 import { useClassNames, useDynamicProperty, useFormatConfig } from "../../utils/hooks";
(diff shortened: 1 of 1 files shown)


Satoshi-Sh:
I have a question.

When I ran Field.spec.tsx, all tests passed. However, I think the test unit below should fail.

```
it("can render markdown", async () => {
        render(<Field value="titi" className="taipy-text" mode="md" />);
        const elt = document.querySelector(".taipy-text");
        await waitFor(() => expect(elt?.querySelector("p")).not.toBeNull());
    });
```

Satoshi-Sh:
The issue might be fixed on #1872 ?

frontend/taipy-gui/src/components/Taipy/Field.tsx
```diff
@@ -39,6 +40,29 @@ const Field = (props: TaipyFieldProps) => {
     const formatConfig = useFormatConfig();
 
     const className = useClassNames(props.libClassName, props.dynamicClassName, props.className);
+    const markdownComponent: Components = {
```
FredLL-Avaiga, line 43: interesting

FredLL-Avaiga (changes requested):
I don't think this answers the requirements of the issue.
please re-read the issue description and ask questions to @FlorianJacta if you need more information.

Satoshi-Sh:
I think the different structures of HTML(md and non-md) cause the issue. 

Please let me know if you have any advice on this PR. @FlorianJacta

frontend/taipy-gui/src/components/Taipy/Field.tsx
Satoshi-Sh: I reread the description and I thought this approach fixed the issue mentioned.  Adjusting the HTML structure is not a good idea here. @FredLL-Avaiga

FredLL-Avaiga: I think you're right, we don't want to mess with the HTML rendered by markdown

Satoshi-Sh: Is there any better way fixing the styling issue?  Md mode adds className to the parent div instead of adding it to the text element. That is why the element spacing doesn't work as non-md mode.

FredLL-Avaiga: I cannot think of an easy way or even a way that would fit all use cases ...

Satoshi-Sh: I see. Is it better to close this PR and leave the issue open/blocked for now as there is no reasonable solution?

Satoshi-Sh:
There seems to be no reasonable solution as discussed in the comments.

Issue #1968, opened 2024-10-08
[DOCS] Consistency in function and method docstrings


jrobinAV:
### Issue Description

Function and method parameters are documented using either `Parameters`, `Arguments`, or `Args` keywords. We should only use `Arguments` everywhere.

### Code of Conduct

- [x] I have checked the [existing issues](https://github.com/Avaiga/taipy/issues?q=is%3Aissue+).
- [ ] I am willing to work on this issue (optional)

Tushar98644:
@jrobinAV can i take up the issue ?

wizaye:
@jrobinAV would like to work on this , could you assign this to me?

FlorianJacta:
@Tushar98644 @wizaye I have assigned this issue to both of you! Thank you for your contribution!

wizaye:
@FlorianJacta I've made a PR with the necessary changes. please check it ;)

Harish-2003:
Is this issue still open

wizaye:
@Harish-2003 its open but we're working on it kindly check other issues for time being

Harish-2003:
> @Harish-2003 its open but we're working on it kindly check other issues for time being

Sure

wizaye:
@trgiangdo @FlorianJacta i had a issue with the branch while updating the previous pr, i've opened a new one #1973 with the requested changes, please look into it

afrinashar:
I would like to work on this.

FlorianJacta:
@wizaye is already working on it and created a PR on this issue. But I can also assign you if you want to collaborate.

margaret-hu:
How do I take up an issue?

wizaye:
@margaret-hu  someone from the team assigns it to you and then you work on it and make a valid PR

margaret-hu:
Can I assist with this issue as well?

wizaye:
@margaret-hu yes you can , wait until someone from the team assigns the issue to you. We're currently working on the issue with an ongoing PR you can check that as well.

jrobinAV:
@margaret-hu Thank you for your help. You are assigned as well.

margaret-hu:
How do I directly contribute to an issue?

DevyManish:
i do like to take on this issue please assign it to me!

Sushil2k4:
I would like to work on this project. Please assign it to me .

FlorianJacta:
[...]

PR #1998, opened 2024-10-09
[DOCS] Consistency in function and method docstrings

afrinashar:
#1968 Used Arguments everywhere in Docs.


jrobinAV:
That's not what we want. The requirement is to use the Arguments keyword instead of Parameters keyword in Python docstrings for methods and functions.

This PR does not address the issue.

PR #2013, opened 2024-10-10
This is a class to export graph to png image.

HimangshuYadav:
This is a class to export graph to png image...
It is from reference to [issue](https://github.com/Avaiga/taipy/issues/1592/)

### DIFF
--- a/taipy/core/config/graph_to_pic.py
+++ b/taipy/core/config/graph_to_pic.py
@@ -1,0 +1,46 @@
+import matplotlib.pyplot as plt
+import networkx as nx
+
+class ScenarioConfig:
+    def __init__(self):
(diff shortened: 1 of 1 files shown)


joaoandre-avaiga:
Hey @HimangshuYadav ! Thank you for your interest in this issue, however, this is not what we need. [`ScenarioConfig`](https://github.com/Avaiga/taipy/blob/develop/taipy/core/config/scenario_config.py) is already a Taipy Class, no need to create a new one. The idea is to add the `draw` method to the existing class where we have a list of attributes that repesents the various nodes and edges of the DAG. Please refer to the original instruction from the issue you mentioned in the PR and adjust the implementation accordingly. In the meantime, this PR will be closed, feel free to submit a new one when you are ready.

PR #2020, opened 2024-10-11
Draft: Adding Testing Documentation for Taipy GUI

gShahr:


FabienLelaquais:
I don't see any added value to this document. Taipy GUI is not even involved.
ChatGPT can tell me what playwright is... so what?

PR #2026, opened 2024-10-11
fix cli collision between host and client_url 

dinhlongviolin1:
Resolve Avaiga/taipy-enterprise#498


FabienLelaquais (approved):
Our collective bad

FredLL-Avaiga (approved):

dinhlongviolin1:
> Our collective bad

too many changes so it is hard to detect the mistake :D

jrobinAV:
Good catch. That is the opportunity to add some gui CLI command tests. We already have some for other commands in `tests/core/test_core_cli.py` that can be used as examples.

PR #2031, opened 2024-10-12
fix cli collision between host and client_url (#2026)

dinhlongviolin1:
Resolve https://github.com/Avaiga/taipy-enterprise/issues/498


FabienLelaquais (approved):
👍

FredLL-Avaiga (approved):
Good catch

Issue #1452, opened 2024-06-26
Add a validation for date with_time


carza06:
### What would you like to share or ask?

I need to validate a date with_time input created with mardown so that it shows me the minutes by 15, is it possible?

### Code of Conduct

- [X] I have checked the [existing issues](https://github.com/Avaiga/taipy/issues?q=is%3Aissue+).
- [ ] I am willing to work on this issue (optional)

FlorianJacta:
So, do you want 2 PM 00, 2 PM 15, 2 PM 30, ...? This is not possible right now.

This issue can bring this new feature to Taipy; we don't know yet if it will be present. Let's discuss it

FlorianJacta:
Would you say that this is related to https://github.com/Avaiga/taipy/issues/848?

carza06:
It's similar but with time

carza06:
Another question is, is there something similar to action_keys but for date

SandeepaDilshanAlagiyawanna:
@jrobinAV Can I solve this ? As a Hacktoberfest participant ?

jrobinAV:
@SandeepaDilshanAlagiyawanna Sure!

Thank you for your contribution.

Aryam2121:
please assign to me

FlorianJacta:
You can also work on it @Aryam2121

SandeepaDilshanAlagiyawanna:
I have fixed this in: 0d845e4c5a16fbcfa08a13039b27a9e4b5117426 

check this out @jrobinAV @FlorianJacta

Aryam2121:
I also fixed this please check it out @jrobinAV @FlorianJacta

Tryxns:
I comment to join the queue after Aryam & Sandeepa =)
Feel free to assign me @jrobinAV  if there's no active assignee for this issue

Aryam2121:
i am active and i completed all work

Tryxns:
@FlorianJacta @jrobinAV :eyes: 
I'm willing to save this issue from being wasted in hacktober event
FYI: 3 days left

Tryxns:
@quest-bot embark

kris2005mehra:
@jrobinAV can  you assign this task to me as I am a Hacktoberfest participant

Tryxns:
@quest-bot embark

SandeepaDilshanAlagiyawanna:
@quest-bot embark

kris2005mehra:
HTML Code
```
[...]

PR #2034, opened 2024-10-12
Update DateRange.tsx

Aryam2121:
Rounding Function: The roundToNearest15Minutes function rounds the selected time to the nearest 15 minutes.
Handle Change Logic: The handleChange function has been modified to use the rounding function before dispatching the updated date values.
Complete Component: The full DateRange component is included for context, with both date and datetime pickers supported.so;ve issue #1452

### DIFF
--- a/frontend/taipy-gui/src/components/Taipy/DateRange.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/DateRange.tsx
@@ -6,8 +6,8 @@
  *
  *        http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an
- * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
  * specific language governing permissions and limitations under the License.
  */
 
@@ -18,7 +18,7 @@ import Typography from "@mui/material/Typography";
 import { DatePicker, DatePickerProps } from "@mui/x-date-pickers/DatePicker";
 import { BaseDateTimePickerSlotProps } from "@mui/x-date-pickers/DateTimePicker/shared";
 import { DateTimePicker, DateTimePickerProps } from "@mui/x-date-pickers/DateTimePicker";
-import { isValid } from "date-fns";
+import { isValid, setMinutes, setHours } from "date-fns";
 import { ErrorBoundary } from "react-error-boundary";
 
 import { createSendUpdateAction } from "../../context/taipyReducers";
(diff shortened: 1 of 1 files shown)


frontend/taipy-gui/src/components/Taipy/DateRange.tsx
```diff
@@ -6,8 +6,8 @@
  *
  *        http://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
- * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
```
FredLL-Avaiga: please do not modify head comment as it is generated and eventually replaced automatically.

frontend/taipy-gui/src/components/Taipy/DateRange.tsx
```diff
@@ -54,7 +62,7 @@ const getRangeDateTime = (
         try {
             dates = JSON.parse(json);
         } catch {
-            // too bad
+            // Handle error if JSON parsing fails
```
FredLL-Avaiga: we don't want to handle this error

frontend/taipy-gui/src/components/Taipy/DateRange.tsx
```diff
@@ -86,7 +94,7 @@ const DateRange = (props: DateRangeProps) => {
         (v: Date | null, start: boolean) => {
             setValue((dates) => {
                 if (v !== null && isValid(v)) {
-                    const newDate = getTimeZonedDate(v, tz, withTime);
+                    const newDate = roundToNearest15Minutes(getTimeZonedDate(v, tz, withTime)); // Round to nearest 15 mins
```
FredLL-Avaiga: this behavior should be optional and commanded by a property.
I propose that this property would be set the value in time format (hh:mm:ss) so that we can define before showing the time selector, which times the user can select
I propose the property name: time_step
In this example, `time_step=00:15` 
or `time_step=01:00` for letting the user select only time on the hour.
This should also apply to date_selector and time_selector

frontend/taipy-gui/src/components/Taipy/DateRange.tsx
```diff
@@ -218,20 +226,23 @@ const DateRange = (props: DateRangeProps) => {
                                 format={props.format}
                                 id={id && id + "-field"}
                                 className={getSuffixedClassNames(className, "-text")}
-                                width={props.width && "100%"}
+                                width={props.width}
+                                disabled
+                                label={props.labelStart}
```
FredLL-Avaiga: I don't understand those changes.

- Why change the width ?
- Field has no `disabled` or `label` property

FredLL-Avaiga (changes requested):
Good work @Aryam2121 
I think this issue needs a bit more thinking before implementation.
In the code you propose, every component is impacted by the stepping to 15 minutes.
I think this feature should be optional and configurable and should be applied before user selection to limit the available choices.
It should also apply to date_selector and time_selector.
If you provide code, you need also to provide tests.

FredLL-Avaiga:
Are you still working on this PR @Aryam2121  ?
If you are, please resolve the conflicts, update your branch and answer the questions
If you are not, please close the PR

Aryam2121:
i update my Branch  already as your requirements you have to check the updated branch i want you review my updated branch

FredLL-Avaiga:
> i update my Branch already as your requirements you have to check the updated branch i want you review my updated branch

I'll review it once you have fixed the conflicts

Aryam2121:
so you merge this please i already fixed all conflicts

FredLL-Avaiga:
> so you merge this please i already fixed all conflicts

image

FabienLelaquais:
Does not do much - and certainly not enough.

PR #2043, opened 2024-10-13
chore: update table_guard_edits.py

eltociear:
minor fix


doc/gui/examples/controls/table_guard_edits.py
```diff
@@ -88,7 +88,7 @@ def check_add(state: State, var_name: str, payload: dict):
 def force_salary(state: State, var_name: str, payload: dict):
     # Get the salary proposal from the callback's payload
     proposed_salary = payload["value"]
-    # Round it the the nearest multiple of 5000
+    # Round it the nearest multiple of 5000
```
JeevaRamanathan, line 91: I think `Round it to the ...` would be more appropriate.

Issue #412, opened 2023-10-09
New JobViewer visual element


jrobinAV:
## Objective
This issue aims to create a new visual element named `job`. The motivation is to have a specific visual element to display information related to a specific `Job`. 

## Requirements
The control should have content very similar to the detail panel of a job entry in the [job_selector](https://docs.taipy.io/en/develop/refmans/gui/viselements/corelements/job_selector/) visual element. The same code should be shared by the two visual elements as much as possible.

The `job` control should accept a [Job](https://docs.taipy.io/en/develop/refmans/reference/taipy.core.Job/) as the main value.

The `job` visual element should always display the following fields. A property should control whether the field is displayed or not.
- The job id - always displayed
- The creation date - always displayed
- The status - always displayed
- The submit id - NOT displayed by default
- The task simple label - displayed by default
- The submit_entity label - displayed by default
- The execution duration - displayed by default
- The pending duration - NOT displayed by default
- The blocked duration - NOT displayed by default
- The submission duration (finished_at - creation_date ) - displayed by default
- The stack trace - displayed by default

The job visual element should provide actions to:
- Cancel the job - displayed by default - active only if the status is pending or submitted or blocked 
- Delete the job - displayed by default - active only if the job is finished

## Expected output
- The implementation of the new control/visual element
- A default style and UX consistent with Taipy styling, in particular, the detail panel of the job_selector
- Unit tests
- Any suggestions to improve the UX, the content of the visual element, or its technical implementation are welcome.

FredLL-Avaiga:
is this a duplicate of #1251 ?
@jrobinAV

jrobinAV:
Not really.
I changed the description so it is more clear that this one is about a dedicated job viewer control, while the other is about a detail panel on the job selector.

Prathamesh010:
Requirement document is not accessible. @jrobinAV

jrobinAV:
@Prathamesh010 Sorry, the document is private. Instead, I copy pasted the part that concerns the issue in the description directly.

I would be pleased to provide more input if you have any questions.
For technical considerations, @FredLL-Avaiga and @FabienLelaquais are the right persons who can help.

Nupoor10:
Hi, can you please assign this issue to me, would love to work on this! @jrobinAV

FlorianJacta:
Here you go @Nupoor10 !

Rishi-0007:
@quest-bot embark

Rishi-0007:
[...]

PR #2051, opened 2024-10-14
Add Job Viewer visual element to display job details

Nupoor10:
This PR is regarding issue #412 

It includes addition of a new job viewer visual element that is similar to the details panel of the Job Selector visual element
1. Added a new React component JobDetailsViewer
2. Displayed all necessary information outlined in the issue with the correct conditions
3. Added a new method to fetch complete job details
4. Added a new method to check if a job can be cancelled
5. Modified the _GuiCoreLib.py and viselements.json files appropriately


Nupoor10:
@quest-bot loot #412

Nupoor10:
I had a few questions regarding this PR.
1. In which subfolder should I add the test file for the JobDetailsViewer component?
2. I tried to implement this component but I get a job_viewer not registered message so how can I implement it?
3. I have implemented the is_cancellable method in job.py folder that returns a boolean value, is that the expected implementation or should it return a ReasonCollection?
4. In the _GuiCoreLib.py file I Have used the __JOB_SELECTOR_ERROR_VAR variable so is it correct?

Additionally if there are any changes do let me know.

frontend/taipy/src/JobDetailsViewer.tsx
FredLL-Avaiga, line 1: Can you not reuse the JobViewer component ?

FredLL-Avaiga (changes requested):
re-use :-)

FredLL-Avaiga:
> 1. In which subfolder should I add the test file for the JobDetailsViewer component?

in the same folder, the file name should end with spec.tsx (check `ScenarioSelector.spec.tsx`)

> 2. I tried to implement this component but I get a job_viewer not registered message so how can I implement it?

frontend component name is calculated from backend name if not provided
you need to add the `react_component` parameter to your `Element` instantiation

> 3. I have implemented the is_cancellable method in job.py folder that returns a boolean value, is that the expected 
implementation or should it return a ReasonCollection?

a `ReasonCollection` would be nicer and more consistent

> 4. In the _GuiCoreLib.py file I Have used the __JOB_SELECTOR_ERROR_VAR variable so is it correct?

No, you need to define a specific error variable name for this new component

PS: please fix the linter errors and update your branch

FredLL-Avaiga:
Are you still working on this PR @Nupoor10 ?

FredLL-Avaiga:
> Are you still working on this PR @Nupoor10 ?

If you're not, please close this PR

PR #2058, opened 2024-10-15
Added a validation for date with_time based on #1452

SandeepaDilshanAlagiyawanna:


FredLL-Avaiga (changes requested):
This issue is a gui issue: it means it has to be solved by adding code to the frontend (and possibly the backend) of taipy-gui.

This is not what you're doing here

PR #2063, opened 2024-10-15
remove unwanted datanode filters

### LINKED ISSUE #2052: Remove input intermediate and output filters on data_node_selector
The input, intermediate, and output filters for the data node selector should be removed. 

https://github.com/Avaiga/taipy/blob/a7f0a4151af307c4201427a0696ccd758c479ea0/taipy/gui_core/_adapters.py#L554

These 3 filters do not make sense globally, as a data node cannot be intrinsically considered an input, intermediate, or output. It is not an end-user concept. On the contrary, it is a developer concept that depends on the DAG it refers to. Plus, a data node can belong to multiple dags (multiple scenarios, scenarios from different scenario configs, multiple sequences, global) and be an input in one context and an output in another.

As per the discussion in issue #1363, it has been decided to propose another solution.

FredLL-Avaiga:
resolves #2052

Removes input, output, intermediate filter declaration

### DIFF
--- a/taipy/common/config/config.pyi
+++ b/taipy/common/config/config.pyi
@@ -847,7 +847,9 @@ class Config:
         aws_s3_bucket_name: str,
         aws_s3_object_key: str,
         aws_region: Optional[str] = None,
-        aws_s3_object_parameters: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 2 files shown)


dinhlongviolin1 (approved):

FabienLelaquais (approved):
😮

PR #2072, opened 2024-10-16
Precompute the rank of a data node in all the scenario config DAGs.

### LINKED ISSUE #2071: Rank data nodes by topological order in the data node selector
> A tradeoff has been proposed.
> The idea is to display in the data node selector the data nodes in a topological order for a scenario. With this proposal, the scenario is the context used to set the data node rank.
> 
> A more formal proposal should come soon. 

 _Originally posted by @jrobinAV in [#1363](https://github.com/Avaiga/taipy/issues/1363#issuecomment-2312412495)_

jrobinAV:
resolves #2071 
- Each data node configuration now has a `_rank` attribute as a dictionary.
- When configuring a new scenario configuration, all the nested data node configurations are updated. An entry is added in the `DataNodeConfig._rank` dictionary. The key is the scenario config ID, and the value is the rank of the data node in the scenario config DAG.
- Each data node exposes a hidden API to retrieve its `_rank` for a given scenario configuration.


taipy/core/config/scenario_config.py
```diff
@@ -86,6 +88,35 @@ def __init__(
...
+        for task_cfg in self._tasks:
+            if has_inputs := task_cfg.inputs:
+                for predecessor in task_cfg.inputs:
+                    dag.add_edges_from([(predecessor, task_cfg)])
+            if has_outputs := task_cfg.outputs:
+                for successor in task_cfg.outputs:
+                    dag.add_edges_from([(task_cfg, successor)])
+            if not has_inputs and not has_outputs:
+                dag.add_node(task_cfg)
+        # Remove tasks wit no input
```
jrobinAV:
```suggestion
        # Remove tasks with no input
```

FlorianJacta (dismissed):
I agree with the principle. I let the others review the code 😄

taipy/core/config/scenario_config.py
```diff
@@ -86,6 +88,35 @@ def __init__(
...
                     self.comparators[_validate_id(k)].append(v)
         super().__init__(id, **properties)
+        self.__build_datanode_configs_ranks()
+
+    def __build_datanode_configs_ranks(self):
+        # build the DAG
+        dag = nx.DiGraph()
+        for task_cfg in self._tasks:
+            if has_inputs := task_cfg.inputs:
+                for predecessor in task_cfg.inputs:
```
FredLL-Avaiga: why not use has_inputs here (maybe rename it ?)

FredLL-Avaiga (dismissed):
with comment

taipy/core/config/scenario_config.py
jrobinAV: Why not, indeed? Let me change that.

taipy/core/data/data_node.py
```diff
@@ -14,14 +14,15 @@
...
-from typing import Any, Dict, List, Optional, Set, Tuple, Union
+from typing import Any, Dict, List, Optional, Set, Tuple, Union, cast
 
 import networkx as nx
 
 from taipy.common.config.common._validate_id import _validate_id
 from taipy.common.config.common.scope import Scope
 from taipy.common.logger._taipy_logger import _TaipyLogger
 
+from ... import Config
```
trgiangdo: This should be `from taipy.common.config import Config`, not the relative import.

taipy/core/data/data_node.py
```diff
@@ -567,6 +572,19 @@ def get_last_edit(self) -> Optional[Edit]:
         """
         return self._edits[-1] if self._edits else None
 
+    def get_rank(self) -> int:
```
trgiangdo: Maybe add a docstring here explain the returned rank since this is a public method

trgiangdo (commented):
Minor comments

taipy/core/data/data_node.py
FredLL-Avaiga: Ok but why?

taipy/core/data/data_node.py
FredLL-Avaiga: Will do it, tx

taipy/core/data/data_node.py
FredLL-Avaiga: Oh I get it ... different packages, you're right it's cleaner

trgiangdo (dismissed):
LGTM

toan-quach (dismissed):
lgtm :D

jrobinAV:
Actually LNGTM.

I don't like it, and I'm having difficulty explaining why. Please give me some time to figure it out.

taipy/core/data/data_node.py
```diff
@@ -605,6 +608,31 @@ def get_last_edit(self) -> Optional[Edit]:
...
+        if not scenario_config_id:
+            return 0xfffb
+        dn_config = Config.data_nodes.get(self._config_id, None)
+        if not dn_config:
+            self._logger.error(f"Data node config `{self.config_id}` for data node `{self.id}` is not found.")
+            return 0xfffd
+        if not dn_config._ranks:
+            self._logger.error(f"Data node config `{self.config_id}` for data node `{self.id}` has no rank.")
+            return 0xfffe
+        return dn_config._ranks.get(scenario_config_id, 0xfffc)
```
jrobinAV: Any reason for having changed the returned values? The doc should be also updated.
```suggestion
            The int value representing the rank of the data node config in the scenario config DAG.
            If the scenario config is None or an empty string, 0xfffb is returned.<br/>
            If the data node config is not found, 0xfffd is returned. This case cannot
            happen in a normal situation.<br/>
            If the data node config has no precomputed ranks, 0xfffe is returned. This case
            cannot happen in a normal situation.<br/>
            If the data node config is not part of the scenario config, 0xfffc is returned as an infinite rank.
        """
        if not scenario_config_id:
            return 0xfffb
        dn_config = Config.data_nodes.get(self._config_id, None)
        if not dn_config:
            self._logger.error(f"Data node config `{self.config_id}` for data node `{self.id}` is not found.")
            return 0xfffd
        if not dn_config._ranks:
            self._logger.error(f"Data node config `{self.config_id}` for data node `{self.id}` has no rank.")
            return 0xfffe
        return dn_config._ranks.get(scenario_config_id, 0xfffc)
```

FredLL-Avaiga: made more sense to me

PR #2119, opened 2024-10-21
Fix wrong Callable type

jrobinAV:


tests/conftest.py
```diff
@@ -54,7 +55,7 @@ def remove_subparser(name: str) -> None:
 
 
 @pytest.fixture
-def clean_argparser() -> callable:
+def clean_argparser() -> t.callable:
```
FredLL-Avaiga: I think it's t.Callable and not t.callable

FredLL-Avaiga (changes requested):

tests/conftest.py
jrobinAV: Yes, of course. Sorry abt that

FredLL-Avaiga (approved):

PR #2128, opened 2024-10-22
Add command line options for run_browser and dark_mode Gui.run() parameters

FabienLelaquais:
Add the --run-browser, --no-run-browser, --dark-mode, and --light-mode CLI options.


taipy/gui/_gui_cli.py
```diff
@@ -81,21 +81,55 @@ class _GuiCLI(_AbstractCLI):
...
     }
 
+    __BROWSER_ARGS: Dict[str, Dict] = {
+        "--run-browser": {
+            "dest": "taipy_run_browser",
+            "help": "Open a new tab in the system browser",
+            "action": "store_true",
+        },
+        "--no-run-browser": {
+            "dest": "taipy_no_run_browser",
```
FredLL-Avaiga, line 91: can we not use the same `dest` with action: `store_false` ?

FredLL-Avaiga (dismissed):
use only one var ?

taipy/gui/_gui_cli.py
FabienLelaquais: I agree. I chose to use the same pattern as for other options for homogeneity.
But I had the same question, with the same answer. This could be simplified.

FabienLelaquais: Actually after some thought I know why it was made this way: we want the CLI to have the most weight at the end of the day.
If you use a CLI option, it will win.
Now we don't have a way to detect whether an option has been mentioned or not, except setting some value to True.
Indeed that means that for Boolean values, we have to separated options ON and OFF, each being potentially set to True so that we don't enforce anything if no option was set.

dinhlongviolin1 (dismissed):
Lgtm

taipy/gui/_gui_cli.py
```diff
@@ -81,21 +81,55 @@ class _GuiCLI(_AbstractCLI):
...
+            "help": "Don't open a new tab for the application",
+            "action": "store_true",
+        },
+    }
+
+    __DARK_LIGHT_MODE_ARGS: Dict[str, Dict] = {
+        "--dark-mode": {
+            "dest": "taipy_dark_mode",
+            "help": "Apply the dark mode to the GUI application",
```
dinhlongviolin1: Maybe Apply dark mode and removing the 'the' is enough

FabienLelaquais: 👍

dinhlongviolin1 (approved):

PR #2137, opened 2024-10-23
Refactor - Expand Event and _get_manager() method

trgiangdo:
Expand some of the class and method to prepare for the User entity


taipy/core/notification/event.py
```diff
@@ -62,23 +62,20 @@ class EventEntityType(_ReprEnum):
...
-)
-_ENTITY_TO_EVENT_ENTITY_TYPE = {
-    "scenario": EventEntityType.SCENARIO,
-    "sequence": EventEntityType.SEQUENCE,
-    "task": EventEntityType.TASK,
-    "data": EventEntityType.DATA_NODE,
-    "job": EventEntityType.JOB,
-    "cycle": EventEntityType.CYCLE,
-    "submission": EventEntityType.SUBMISSION,
-}
```
toan-quach, line 81: is there a reason for the move?

toan-quach (commented):
so far so good!

taipy/core/notification/event.py
trgiangdo: These constants are moved so that they become class variables, which can be edit properly by enterprise.

If they stay outside of the class, importing these constants and using them may not have the edit from enterprise.

taipy/core/notification/event.py
```diff
@@ -62,23 +62,20 @@ class EventEntityType(_ReprEnum):
...
+    @classmethod
+    def add_member(cls, name, value):
```
jrobinAV, line 66: EventEntityType is exposed so that our users can use it. I suspect the new dynamically added members won't have auto-completion, even if they are added by Taipy (or Taipy Enterprise) directly.
That could be a problem.

trgiangdo: True, there won't be autocompletion. The only way for autocompletion is by defined all member here.

jrobinAV: Yep. Let's keep that in mind for future.

toan-quach (dismissed):

jrobinAV (dismissed):

jrobinAV:
Still some linter issues to fix

toan-quach (dismissed):

jrobinAV (dismissed):

jrobinAV (approved):

PR #2139, opened 2024-10-24
Modify Release Action to Bump development versions and create a PR

joaoandre-avaiga:


.github/workflows/build-and-release.yml
```diff
@@ -238,11 +239,23 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Bump Version
+        id: bump-version
+        run: |
+          python tools/release/bump_version.py
+
       - uses: stefanzweifel/git-auto-commit-action@v5
         with:
-          file_pattern: '*/version.json'
```
FredLL-Avaiga, line 243: Should package.json files be impacted too ?

tools/release/setup_version.py
```diff
@@ -79,10 +79,12 @@ def __setup_dev_version(version: Version, _base_path: str, name: Optional[str] =
     name = f"{name}_VERSION" if name else "VERSION"
     print(f"{name}={version.dev_name}")  # noqa: T201
 
-    version.bump_ext_version()
+    print(f"NEW_{name}={version.dev_name}")  # noqa: T201
```
jrobinAV: These 2 consecutive prints look strange to me. Can you double-check?

tools/release/setup_version.py
```diff
@@ -119,6 +121,7 @@ def __setup_prod_version(version: Version, target_version: str, branch_name: str
 
         if _environment == "dev":
             __setup_dev_version(_version, _path, _name)
+            bump_ext_version(_version, _path)
```
jrobinAV: It's not clear to me. Here is my understanding of the flow: 

- We dev release using the version in the version.json files.
- We bump up the version and write the version.json files
- We commit and push the changes

If this is correct, I don't get why you changed the version before releasing.

.github/workflows/build-and-release.yml
joaoandre-avaiga: Is package.json a file that should be updated after releasing?

FredLL-Avaiga: yes it would be nice

FredLL-Avaiga:
I'll keep on updating versions in package.json manually then

PR #2157, opened 2024-10-25
Update release version

dinhlongviolin1:


.github/workflows/publish.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       - name: Checks if package is already on Pypi
         id: check-version
         run: |
-          if curl https://pypi.org/simple/taipy-${{ matrix.package }} | grep -o ">taipy-${{ matrix.package }}-${{ github.event.inputs.version }}\.tar\.gz<"; then
+          if curl https://pypi.org/simple/taipy-${{ matrix.package }}/ | grep -o ">taipy-${{ matrix.package }}-${{ github.event.inputs.version }}\.tar\.gz<"; then
```
FredLL-Avaiga, line 54: What is this for ?

FredLL-Avaiga (approved):

.github/workflows/publish.yml
dinhlongviolin1: not sure, but joao did it in the last dev release so I think it is needed

dinhlongviolin1: https://github.com/Avaiga/taipy/tree/release/4.0.0.dev6

FabienLelaquais (approved):
👍

Issue #588, opened 2023-12-11
Make more MUI classes available for Buttons


AAstridVH:
Make the MUI classes "Small", "Medium" and "Large" available for buttons.

https://mui.com/material-ui/react-button/

![image](https://github.com/Avaiga/taipy/assets/96520448/01457b92-6452-4219-b28d-f83f86b2a8ae)

FlorianJacta:
This issue comes from a user on Discord. It is quite hard to change the size of a button if you don't know CSS. 

main.py:

```python
from taipy import Gui 

buttons = """
<|Small|button|class_name=smallButton|>
<|Big|button|class_name=bigButton|>"
"""

Gui(buttons).run()
```

main.css:
```css
.smallButton{
    min-height: unset;
    line-height: 1.2rem;
    padding: 0;
    font-size: 1rem;
    text-transform: initial;
}

.bigButton{
    min-height: unset;
    line-height: 3.6rem;
    padding: 0;
    font-size: 3rem;
    text-transform: initial;
}
```

![image](https://github.com/Avaiga/taipy/assets/98709993/5021b97a-8307-478d-9380-c3f622e2854b)

## Environment

Taipy 3.0.0

Dljdd:
Hi, I can help with the implementation of this

jrobinAV:
@Dljdd Thank you for your help.

ArjunCodess:
@quest-bot embark

ArjunCodess:
@quest-bot embark

Dljdd:
@quest-bot embark

Rishi-0007:
Hi @jrobinAV , could you please assign this to me?

FlorianJacta:
You are now assigned @Rishi-0007

Ryadav0654:
Hi @FlorianJacta ,
I would like to work on this issue. Could you please assign it to me?
Thank you!

FlorianJacta:
Assigned!

Shayan12456:
i am working on it!

VivekKrYadav2078:
I would like to work on it, and I think I can solve it .Can you please assign this issue to me?
And as a first time contributor to this ,I need some help and guidance.

FlorianJacta:
You are assigned @VivekKrYadav2078

JeevaRamanathan:
Hi @FlorianJacta, I would like to work on this issue

nunoferna:
Hi, I would like to work on this issue!

[...]

PR #2159, opened 2024-10-25
added more mui classes for buttons

JeevaRamanathan:
Address #588
Added `size` and `variant` for the button, along with testcase and example. Please let me know if this is on the right track and required any changes.


taipy/gui/_renderers/factory.py
```diff
@@ -99,6 +99,8 @@ class _Factory:
                 ("active", PropertyType.dynamic_boolean, True),
                 ("hover_text", PropertyType.dynamic_string),
                 ("width", PropertyType.string_or_number),
+                ("size", PropertyType.dynamic_string),
```
FredLL-Avaiga: no need for these to be dynamic (string is enough)

frontend/taipy-gui/src/components/Taipy/Button.tsx
```diff
@@ -40,7 +44,8 @@ const Button = (props: ButtonProps) => {
     const className = useClassNames(props.libClassName, props.dynamicClassName, props.className);
     const active = useDynamicProperty(props.active, props.defaultActive, true);
     const hover = useDynamicProperty(props.hoverText, props.defaultHoverText, undefined);
-
+    const variant = useDynamicProperty(props.variant,props.defaultVariant,"outlined") as "outlined"| "text" | "contained";
+    const size = useDynamicProperty(props.size,props.defaultSize,"medium") as "small"| "medium" | "large";
```
FredLL-Avaiga: no need to make these dynamic

FredLL-Avaiga (changes requested):
the viselements.json file should be also impacted.

frontend/taipy-gui/src/components/Taipy/Button.tsx
JeevaRamanathan: Hi @FredLL-Avaiga, I couldn’t find any hooks for string similar to useDynamicProperty. Does this approach is valid here?
`const variant = (props.variant ?? "outlined") as "outlined" | "text" | "contained"`

FredLL-Avaiga: No need for hooks
You can simply set the default value when you destructure the props

FredLL-Avaiga (changes requested):
Tx

frontend/taipy-gui/src/components/Taipy/Button.tsx
JeevaRamanathan: @FredLL-Avaiga Quick question: a bit confused, is the updated code (doesn't use hooks) I used fine, or does it need any changes?

FredLL-Avaiga: Please change it
You can set the default value in the line destructuring the props
Something like
const { variant = "outlined", ... } = props;

JeevaRamanathan: Done, I’ve changed it. Thanks!

JeevaRamanathan:
Hi @FredLL-Avaiga, the changes have been made. Please have a look based on your time.

FredLL-Avaiga:
We're discussing with @FabienLelaquais what should be done ...

JeevaRamanathan:
@quest-bot loot #588

FabienLelaquais (approved):
Good job

FabienLelaquais (approved):
👍

FredLL-Avaiga (approved):

PR #2161, opened 2024-10-26
Replacing all 'args:' and 'parameteres' with 'arguments'

unsparsh:
Issue Description
Function and method parameters are documented using either Parameters, Arguments, or Args keywords. We should only use Arguments everywhere.


jrobinAV:
The changes made on this PR are not on the documentation which does not match the description. 
Moreover there is no issue related and no explicit added value.

I am closing the PR.

Issue #848, opened 2024-02-19
Add validation for date picker and other time/date-related controls


arcanaxion:
## Problem
Currently, the date picker does not support validation, i.e., the date picker component itself does not restrict the user from selecting particular dates. Often, date pickers require some constraints, e.g.:

- Can only select weekdays
- Can only select past/future days
- Cannot select days where the restaurant is closed or unavailable for reservation

Currently in taipy, to implement validation, we would do something like:
1. Write some text informing the user of which dates are (in)valid
2. If the user selects an invalid date, reset the date to a valid date and send a notification

This is cumbersome for the user, as they'd have to trial and error through the dates to discover which dates still have slots for reservation, etc.

## Proposed solution

Date pickers could be substantially improved by adding validation -- disallowing the user from selecting invalid dates.

In the following example, the user may not select weekends (weekends are grayed out):

![image](https://github.com/Avaiga/taipy/assets/48582724/7ac5fe56-b455-4ae9-9ae2-985eb06cb1ae)

A good way to do this might be by allowing the developer to specify a function which returns true for valid dates, like so: https://mui.com/x/react-date-pickers/validation/#disable-specific-dates

Please see the following link for other possibilities: https://mui.com/x/react-date-pickers/validation/

Thanks!

## Standards appliance

We should closely look at [RFC5545](https://datatracker.ietf.org/doc/html/rfc5545) that looks like a standard manner for defining events recurrence, therefore we could use that to define recurrent allowed or disallowed time spans.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

jrobinAV:
Please provide more details.

arcanaxion:
> Please provide more details.

Hi, I've updated the main comment. Thanks

FlorianJacta:
A user asked for it [here](https://github.com/Avaiga/taipy/discussions/954#discussioncomment-8742391)

yaten2302:
Hey @FlorianJacta @jrobinAV, I would like to work on this issue, may I take this if it's required?

FlorianJacta:
What are your thoughts @FabienLelaquais @FredLL-Avaiga ?

FredLL-Avaiga:
I think @namnguyen20999 is already working on this

yaten2302:
[...]

PR #2165, opened 2024-10-26
Implement Validation for Date Picker

JeevaRamanathan:
Hi @FredLL-Avaiga,
Referencing [issue #848](https://github.com/avaiga/taipy/issues/848#issuecomment-2439446106), I’ve added the following specifications for date validation to handle date disabling options:
```
{
    disableWeekdays?: boolean;
    disableWeekends?: boolean;
    disablePastDays?: boolean;
    disableFutureDays?: boolean;
    dayOfWeek?: number; // 0-6 (0 = Monday, ..., 6 = Sunday)
    interval?: number;
    oddInterval?:boolean;
    occurence?: number;
}
```

The rules for disabling dates include options for disabling past days or future days, weekdays, weekends, or specific day of the month with `disablePastDays`, `disableFutureDays`,`disableWeekdays`, `disableWeekends`,`dayOfWeek`.

To disable alternating days:
```
disable_date_config_rule = {
    dayOfWeek: 3,
    interval: 2 // Disables alternate Wednesdays by default (even weeks)
}
<|{date}|date|disable_date_config={disable_date_config_rule}|>
```
Setting `oddInterval: True` will disable Wednesdays on the 1st, 3rd, and 5th weeks

To disable the nth occurrence of a specific weekday in the month:
```
disable_date_config_rule = {
    dayOfWeek: 1,
    occurrence: 1 // Disables the first Monday of each month
}
```
Could you please review this and let me know if any modifications are needed for better handling of these specifications?


FredLL-Avaiga:
I think a dict (object) might not be the more user-friendly way for the user...
Let's call a friend : @FabienLelaquais what do you think?

FabienLelaquais:
> I think a dict (object) might not be the more user-friendly way for the user... Let's call a friend : @FabienLelaquais what do you think?

Actually I do think that a dict may be the only option we have.
The expressivity we want is going to need several slots that are specified or not.

However, I wouldn't hardcode the notion of weekend or workday. This varies across cultures.

My first thoughts about this would be a dict with two optional slots: one would indicate dates that must be forbidden, another would be dates that are explicitly available.
Say you want to book a table at a restaurant that opens only on Thursdays and Fridays, but closes for the Christmas season.
So you want to allow the selection of all the future Thursdays and Fridays to be selectable, except for the vacation period.

The way I see you can express this is:
```
date_constraints = {
   "accept": {
       "days_of_week": [3, 4]
     },
     "reject": {
       "past": True,
       "period": [ [ (2024, 12, 23), (2024, 12, 31) ] ]
      }
}
```
This has the issue of having days that fall in the two buckets, the a priority may be needed.

Now I like the 'occurrence' thing but we need to allow for more expressivity, again.
Like catching "U.S. presidential elections are held every four years on the first Tuesday after the first Monday in November".
That sounds tricky to me... except if you allow for an explicit list of accepted dates.

Same for 'repeat' (or 'interval'). The period (week, month, year...) needs to be expressed somehow.

JeevaRamanathan:
> This has the issue of having days that fall in the two buckets, the a priority may be needed.

@FabienLelaquais, for example, Wednesday and Thursday (days 3 and 4) are accepted but also fall within the rejection period. Are you referring to this as falling into two buckets and needing prioritization, or is it something different?

JeevaRamanathan:
Hi @FabienLelaquais, I'd like to get your thoughts on the updated structure for date_constraints. I’ve added support for priority, weekly, monthly, yearly rules along with the after week attribute to address the cases like US election logic.

1. Weekly
```
  const date_constraints = {
    priority: "reject", //optional
    accept: {
      day_of_week: [0, 1],
    },
    reject: {
      past: true,
      period: [
        { start: "2024, 11, 23", end: "2024, 11, 30 " },
        { start: "2025, 11, 23", end: "2025, 11, 30" },
      ],
      repeat: {
        frequency: "weekly",
        occurrence: [
          {
            day_of_week: 1, // disables every monday
            interval: 1,
            odd_interval: false,
          },
          {
            day_of_week: 3,  // disables alternate wednesday
            interval: 2,
            odd_interval: false,
          },
        ],
      },
    },
  };
```

2. Monthly
```
const date_constraints = {
    priority: "reject",
    accept: {
      // day_of_week: [0, 1],
    },
    reject: {
      future:true,
      repeat: {
        frequency: "monthly",
        occurrence: [
          {
            day_of_week: 6, // saturday
            after: {
              week: 1, //first week is skipped
            },
            interval: 2, // specific interval for this occurrence
          },
          {
            day_of_week: 5,
            after: {
              week: 3,
            },
            interval: 3,
          },
        ],
      },
    },
  };
```

3. Yearly - This handles the yearly constraints for specific days, like the US election, where the date pattern is every 4 years on the first Tuesday after the first Monday in November using the `after` week property.
```
const date_constraints = {
    accept: {
      // day_of_week: [0, 1]
    },
    reject: {
      past: true,
      // future: false,
      period: [],
      repeat: {
        frequency: "yearly",
        occurrence: [
          {
            month: 10, // November
            day_of_week: 2, // Tuesday
            after: {
              week: 1, // After the first week
            },
            interval: 4, // Every 4 years
            start_year: 2024, // Start from the year 2024
          },
          {
            month: 6, // July 
            day_of_week: 5, // Friday
            after: {
              week: 2// After the second week
            },
            interval: 3, // Every 3 years
            start_year: 2025, // Start from the year 2025
          },
        ],
      },
    },
  };
```
Prolly the same repeat format can be used for accept; [Here's](https://dub.sh/Tbyonnc) the sample frontend workaround logic for this structure.

taipy/gui/_renderers/factory.py
```diff
@@ -169,6 +169,7 @@ class _Factory:
                 ("on_change", PropertyType.function),
                 ("format",),
                 ("width", PropertyType.string_or_number),
+                ("disable_date_config",PropertyType.dynamic_dict)
```
FredLL-Avaiga: SHould this be dynamic ?

FredLL-Avaiga (changes requested):
Can you fix the failing checks ?

taipy/gui/_renderers/factory.py
JeevaRamanathan: I believe it should be dynamic, as some keys in the dictionary are optional. Please correct me if I'm wrong.

FredLL-Avaiga: > I believe it should be dynamic, as some keys in the dictionary are optional. Please correct me if I'm wrong.

Dynamic means that the bound variable can be updated.

JeevaRamanathan: Thank you for the clarification, @FredLL-Avaiga. I will change it to dict.

I have fixed the checks that failed due to spelling. The remaining tests are failing because no test cases were written, correct?

FredLL-Avaiga: there are some tests failing due to an exterior factor (they're failing in the rest subtree)

frontend/taipy-gui/src/components/Taipy/DateSelector.tsx
```diff
@@ -46,7 +56,9 @@ interface DateSelectorProps extends TaipyActiveProps, TaipyChangeProps {
     editable?: boolean;
     label?: string;
     width?: string | number;
-    analogic? :boolean;
+    analogic?: boolean;
+    disableDateConfig?: string;
+    defaultDisableDateConfig?: string;
```
FredLL-Avaiga, line 61: no need for this as it is not dynamic

frontend/taipy-gui/src/components/Taipy/DateSelector.tsx
```diff
@@ -74,6 +86,8 @@ const DateSelector = (props: DateSelectorProps) => {
     const hover = useDynamicProperty(props.hoverText, props.defaultHoverText, undefined);
     const min = useDynamicProperty(props.min, props.defaultMin, undefined);
     const max = useDynamicProperty(props.max, props.defaultMax, undefined);
+    const emptyDateConfig = {} as Partial<DisableDateConfig>;
+    const disableDateConfig = useDynamicJsonProperty(props.disableDateConfig, props.defaultDisableDateConfig || "", emptyDateConfig);
```
FredLL-Avaiga, line 90: no need for this as it is not dynamic, but your need to JSON parse it

frontend/taipy-gui/src/components/Taipy/DateSelector.tsx
```diff
@@ -115,25 +129,76 @@ const DateSelector = (props: DateSelectorProps) => {
         }
     }, [props.date, tz, withTime, max, min]);
 
+
+    const getWeekNumberInMonth = (date: Date) => {
```
FredLL-Avaiga, line 133: this should be defined outside the component

frontend/taipy-gui/src/components/Taipy/DateSelector.tsx
```diff
@@ -115,25 +129,76 @@ const DateSelector = (props: DateSelectorProps) => {
...
+        const d = new Date(Date.UTC(date.getFullYear(), date.getMonth(), date.getDate()));
+        const firstDayOfMonth = new Date(Date.UTC(d.getUTCFullYear(), d.getUTCMonth(), 1));
+        d.setUTCDate(d.getUTCDate() + 4 - (d.getUTCDay() || 7));
+
+        const weekNo: number = Math.ceil(((d.getTime() - firstDayOfMonth.getTime()) / 86400000 + 1) / 7);
+        return weekNo;
+    }
+    const isDisabledDate = (date: Date) => {
```
FredLL-Avaiga, line 141: useCallback

FredLL-Avaiga (changes requested):
I'd like to see a few frontend tests.
Can you add a time constraint in the rules ? (ie for example allow only to select time that are multiple of 15 minutes)

frontend/taipy-gui/src/components/Taipy/DateSelector.tsx
JeevaRamanathan: I would like to confirm the structure. Can I make these changes to the initial structure for the time?
> {
>     disableWeekdays?: boolean;
>     disableWeekends?: boolean;
>     disablePastDays?: boolean;
>     disableFutureDays?: boolean;
>     dayOfWeek?: number; // 0-6 (0 = Monday, ..., 6 = Sunday)
>     interval?: number;
>     oddInterval?:boolean;
>     occurence?: number;
> }

Or should I follow the new logic that is currently under discussion and has not been finalized after FabienLelaquais comment?

FredLL-Avaiga: we'll wait for @FabienLelaquais 's input

FredLL-Avaiga:
Hi @JeevaRamanathan 
Are you still working on this ?

JeevaRamanathan:
> we'll wait for @FabienLelaquais 's input

Hi @FredLL-Avaiga, awaiting the input since the other changes are interconnected with the structure. Thanks!

FredLL-Avaiga:
> > we'll wait for @FabienLelaquais 's input
> 
> Hi @FredLL-Avaiga, awaiting the input since the other changes are interconnected with the structure. Thanks!

I'll ping him
There are a few comments that you can answer in the meantime.

FabienLelaquais:
Thank you, @FredLL-Avaiga, for the ping, and @JeevaRamanathan, for your patience.
Here is what's happening: I cannot figure out an API and the underlying object model that pleases my meticulous mindset.
I'm digging into  [IETF's RFC 5545](https://datatracker.ietf.org/doc/html/rfc5545) to get some insight.
This is what iCalendar uses, so I think it has been thought through.
Now the question will be, how to we implement that and expose it to Python programmers, then serialize and analyze on the front-end so it's both maintainable and reasonable in development time.
I'm on it, believe me. And I'll keep you posted for sure.

PR #2183, opened 2024-10-30
feat: add cleanup dev releases action

joaoandre-avaiga:


tools/release/delete_dev_releases.sh
```diff
@@ -0,0 +1,36 @@
...
+      echo "Latest release found! Skipping"
+    else
+      to_delete+=("$tag")
+    fi
+done
+
+# Delete the releases
+for tag in "${to_delete[@]}"; do
+    echo "Deleting pre-release with tag: $tag"
+    gh release delete "$tag" --repo "$REPO" --yes ----cleanup-tag
```
jrobinAV, line 33: I believe this does not delete the tag but just the release. We should delete both, no?

joaoandre-avaiga: `--cleanup-tag` deletes the associated tag with the release.

jrobinAV: MAke sense. Sorry.

jrobinAV (approved):

PR #2186, opened 2024-10-30
Feature - Be able to trigger Benchmark action manually with a specific commit SHA

trgiangdo:
A part of https://github.com/Avaiga/taipy-benchmark/issues/11


trgiangdo:
Trigger taipy-benchmark from taipy can only be done on develop branch of taipy-benchmark.

What we want is to be able to trigger taipy-benchmark on a specific branch of taipy-benchmark.

PR #2192, opened 2024-10-31
support glob declaration in resource descriptions

FredLL-Avaiga:
```
    def get_scripts(self) -> t.List[str]:
        return ["lib/*.js"]

```


FabienLelaquais (dismissed):
Great idea!

namnguyen20999 (approved):

dinhlongviolin1 (approved):

PR #2194, opened 2024-11-01
file_selector example

FredLL-Avaiga:
🛑make `State` inherit from `SimpleNamespace` so that the linters don't mark `state.var` as an error
Should this single line change to State be backported to 4.0  @FabienLelaquais ?

### DIFF
--- a/doc/gui/examples/controls/file_selector_image.py
+++ b/doc/gui/examples/controls/file_selector_image.py
@@ -22,16 +22,16 @@ from taipy.gui import Gui, State
 path = ""
 image = None
 
-page = """
-<|{path}|file_selector|on_action=upload|extensions=png,jpg|>
-<|{image}|image|>
-"""
-
 def upload(state: State):
     img = Image.open(state.path)
     img_byte_arr = io.BytesIO()
     img.save(img_byte_arr, format="PNG")
     state.image = img_byte_arr.getvalue()
+
+page = """
+<|{path}|file_selector|on_action=upload|extensions=png,jpg|>
+<|{image}|image|>
+"""
 
 if __name__ == "__main__":
     Gui(page).run(title="File Selector - With image")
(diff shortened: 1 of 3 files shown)


doc/gui/examples/controls/file_selector_image.py
```diff
@@ -0,0 +1,36 @@
...
+import io
+
+from PIL import Image
+
+from taipy.gui import Gui, State
+
+path = ""
+image = None
+
+page = """
```
FabienLelaquais: Please split the <|file_selector|> and <|image|> fragments in two lines to make them more obvious to see.
Also, the callback function should be declared before the page definition, for homogeneity.

doc/gui/examples/controls/file_selector_simple.py
```diff
@@ -0,0 +1,33 @@
...
+import taipy.gui.builder as tgb
+from taipy.gui import Gui, State
+
+filename = "<unknown>"
+
+
+def print_filename(state: State):
+    print("filename is ", state.filename)  # noqa: T201
+
+def on_change(state: State, var: str, val: str):
```
FabienLelaquais: Why do we need this?

doc/gui/examples/controls/file_selector_simple.py
FabienLelaquais, line 1: If you really want to stress a specific Page Builder API feature, please let me know what that is.
Otherwise, Markdown would be the default code example implementation.

taipy/gui/state.py
```diff
@@ -26,7 +26,7 @@
     from .gui import Gui
 
 
-class State(ABC):
+class State(SimpleNamespace):
```
FabienLelaquais, line 29: I like that!

FabienLelaquais (dismissed):
Great!

Regarding the backport, I think it is safe and easy.
Cherry-pick will not be an option though.

FabienLelaquais:
> 🛑make `State` inherit from `SimpleNamespace` so that the linters don't mark `state.var` as an error Should this single line change to State be backported to 4.0 @FabienLelaquais ?

Actually, when code examples are neat, I guess they can be part of 4.0.1 as well.... Cherry-pick is therefore just fine.

doc/gui/examples/controls/file_selector_simple.py
FredLL-Avaiga: I'll stick to KISS

FabienLelaquais (approved):
👌

PR #2237, opened 2024-11-12
Better UI for Core selector elements

FabienLelaquais:
Addresses #2232.

* Minor rewording.
* Added customization for field name in Filter and Sort.

## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

- [X] Does this solution meet the acceptance criteria of the related issue?
- [X] Is the related issue checklist completed?
- [X] Was the documentation updated, or a dedicated issue for documentation created? (If applicable)

PR #2238, opened 2024-11-12
Rewording for Core selector filter/sort

### LINKED ISSUE #2232: Improve CoreSelector UI
### Description

# Problem

The generic selector (used in `scenario_selector` and `data_node_selector`) is slightly off in terms of UI concerns:

- The title of the property name to filter/sort on is 'Column' which is confusing
- If there is only a single property to filter/sort on, it should be selected by default (and probable to be a selector in the first place)
- The wording for the test is a bit off.
    - string: 'not equals' is unproper.
    - boolean: 'equals' are 'not equals' could better be 'is' and 'is not'
    - date: 'before' and 'after' are not verbs, contrary to 'equals', which is.
    - number has similar problems
- The "Empty string" default value for the last field is confusing (even worse when selecting it, as it become the field label).
  This might be simply 'Value'.
 
# Proposal

This proposal impacts both the `TableFilter` and `TableSort` components.

- Augment the `TableFilter` and `TableSort` APIs to provide the 'Column' label.
  This would be 'Property' when used by the Core Visual Elements.
- The 'Action' label would be better called "Condition"
- The operators would be better worded:
    - string: is, is not, contains, does not contain, starts with, end with...
    - Boolean: is, is not (better would have been "is true" and "is false" but that would break the homogeneity)
    - date: is on, is before, is after, is on or before, is on or after
    - number: is, is not, is greater than, is less than, is greater than or equal to, is less than or equal to
- Tooltips should be provided where relevant:
   - "Column" selector: "Select the table column you want to filter/sort on"/"Select the property name you want to filter/sort on"
   - "Action" selector: "Select the condition to apply to the selected column/property"
   - "Value": "Enter the value you want to filter/sort by"
- "Empty string" become "Value"

### Acceptance Criteria

### Code of Conduct

FabienLelaquais:
Backport of [PR#2237](https://github.com/Avaiga/taipy/pull/2237), resolves #2232.

### DIFF
--- a/frontend/taipy-gui/packaging/taipy-gui.d.ts
+++ b/frontend/taipy-gui/packaging/taipy-gui.d.ts
@@ -124,6 +124,8 @@ export interface FilterDesc {
     type: string;
 }
 export interface TableFilterProps {
+    fieldHeader?: string;
+    fieldHeaderTooltip?: string;
     columns: Record<string, ColumnDesc>;
     colsOrder?: Array<string>;
     onValidate: (data: Array<FilterDesc>) => void;
@@ -139,6 +141,8 @@ export interface SortDesc {
 }
 
 export interface TableSortProps {
+    fieldHeader?: string;
+    fieldHeaderTooltip?: string;
     columns: Record<string, ColumnDesc>;
     colsOrder?: Array<string>;
     onValidate: (data: Array<SortDesc>) => void;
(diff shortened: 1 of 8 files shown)


FredLL-Avaiga (dismissed):
🥇

FredLL-Avaiga (approved):

Issue #2017, opened 2024-10-10
[🐛 BUG] Data node's editor id mechanism does not work 


bobbyshermi:
### What went wrong? 🤔

When using a data node control, the lock/unlock/write mechanism works fine. But we cannot always use a datanode control. 

When locking a data node providing an "editor_id", this is only prevent another editor to lock the data node. It does not prevent the other editor to write the data node.

According to the documentation it should prevent another editor to write (append, upload) the data node. 

### Expected Behavior

When a data node is locked by an editor, no other editor should be able to write it. Same for append and _upload.

Question: When the data node is written without any `editor_id` (that's what the `Orchestrator` does), should this be accepted even if an editor has locked it? 

### Steps to Reproduce Issue

```python
from taipy import Config, Scope, create_scenario
from taipy.core.exceptions import DataNodeIsBeingEdited

d1 = Config.configure_data_node("d1", scope=Scope.GLOBAL, default_data=1)
s1 = Config.configure_scenario("s1", additional_data_node_configs=[d1])
scenario_1 = create_scenario(s1)
dn1 = scenario_1.d1

dn1.lock_edit()
try:
    dn1.write(4, editor_id="editor_id")
    print("SHOULD HAVE RAISED AN EXCEPTION! IT DIDN'T.")
except DataNodeIsBeingEdited as e:
    pass
dn1.unlock_edit()

dn1.lock_edit("editor_id")
try:
    dn1.write(5, editor_id="other")
    print("SHOULD HAVE RAISED AN EXCEPTION! IT DIDN'T.")
except DataNodeIsBeingEdited as e:
    pass
```

### Solution Proposed

The `editor_id` should be an optional parameter of the `write` method, just like the `lock` and `unlock` methods. 

```python

def write(self, data, job_id: Optional[JobId] = None, editor_id: Optional[str] = None, **kwargs: Dict[str, Any]):
        from ._data_manager_factory import _DataManagerFactory
        if editor_id:
            if not self.editor_id:
                raise DataNodeIsBeingEdited(self.id, self._editor_id)
            if (
                self.edit_in_progress
                and self.editor_id != editor_id
                and self.editor_expiration_date
                and self.editor_expiration_date > datetime.now()
            ):
                raise DataNodeIsBeingEdited(self.id, self._editor_id)
        self._write(data)
        self.track_edit(job_id=job_id, editor_id = editor_id, **kwargs)
        self.unlock_edit(editor_id)
[...]

PR #2240, opened 2024-11-13
CLean parameters for the datanode track_edit method.

jrobinAV:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
The PR aims to clean how the data node track edit method parameters are managed.
- A better split between the job_id and the editor_id.
- An explicit parameter naming so we can use constants as the edit keys.
- Unit tests
- Auto set the data node when an edit is added.

- Related Issue #2017 
- Closes #

## Checklist

- [x] Does this PR add unit tests for the developed code? If not, why?
- [x] Was the documentation updated or was a dedicated issue for documentation created? (If applicable)
- [x] Is the release notes updated? Not applicable

### DIFF
--- a/taipy/core/data/csv.py
+++ b/taipy/core/data/csv.py
@@ -116,16 +116,17 @@ class CSVDataNode(DataNode, _FileDataNodeMixin, _TabularDataNodeMixin):
     def storage_type(cls) -> str:
         return cls.__STORAGE_TYPE
 
-    def write_with_column_names(self, data: Any, columns: Optional[List[str]] = None, job_id: Optional[JobId] = None):
+    def write_with_column_names(self, data: Any, columns: Optional[List[str]] = None, editor_id: Optional[str] = None):
         """Write a selection of columns.
 
         Arguments:
             data (Any): The data to write.
             columns (Optional[List[str]]): The list of column names to write.
-            job_id (JobId): An optional identifier of the writer.
+            editor_id (str): An optional identifier of the writer.
         """
         self._write(data, columns)
-        self.track_edit(timestamp=datetime.now(), job_id=job_id)
+        timestamp = datetime.now()
+        self.track_edit(editor_id=editor_id, timestamp=timestamp)
 
     def _read(self):
         return self._read_from_path()
--- a/taipy/core/data/csv.py
+++ b/taipy/core/data/csv.py
@@ -124,8 +124,7 @@ class CSVDataNode(DataNode, _FileDataNodeMixin, _TabularDataNodeMixin):
             editor_id (str): An optional identifier of the writer.
         """
         self._write(data, columns)
-        timestamp = datetime.now()
-        self.track_edit(editor_id=editor_id, timestamp=timestamp)
+        self.track_edit(editor_id=editor_id, timestamp=datetime.now())
 
     def _read(self):
         return self._read_from_path()
--- a/taipy/core/data/csv.py
+++ b/taipy/core/data/csv.py
@@ -20,7 +20,6 @@ from taipy.common.config.common.scope import Scope
 
 from .._entity._reload import _Reloader
 from .._version._version_manager_factory import _VersionManagerFactory
-from ..job.job_id import JobId
 from ._file_datanode_mixin import _FileDataNodeMixin
 from ._tabular_datanode_mixin import _TabularDataNodeMixin
 from .data_node import DataNode
--- a/taipy/core/data/data_node.py
+++ b/taipy/core/data/data_node.py
@@ -11,6 +11,7 @@
 
 import functools
 import os
+import typing
 import uuid
 from abc import abstractmethod
 from datetime import datetime, timedelta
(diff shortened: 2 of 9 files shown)


taipy/core/data/csv.py
```diff
@@ -116,16 +115,17 @@ def __init__(
...
         Arguments:
             data (Any): The data to write.
             columns (Optional[List[str]]): The list of column names to write.
-            job_id (JobId): An optional identifier of the writer.
+            editor_id (str): An optional identifier of the writer.
         """
         self._write(data, columns)
-        self.track_edit(timestamp=datetime.now(), job_id=job_id)
+        timestamp = datetime.now()
+        self.track_edit(editor_id=editor_id, timestamp=timestamp)
```
jrobinAV:
```suggestion
        self.track_edit(editor_id=editor_id, timestamp=datetime.now())
```

taipy/core/data/data_node.py
```diff
@@ -447,20 +452,35 @@ def write(self, data, job_id: Optional[JobId] = None, **kwargs: Dict[str, Any]):
...
+            edit[EDIT_JOB_ID_KEY] = job_id
+        if editor_id:
+            edit[EDIT_EDITOR_ID_KEY] = editor_id
+        if comment:
+            edit[EDIT_COMMENT_KEY] = comment
+        if not timestamp:
+            timestamp = self._get_last_modified_datetime(self._properties.get(self._PATH_KEY)) or datetime.now()
+        edit[EDIT_TIMESTAMP_KEY] = timestamp
+        self.last_edit_date = edit.get(EDIT_TIMESTAMP_KEY)
+        self._edits.append(edit) # type: ignore
```
FredLL-Avaiga: `type: ignore` should be banned :-)
at least it should give a list of errors `type: ignore[...]`
but a cast would be better ifg t is a cast issue ;-)

taipy/core/data/data_node.py
```diff
@@ -415,19 +420,19 @@ def read(self) -> Any:
             )
             return None
 
-    def append(self, data, job_id: Optional[JobId] = None, **kwargs: Dict[str, Any]):
+    def append(self, data, editor_id: Optional[str] = None, **kwargs: Dict[str, Any]):
```
FredLL-Avaiga: kwargs shouldn't be typed like this, more like `**kwargs: Any`
https://stackoverflow.com/questions/37031928/type-annotations-for-args-and-kwargs

taipy/core/data/data_node.py
```diff
@@ -447,20 +452,35 @@ def write(self, data, job_id: Optional[JobId] = None, **kwargs: Dict[str, Any]):
         self.unlock_edit()
         _DataManagerFactory._build_manager()._set(self)
 
-    def track_edit(self, **options):
+    def track_edit(self,
+                   job_id: Optional[str] = None,
+                   editor_id: Optional[str] = None,
+                   timestamp: Optional[datetime] = None,
+                   comment: Optional[str] = None,
+                   **options: Dict[str, Any]):
```
FredLL-Avaiga: idem `**options: Any`

taipy/core/data/data_node.py
```diff
@@ -415,19 +420,19 @@ def read(self) -> Any:
...
-            job_id (JobId): An optional identifier of the writer.
+            editor_id (str): An optional identifier of the editor.
             **kwargs (dict[str, any]): Extra information to attach to the edit document
                 corresponding to this write.
         """
         from ._data_manager_factory import _DataManagerFactory
 
         self._append(data)
-        self.track_edit(job_id=job_id, **kwargs)
+        self.track_edit(editor_id=editor_id, **kwargs)
```
FredLL-Avaiga, line 436: will the current editor_id (set by lock_edit) be picked up automagically or is it not needed ?

FredLL-Avaiga (changes requested):

taipy/core/data/data_node.py
jrobinAV: yep, indeed.

taipy/core/data/data_node.py
jrobinAV: Ah, OK. Thanks! 
I believe that's the type ignore root cause.

taipy/core/data/data_node.py
jrobinAV: I don't understand the question. What do you mean by: "be picked up automagically or is it not needed" ?

The `editor_id` must be explicitly passed by the client when writing/appending the data so we can ensure it is the right editor and not someone else. The track_edit method is automatically called by the write and append methods with the appropriate editor id.

FredLL-Avaiga: that is *not* what's done currently in the vis elements
1/ datanode.lock_edit(client_id)
2/ user edit the data
3/ datanode.write(data, comment)
4/ datanode.unlock_edit(client_id)

So you're saying that I should pass client_id as editor_id in the write call ?

jrobinAV: Oh, ok. I get it now. This PR is really about cleaning the parameters of the track_edit method. But indeed, it is related to the behavior of the lock_edit and write methods.

Yes, we should change that in the future. There will be another (hacktoberfest) PR that will simplify the API soon, so you won't need to call the unlock_edit after writing the data.

In any case, what you did is working just fine because you ensure there is no concurrent edition on your side, and when you don't pass any editor_id to the write function, the data is still written. No problem.

I let you know when the other PR is done as well.

FredLL-Avaiga (approved):

PR #2266, opened 2024-11-21
Fix - Set default value for _in_context_attributes_changed_collector

trgiangdo:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

This PR fixes the problem in https://github.com/Avaiga/taipy-enterprise/pull/559.

When the `_in_context_attributes_changed_collector` has no value at initialization, and the `__getattr__` of an `_Entity` is called, it will raise `AttributeError`.

### DIFF
--- a/taipy/core/_entity/_entity.py
+++ b/taipy/core/_entity/_entity.py
@@ -19,7 +19,7 @@ class _Entity:
     _ID_PREFIX: str
     _MANAGER_NAME: str
     _is_in_context = False
-    _in_context_attributes_changed_collector: List
+    _in_context_attributes_changed_collector: List = []
 
     def __enter__(self):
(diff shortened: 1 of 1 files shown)


jrobinAV (commented):
Can we write a small unit test to exhibit the issue?

trgiangdo:
It only happens in https://github.com/Avaiga/taipy-enterprise/pull/559, where the `User.__setattr__()` tries to edit the `_properties`, which triggers the `_Properties.__setitem__` when in context, where the `_in_context_attributes_changed_collector` has no value yet.

jrobinAV (approved):

PR #2284, opened 2024-11-27
Feature/#521 refactor code for databricks integration

toan-quach:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### DIFF
--- a/taipy/core/config/checkers/_data_node_config_checker.py
+++ b/taipy/core/config/checkers/_data_node_config_checker.py
@@ -227,7 +227,7 @@ class _DataNodeConfigChecker(_ConfigChecker):
                             prop_key,
                             prop_value,
                             f"`{prop_key}` of DataNodeConfig `{data_node_config_id}` must be"
-                            f" populated with a {class_type.__name__}.",
+                            f" populated with a {class_type.__qualname__}.",
                         )
 
     def _check_exposed_type(self, data_node_config_id: str, data_node_config: DataNodeConfig):
--- a/taipy/core/config/checkers/_data_node_config_checker.py
+++ b/taipy/core/config/checkers/_data_node_config_checker.py
@@ -222,8 +222,8 @@ class _DataNodeConfigChecker(_ConfigChecker):
     def _check_class_type(self, data_node_config_id: str, data_node_config: DataNodeConfig):
         properties_to_check = _DataNodeConfigChecker._get_class_type_and_properties()
 
-        for storage_type in properties_to_check.keys():
-            for class_type, prop_keys in properties_to_check[storage_type]:
+        if data_node_config.storage_type in properties_to_check.keys():
+            for class_type, prop_keys in properties_to_check[data_node_config.storage_type]:
                 for prop_key in prop_keys:
                     prop_value = data_node_config.properties.get(prop_key) if data_node_config.properties else None
                     if prop_value and not isinstance(prop_value, class_type):
--- a/taipy/core/config/checkers/_data_node_config_checker.py
+++ b/taipy/core/config/checkers/_data_node_config_checker.py
@@ -23,6 +23,27 @@ from ..data_node_config import DataNodeConfig
 
 
 class _DataNodeConfigChecker(_ConfigChecker):
+    _PROPERTIES_TYPES: Dict[str, List[Tuple[Any, List[str]]]] = {
+        DataNodeConfig._STORAGE_TYPE_VALUE_GENERIC: [
+            (
+                Callable,
+                [
+                    DataNodeConfig._OPTIONAL_READ_FUNCTION_GENERIC_PROPERTY,
+                    DataNodeConfig._OPTIONAL_WRITE_FUNCTION_GENERIC_PROPERTY,
+                ],
+            )
+        ],
+        DataNodeConfig._STORAGE_TYPE_VALUE_SQL: [
+            (
+                Callable,
+                [
+                    DataNodeConfig._REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY,
+                    DataNodeConfig._OPTIONAL_APPEND_QUERY_BUILDER_SQL_PROPERTY,
+                ],
+            ),
+        ],
+    }
+
     def __init__(self, config: _Config, collector: IssueCollector):
         super().__init__(config, collector)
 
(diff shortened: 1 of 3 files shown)


taipy/core/config/checkers/_data_node_config_checker.py
```diff
@@ -196,28 +196,43 @@ def _check_generic_read_write_fct_and_args(self, data_node_config_id: str, data_
...
-    def _check_callable(self, data_node_config_id: str, data_node_config: DataNodeConfig):
-        properties_to_check = {
+    @staticmethod
+    def _get_class_type_and_properties() -> Dict[str, List[Tuple[Any, List[str]]]]:
+        return {
             DataNodeConfig._STORAGE_TYPE_VALUE_GENERIC: [
-                DataNodeConfig._OPTIONAL_READ_FUNCTION_GENERIC_PROPERTY,
-                DataNodeConfig._OPTIONAL_WRITE_FUNCTION_GENERIC_PROPERTY,
+                (
+                    Callable,
```
jrobinAV: I know that it was already the case before, but I actually believe it is buggy.
Callables include functions, classes, and lambdas. 
I am not sure lambdas are accepted as they must be serializable (JSON and TOML). Can you check with a unit test?

taipy/core/config/checkers/_data_node_config_checker.py
```diff
@@ -196,28 +196,43 @@ def _check_generic_read_write_fct_and_args(self, data_node_config_id: str, data_
                     f"DataNodeConfig `{data_node_config_id}` must be populated with a Callable function.",
                 )
 
-    def _check_callable(self, data_node_config_id: str, data_node_config: DataNodeConfig):
-        properties_to_check = {
+    @staticmethod
+    def _get_class_type_and_properties() -> Dict[str, List[Tuple[Any, List[str]]]]:
```
toan-quach: this can be turned into a class attribute instead of a function, as we were using function, I just went with it so can quickly implement it

taipy/core/config/checkers/_data_node_config_checker.py
toan-quach: Apparently both ways still think a `lambda` is the same as a function 
image

But I found this approach, it's a bit clunky but it works by adding `fct.__name__ == "<lambda>"`
image

Or this approach works as well https://stackoverflow.com/a/3655857

taipy/core/config/checkers/_data_node_config_checker.py
jrobinAV: I agree. We can implement it as a dictionary class attribute of the `DataNoceConfig` class. Just like we have `_REQUIRED_PROPERTIES: Dict[str, List]` or `_OPTIONAL_PROPERTIES: Dict[str, Dict[str, Any]]`, we could have something like :
```python
_PROPERTIES_TYPES: Dict[str, Dict[str, type]] = {
    DataNodeConfig._STORAGE_TYPE_VALUE_GENERIC: {
        DataNodeConfig._OPTIONAL_READ_FUNCTION_GENERIC_PROPERTY: Callable,
        DataNodeConfig._OPTIONAL_WRITE_FUNCTION_GENERIC_PROPERTY: Callable},
    DataNodeConfig._STORAGE_TYPE_VALUE_SQL: {
        DataNodeConfig._REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY: Callable,
        DataNodeConfig._OPTIONAL_APPEND_QUERY_BUILDER_SQL_PROPERTY: Callable}
}
```
By the way, we could use it to check the types of all properties (both required and optional), not only for callables.
Then, this dictionary could be extended to enterprise.

Does it make sense?

toan-quach: Yep, that's my initial intention with this refactoring. Trying to reuse the code for different types of instances. Though I'd say we should find a different ticket for that 😅

taipy/core/config/checkers/_data_node_config_checker.py
toan-quach: @jrobinAV should we create a ticket for this or work directly on this one as well? I think we can group it with the "check the types of all properties" topic into 1 ticket

taipy/core/config/checkers/_data_node_config_checker.py
```diff
@@ -23,6 +23,27 @@
...
+        DataNodeConfig._STORAGE_TYPE_VALUE_SQL: [
+            (
+                Callable,
+                [
+                    DataNodeConfig._REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY,
+                    DataNodeConfig._OPTIONAL_APPEND_QUERY_BUILDER_SQL_PROPERTY,
+                ],
+            ),
+        ],
+    }
```
jrobinAV, line 45: 1. We should also add all the properties to check. not only the callable. 
2. The dictionary should be moved close to the definition of the properties, in the DataNodeCOnfig class.
3. The format seems strange to me. I would have used a `Dict[str, Dict[str, Any]]` instead of `Dict[str, List[Tuple[Any, List[str]]]]`

    ```
    {
        DataNodeConfig._STORAGE_TYPE_VALUE_GENERIC: 
            {
                DataNodeConfig._OPTIONAL_READ_FUNCTION_GENERIC_PROPERTY: Callable,
                DataNodeConfig._OPTIONAL_WRITE_FUNCTION_GENERIC_PROPERTY: Callable
            },
        DataNodeConfig._STORAGE_TYPE_VALUE_SQL:
            {
                DataNodeConfig._REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY: Callable,
                DataNodeConfig._OPTIONAL_APPEND_QUERY_BUILDER_SQL_PROPERTY: Callable
            }
    }
    ```
    What do you think? Is there a reason to group the properties by type?

taipy/core/config/checkers/_data_node_config_checker.py
```diff
@@ -196,28 +217,25 @@ def _check_generic_read_write_fct_and_args(self, data_node_config_id: str, data_
...
+                            prop_value,
+                            f"`{prop_key}` of DataNodeConfig `{data_node_config_id}` must be"
+                            f" populated with a {'Callable' if class_type == Callable else class_type.__name__}.",
+                        )
+                    if class_type == Callable and callable(prop_value) and prop_value.__name__ == "<lambda>":
+                        self._error(
+                            prop_key,
+                            prop_value,
+                            f"`{prop_key}` of DataNodeConfig `{data_node_config_id}` must be"
+                            f" populated with a Callable and not a lambda.",
```
jrobinAV: The sentence is strange as a lambda is a Callable (can be called with the `()` operator).
```suggestion
                            f" populated with a serializable Callable function but not a lambda.",
```

tests/core/config/checkers/test_data_node_config_checker.py
```diff
@@ -581,6 +581,22 @@ def test_check_callable_properties(self, caplog):
...
+        expected_error_messages = [
+            "`write_fct` of DataNodeConfig `new` must be populated with a Callable and not a lambda."
+            " Current value of property `write_fct` is <function TestDataNodeConfigChecker."
+            "test_check_callable_properties.<locals>.<lambda>",
+            "`read_fct` of DataNodeConfig `new` must be populated with a Callable and not a lambda."
+            " Current value of property `read_fct` is <function TestDataNodeConfigChecker."
+            "test_check_callable_properties.<locals>.<lambda>",
+        ]
+        assert all(message in caplog.text for message in expected_error_messages)
+
```
jrobinAV: BTW, this deserves: 
- an entry in the release notes as an improvement of the checkers for checking the property types.
- A new item in the list: https://docs.taipy.io/en/develop/userman/advanced_features/configuration/config-checker/

taipy/core/config/checkers/_data_node_config_checker.py
toan-quach: Ah sure, well don't have much reason haha. Since the original code focused mainly on Callable, I just went along with that :D I have created a ticket to resolve this if you're okay with that since the changes are bigger than what we're having atm. https://github.com/orgs/Avaiga/projects/6/views/2?pane=issue&itemId=89479596

tests/core/config/checkers/test_data_node_config_checker.py
toan-quach: The PR I have created for this topic also contains changes required to the doc. https://github.com/orgs/Avaiga/projects/6/views/2?pane=issue&itemId=89479596

taipy/core/config/checkers/_data_node_config_checker.py
jrobinAV: Great. Thx!

jrobinAV (approved):

PR #2299, opened 2024-12-03
selector: selection message and select all

### LINKED ISSUE #1834: Add drop-down selectors with multiple values
### Description

Hi,
When I use selectors, there isn't a combination that lets me have a multiple-value selector that is also a drop-down. 

### Solution Proposed

Something like drop-down and check boxes for example. The following doesn't give an error but also doesn't "work":

```
tgb.selector
    (...
        multiple=True,
        dropdown=True,
        mode="check",)
```

**Another thing that could be great, is if the callback is only called after I select/unselect all my values (when I go out of the drop-down list). Right now, every time I check a box, it calls on_action, so if I check (or uncheck) a lot of boxes, it calls the Callback many times.**

### Impact of Solution

To give an example, I used a side pane to hold a long list of checkboxes, but it would have been quicker to have this drop-down list, and also more efficient. 

### Additional Context

_No response_

### Acceptance Criteria

### Code of Conduct

FredLL-Avaiga:
resolves #1834

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
- selection message property dynamic(string)
- show_select_all property boolean

- Closes #1834 

## How to reproduce the issue

```py

sel = None
sel_msg = None

def on_change(state: State, var_name: str, value):
  if var_name == "sel":
    state.sel_msg = f"{len(value)} selected elements"

page = """
<|{sel}|selector|dropdown|multiple|lov=Item 1;Item2 2;Item 3;Item 4|selection_message={sel_msg}|show_select_all|label=select an item|>
<|{sel}|selector|multiple|lov=Item 1;Item2 2;Item 3;Item 4|show_select_all|>
"""

if __name__ == "__main__":
  Gui(page).run(title="1834 Selector Dropdown Multiple Message")

```

### DIFF
--- a/taipy/gui/_renderers/factory.py
+++ b/taipy/gui/_renderers/factory.py
@@ -71,8 +71,7 @@ class _Factory:
     __LIBRARIES: t.Dict[str, t.List["ElementLibrary"]] = {}
 
     __CONTROL_BUILDERS = {
-        "alert":
-        lambda gui, control_type, attrs: _Builder(
+        "alert": lambda gui, control_type, attrs: _Builder(
             gui=gui,
             control_type=control_type,
             element_name="Alert",
@@ -507,6 +506,8 @@ class _Factory:
                 ("label",),
                 ("mode",),
                 ("lov", PropertyType.lov),
+                ("selection_message", PropertyType.dynamic_string),
+                ("show_select_all", PropertyType.boolean),
             ]
         )
         ._set_propagate(),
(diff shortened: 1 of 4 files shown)


namnguyen20999 (approved):
LGTM

taipy/gui/_renderers/factory.py
```diff
@@ -507,6 +506,8 @@ class _Factory:
                 ("label",),
                 ("mode",),
                 ("lov", PropertyType.lov),
+                ("selection_message", PropertyType.dynamic_string),
```
dinhlongviolin1, line 509: I think `selected_message` sounds better

dinhlongviolin1 (approved):
Lgtm! just a naming suggestion

FabienLelaquais (approved):
👌

taipy/gui/_renderers/factory.py
FredLL-Avaiga: I think `selection_message` is closer to what this actually does

FlorianJacta:
This is not working with filter=True. I'll create the issue on that

PR #2373, opened 2024-12-27
feature/#397 scenario duplication

### LINKED ISSUE #397: Possibility to duplicate scenarios
I want to duplicate a scenario. That means creating a new scenario and the related entities. The data nodes of the new scenario should have already been written and populated with the same data as the first scenario.

**Motivations:**
1. The possibility of starting with a scenario already set up represents an important time saving when many parameters or input data must be set before running a scenario.
 
2. When output data nodes of long tasks (for instance, formatting, preprocessing, or training) are already computed in a scenario, we want to duplicate the data and benefit from the task-skipping feature to save computation and time.

***Description of the ideal solution***
A new Scenario API should be exposed to duplicate a scenario. It should accept the new name and an optional list of data nodes to copy (by default all the scenario-scoped data nodes should be copied)

toan-quach:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Implementation of a new feature to duplicate a scenario.
+
Make the various reason messages consistent.

#397


tests/core/scenario/test_scenario_manager.py
```diff
@@ -1553,3 +1553,35 @@ def test_filter_scenarios_by_creation_datetime():
...
+    old_scenario_id = scenario.id
+
+    assert len(_ScenarioManager._get_all()) == 1
+    assert len(_DataManager._get_all()) == 3
+    assert len(_TaskManager._get_all()) == 1
+
+    new_scenario = _ScenarioManager._clone(scenario)
+    old_scenario = _ScenarioManager._get(old_scenario_id)
+
+    assert old_scenario.id != new_scenario.id
```
jrobinAV: We don't want to change the scenario (even in memory). we want to create a new one.

THe following must be tested.

```suggestion
    assert old_scenario.id != new_scenario.id
    assert old_scenario.id == scenario.id
```

Unfortunately, this means we need to change the implementation.

jrobinAV (changes requested):

tests/core/scenario/test_scenario_manager.py
toan-quach: fixed! :D

taipy/core/scenario/_scenario_manager.py
```diff
@@ -521,3 +521,47 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
...
+
+        Arguments:
+            scenario (Scenario): The scenario to clone.
+
+        Returns:
+            Scenario: The cloned scenario.
+        """
+        cloned_scenario = cls._get(scenario)
+        cloned_scenario_id = cloned_scenario._new_id(cloned_scenario.config_id)
+        cloned_scenario.id = cloned_scenario_id
```
jrobinAV:
```suggestion
        cloned_scenario.id = cloned_scenario._new_id(cloned_scenario.config_id)
```

taipy/core/data/_data_manager.py
jrobinAV, line 1: We should implement a can_duplicate method, returning reasons, just like we have a can_create method.

taipy/core/data/_data_manager.py
```diff
@@ -174,3 +179,19 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
...
+    ) -> DataNode:
+        cloned_dn = cls._get(dn)
+
+        cloned_dn.id = cloned_dn._new_id(cloned_dn._config_id)
+        cloned_dn._owner_id = cls._get_owner_id(cloned_dn._scope, cycle_id, scenario_id)
+        cloned_dn._parent_ids = set()
+
+        cls._set(cloned_dn)
+
+        cloned_dn._clone_data()
```
jrobinAV: If the data node has a cycle scope, and if the new scenario is from the same cycle, then we want to share the same data node between scenarios.
If the scope is global it always already exists, and we also want to share the existing one.

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -212,3 +222,14 @@ def _migrate_path(self, storage_type, old_path) -> str:
...
+    def _clone_data_file(self, id: str) -> Optional[str]:
+        if os.path.exists(self.path):
+            folder_path, base_name = os.path.split(self.path)
+            new_base_path = os.path.join(folder_path, f"TAIPY_CLONE_{id}_{base_name}")
+            if os.path.isdir(self.path):
+                shutil.copytree(self.path, new_base_path)
+            else:
+                shutil.copy(self.path, new_base_path)
+            return new_base_path
+        return ""
```
jrobinAV: Do we want to differentiate the cases where the initial path is generated by Taipy or provided by the user?
I believe it would be better. If it is Taipy generated, we can just replace the old id by the new one. otherwise, adding a prefix or a suffix as you did make sense.

taipy/core/scenario/_scenario_manager.py
```diff
@@ -521,3 +521,47 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
...
+        _data_manager = _DataManagerFactory._build_manager()
+
+        cloned_tasks = set()
+        for task in cloned_scenario.tasks.values():
+            cloned_tasks.add(_task_manager._clone(task, None, cloned_scenario_id))
+        cloned_scenario._tasks = cloned_tasks
+
+        cloned_additional_data_nodes = set()
+        for data_node in cloned_scenario.additional_data_nodes.values():
+            cloned_additional_data_nodes.add(_data_manager._clone(data_node, None, cloned_scenario_id))
```
jrobinAV: Duplicating data nodes should depend on its scope.

taipy/core/scenario/_scenario_manager.py
```diff
@@ -521,3 +521,47 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
...
+        # TODO: update sequences
+
+        # Clone tasks and data nodes
+        _task_manager = _TaskManagerFactory._build_manager()
+        _data_manager = _DataManagerFactory._build_manager()
+
+        cloned_tasks = set()
+        for task in cloned_scenario.tasks.values():
+            cloned_tasks.add(_task_manager._clone(task, None, cloned_scenario_id))
+        cloned_scenario._tasks = cloned_tasks
```
jrobinAV: Duplicating tasks should depend on minimal scope of inputs and outputs data nodes.

taipy/core/task/_task_manager.py
```diff
@@ -226,3 +231,22 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
...
         return cls._repository._load_all(filters)
+
+    @classmethod
+    def _clone(cls, task: Task, cycle_id: Optional[CycleId] = None, scenario_id: Optional[ScenarioId] = None) -> Task:
+        data_manager = _DataManagerFactory._build_manager()
+
+        cloned_task = cls._get(task)
+
+        inputs = [data_manager._clone(i, cycle_id, scenario_id) for i in cloned_task.input.values()]
+        outputs = [data_manager._clone(o, cycle_id, scenario_id) for o in cloned_task.output.values()]
```
jrobinAV: THis should depend on the scope of the data nodes.

taipy/core/data/_data_manager.py
```diff
@@ -174,3 +179,19 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
         for fil in filters:
             fil.update({"config_id": config_id})
         return cls._repository._load_all(filters)
+
+    @classmethod
+    def _clone(
```
jrobinAV: I would rename _clone to _duplicate

taipy/core/scenario/_scenario_manager.py
```diff
@@ -521,3 +521,47 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
         for fil in filters:
             fil.update({"config_id": config_id})
         return cls._repository._load_all(filters)
+
+    @classmethod
+    def _clone(cls, scenario: Scenario) -> Scenario:
```
jrobinAV: I would rename _clone method to _duplicate as it does not return another instance of the same object. It returns a similar object with a few differences (ids, sub-entities' ids, paths, etc...)

taipy/core/scenario/_scenario_manager.py
jrobinAV, line 1: We should implement a can_duplicate method, returning reasons, just like we have a can_create method.

taipy/core/scenario/_scenario_manager.py
jrobinAV: We should accept an optional name and an optional date in the method signature. It does not change much for the name but it does for the creation date. This has an impact on the potential cycle as the new scenario might be on a different cycle.

jrobinAV: According to the issue, we should also pass an optional list of data nodes or data node IDs. Without any list; we should copy all the data. If the list is provided, only the files of the data nodes in the list should be copied.

taipy/core/task/_task_manager.py
```diff
@@ -226,3 +231,22 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
         for fil in filters:
             fil.update({"config_id": config_id})
         return cls._repository._load_all(filters)
+
+    @classmethod
+    def _clone(cls, task: Task, cycle_id: Optional[CycleId] = None, scenario_id: Optional[ScenarioId] = None) -> Task:
```
jrobinAV: I would rename _clone method to _duplicate as it does not return another instance of the same object. It returns a similar object with a few differences (ids, sub-entities' ids, paths, etc...)

taipy/core/data/csv.py
```diff
@@ -192,3 +192,8 @@ def _write(self, data: Any, columns: Optional[List[str]] = None):
             encoding=properties[self.__ENCODING_KEY],
             header=properties[self._HAS_HEADER_PROPERTY],
         )
+
+    def _clone_data(self):
+        new_data_path = self._clone_data_file(self.id)
+        self._properties[self._PATH_KEY] = new_data_path
+        return new_data_path
```
jrobinAV: I would move that outside the data node, to put it in the data_manager.
Just like we handle the parent_ids and the owner_id in the manager, I would set the path property in the manager as well (still retrieving the value from a fileDatanodeMixing method).
This is debatable, though...

taipy/core/scenario/_scenario_manager.py
toan-quach: hm? I can understand the name, but the creation date?

jrobinAV: Let's say I have a January scenario I am happy with. I want to start with a duplicate of this one to compute my February scenario. So beginning of February I duplicate my January scenario passing the current date so the new scenario is in the February cycle. 

Does it make sense for you? And @FlorianJacta any opinion on that?

FlorianJacta: I agree; duplication for me was not about the creation date. Your use case is a real use case from CFM

taipy/core/data/csv.py
toan-quach: well this is specific for file DNs only, if we put it in the data manager, we're grouping it with other types of DNs like Sql or mongo, I don't think it's a good idea

taipy/core/scenario/_scenario_manager.py
toan-quach: I'll leave it as the next step for now

taipy/core/_repository/_filesystem_repository.py
```diff
@@ -191,10 +191,14 @@ def __filter_files_by_config_and_owner_id(
         return None
 
     def __match_file_and_get_entity(self, filepath, config_and_owner_ids, filters):
-        if match := [(c, p) for c, p in config_and_owner_ids if c.id in filepath.name]:
+        if match := [(c, p) for c, p in config_and_owner_ids if (c if isinstance(c, str) else c.id) in filepath.name]:
```
jrobinAV, line 194: How can c be a str here?
I don't get this change.

taipy/core/_repository/_filesystem_repository.py
```diff
@@ -227,6 +231,7 @@ def __filter_by(self, filepath: pathlib.Path, filters: Optional[List[Dict]]) ->
         except (FileNotFoundError, FileCannotBeRead, FileEmpty):
             return None
 
+        # breakpoint()
```
jrobinAV: To be removed

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -136,20 +139,23 @@ def _upload(self,
...
             reasons._add_reason(self.id, DataNodeEditInProgress(self.id))  # type: ignore[attr-defined]
             return reasons
 
         up_path = pathlib.Path(path)
         try:
             upload_data = self._read_from_path(str(up_path))
         except Exception as err:
-            self.__logger.error(f"Error uploading `{up_path.name}` to data "
-                                f"node `{self.id}`:")  # type: ignore[attr-defined]
+            self.__logger.error(f"Error uploading `{up_path.name}` to data " f"node `{self.id}`:")  # type: ignore[attr-defined]
```
jrobinAV: Please check your automatic formatter as this line is over 120 chars. Please revert the change.

taipy/core/scenario/_scenario_manager.py
```diff
@@ -521,3 +521,70 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
         for fil in filters:
             fil.update({"config_id": config_id})
         return cls._repository._load_all(filters)
+
+    @classmethod
+    def _duplicate(cls, scenario: Scenario, creation_date: Optional[datetime] = None) -> Scenario:
```
jrobinAV: what about passing an optional name for the new scenario and an optional list of data nodes to duplicate? Do you plan to do it in another PR?

taipy/core/_repository/_filesystem_repository.py
toan-quach: I want to be able to pass config.id as well as the config. As when duplicating, I'm also looking for existing entities based on the scope + owner.id + config, but if so, we will need to find the config again, while in the entity, we already have config_id and we only need config_id anyway so minor change here is easier and cleaner than looking for the config object from config_id

taipy/core/scenario/_scenario_manager.py
toan-quach: yep I'm planning to do it in another PR to make it cleaner as the current one is already rather complicated

jrobinAV: ALl right. I let you create the potential issue if you need it.

taipy/core/_repository/_filesystem_repository.py
jrobinAV: I don't get all the details, but you seem to have deeply thought about it. That's enough for me. 😆 
Good to me.

jrobinAV (dismissed):

jrobinAV (dismissed):

taipy/core/data/excel.py
```diff
@@ -339,3 +339,10 @@ def _write(self, data: Any):
             self._write_excel_with_single_sheet(
                 data.to_excel, self._path, index=False, header=properties[self._HAS_HEADER_PROPERTY] or None
             )
+
+    def _duplicate_data(self):
+        new_data_path = self._duplicate_data_file(self.id)
+        if hasattr(self._properties, "_entity_owner"):
+            del self._properties._entity_owner
+        self._properties[self._PATH_KEY] = new_data_path
+        return new_data_path
```
trgiangdo: Is this repetitive?
Can we put this in the parent `_FileDataNodeMixin` class and override it when necessary?

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -212,3 +222,16 @@ def _migrate_path(self, storage_type, old_path) -> str:
         if os.path.exists(old_path):
             shutil.move(old_path, new_path)
         return new_path
+
+    def _duplicate_data_file(self, id: str) -> Optional[str]:
+        if os.path.exists(self.path):
+            folder_path, base_name = os.path.split(self.path)
+            if base_name.startswith(self.__TAIPY_CLONED_PREFIX):
+                base_name = "".join(base_name.split("_")[5:])
+            new_base_path = os.path.join(folder_path, f"{self.__TAIPY_CLONED_PREFIX}_{id}_{base_name}")
```
trgiangdo: I have 2 questions.
What would be the new file name if a data node is duplicated twice?
Similarly, if a duplicated data node is duplicated (dn -> duplicated_dn -> duplicated duplicated_dn)?

taipy/core/scenario/_scenario_manager.py
```diff
@@ -521,3 +521,76 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
...
+        """
+        creation_date = creation_date or datetime.now()
+        cloned_scenario = cls._get(scenario)
+        cloned_scenario.id = cloned_scenario._new_id(cloned_scenario.config_id)
+
+        frequency = cls.__get_config(scenario).frequency
+        cycle = _CycleManagerFactory._build_manager()._get_or_create(frequency, creation_date) if frequency else None
+        cycle_id = cycle.id if cycle else None
+
+        # TODO: update sequences
```
trgiangdo: This is not handled yet right?

taipy/core/data/_data_manager.py
```diff
@@ -174,3 +179,39 @@ def _get_by_config_id(cls, config_id: str, version_number: Optional[str] = None)
...
+            cloned_dn._owner_id = cls._get_owner_id(cloned_dn._scope, cycle_id, scenario_id)
+            cloned_dn._parent_ids = set()
+
+            cloned_dn._duplicate_data()
+
+            cls._set(cloned_dn)
+            return cloned_dn
+
+    @classmethod
+    def _can_duplicate(cls, dn: DataNode) -> ReasonCollection:
```
trgiangdo: Should we also add a Reason here when the datanode is not a file-based one?

tests/core/test_taipy.py
```diff
@@ -869,3 +869,29 @@ def test_get_entities_by_config_id_in_multiple_versions_environment(self):
         assert len(tp.get_scenarios()) == 5
         assert len(tp.get_entities_by_config_id(scenario_config_1.id)) == 3
         assert len(tp.get_entities_by_config_id(scenario_config_2.id)) == 2
+
+    def test_can_duplicate(self):
+        dn_config = Config.configure_in_memory_data_node("dn", 10)
```
trgiangdo, line 874: If an `InMemory` data node is duplicated, won't it raise the `NotImplementedError`?

trgiangdo (commented):
I believe there are still a few areas in the code that can be improve

taipy/core/scenario/_scenario_manager.py
toan-quach: ah yep, I'm not sure about it at the moment as if we want to duplicate a sequence, we either:
- duplicate a scenario (which is weird)
- duplicate the sequence and store it within the same scenario (this is also weird as not sure about its benefit)

taipy/core/data/_file_datanode_mixin.py
toan-quach: yep so take `example.csv` as the file name, after you first duplicate the dn, it will create a new file `TAIPY_CLONED_new_dn_id_example.csv`. But if you want to duplicate this newly duplicated dn, another file will be created with the name pattern being `TAIPY_CLONED_another_new_dn_id_example.csv`. The `TAIPY_CLONED` prefix won't be repeated.

tests/core/test_taipy.py
toan-quach: yep, as we're currently focusing on file-based DN, I decided to ignore other cases and let `NotImplementedError` be raised

taipy/core/data/_data_manager.py
toan-quach: well eventually we want to be able to duplicate all DNs so it will only be applicable for now and will be removed later. So I think it won't bring much value for now as we will change it later anyway.

taipy/core/data/excel.py
toan-quach: hmmm I think it's a yes and no answer. The `self._properties` attribute doesn't exist in `_FileDataNodeMixin`, but since ExcelDN, etc. are implementing `_FileDataNodeMixin`, it can access it. But the syntax feels off for me, and since `_FileDataNodeMixin` is a Mixin, I'm not comfortable putting `self._properties` in it. (I understand that we do have similar things for `def path(self, value)` (setter) in `_FileDataNodeMixin`, but I'm still not sure if it's the best way to do it).

trgiangdo: I understand your point. However, other methods in the Mixin class doesn't seem to fit as well.

I think we did discuss this once. @jrobinAV Let us know what you think about this

taipy/core/data/_file_datanode_mixin.py
trgiangdo: I see. So what about duplicating a data node twice?

tests/core/test_taipy.py
trgiangdo: I think we should return a Reason, not raise an exception though.

toan-quach: ah yep, agree. But I left`NotImplementedError` there as we will fill in for other data nodes eventually. By then, Reason will become outdated. So to avoid confusion in later date, I just raise `NotImplementedError` for now :D

taipy/core/data/_file_datanode_mixin.py
toan-quach: then we will have `TAIPY_CLONED_dn_duplicated_id_1_example.csv` for the 1st dn, and for 2nd dn, we will have `TAIPY_CLONED_dn_duplicated_id_2_example.csv`

jrobinAV: I believe having both the prefix `TAIPY_CLONED` and the unique ID in the name is not necessary. If we can make the file name unique with the new ID, what is the purpose of keeping the prefix? We can probably propose something slightly better differentiating the case where the file name is generated by taipy or if it is provided by the user.

If the file of the initial dn is generated (`dn.is_generated`), we can simply generate a new name for the duplicate with the same function (`dn._build_path`). Otherwise, we can use your proposal without the prefix: `{new_id}_{base_name}`.

What do you think?

tests/core/data/test_csv_data_node.py
```diff
@@ -429,3 +430,29 @@ def check_data_is_positive(upload_path, upload_data):
 
         # The upload should succeed when check_data_is_positive() return True
         assert dn._upload(new_csv_path, upload_checker=check_data_is_positive)
+
+    def test_clone_data_file(self):
```
jrobinAV: rename to `test_duplicate_data_file`

taipy/core/scenario/_scenario_manager.py
jrobinAV: I don't understand.

In the new scenario, we want to create a new sequence made of the duplicates tasks.

taipy/core/data/excel.py
jrobinAV: Yes, this is debatable. I would personally do the same as the path setter for consistency.

taipy/core/data/_file_datanode_mixin.py
toan-quach: Well because I think the prefix will make the naming clear as to what it is, just the id is fine but we will have to state clearly in the doc that this is what we generated. While they can understand it just from the prefix without doc.

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -161,7 +167,8 @@ def _upload(self,
                 self.__logger.error(
                     f"Error with the upload checker `{upload_checker.__name__}` "
                     f"while checking `{up_path.name}` file for upload to the data "
-                    f"node `{self.id}`:") # type: ignore[attr-defined]
+                    f"node `{self.id}`:"
+                )  # type: ignore[attr-defined]
```
jrobinAV: The type ignore should apply to the `self.id` code.

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -212,3 +222,23 @@ def _migrate_path(self, storage_type, old_path) -> str:
         if os.path.exists(old_path):
             shutil.move(old_path, new_path)
         return new_path
+
+    def _duplicate_data_file(self, id: str) -> Optional[str]:
```
jrobinAV:
```suggestion
    def _duplicate_file(self, id: str) -> Optional[str]:
```

tests/core/data/test_csv_data_node.py
```diff
@@ -429,3 +430,29 @@ def check_data_is_positive(upload_path, upload_data):
 
         # The upload should succeed when check_data_is_positive() return True
         assert dn._upload(new_csv_path, upload_checker=check_data_is_positive)
+
+    def test_duplicate_data_file(self):
```
jrobinAV:
```suggestion
    def test_duplicate_data(self):
```

tests/core/data/test_csv_data_node.py
```diff
@@ -429,3 +430,29 @@ def check_data_is_positive(upload_path, upload_data):
...
+        path = os.path.join(pathlib.Path(__file__).parent.resolve(), "data_sample/example.csv")
+        dn = CSVDataNode("foo", Scope.SCENARIO, properties={"path": path, "exposed_type": "pandas"})
+        _DataManager._set(dn)
+
+        read_data = dn.read()
+        assert read_data is not None
+
+        old_path = dn.path
+        new_file_path = str(dn._duplicate_data())
+        assert filecmp.cmp(path, new_file_path)
```
jrobinAV:
```suggestion
        assert filecmp.cmp(old_path, new_file_path)
```

jrobinAV: What about adding an explicit assert to ensure the new path is correct?
```suggestion
        assert new_file_path == "data_sample/TAIPY_DUPLICATED_example.csv"
        assert filecmp.cmp(path, new_file_path)
```

jrobinAV (dismissed):

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -136,20 +139,23 @@ def _upload(self,
...
             reasons._add_reason(self.id, DataNodeEditInProgress(self.id))  # type: ignore[attr-defined]
             return reasons
 
         up_path = pathlib.Path(path)
         try:
             upload_data = self._read_from_path(str(up_path))
         except Exception as err:
-            self.__logger.error(f"Error uploading `{up_path.name}` to data "
-                                f"node `{self.id}`:")  # type: ignore[attr-defined]
+            self.__logger.error(f"Error uploading `{up_path.name}` to data " f"node `{self.id}`:")  # type: ignore[attr-defined]
```
jrobinAV:
```suggestion
            self.__logger.error(f"Error uploading `{up_path.name}` to data "
                                f"node `{self.id}`:")  # type: ignore[attr-defined]
```

jrobinAV (dismissed):
I believe I know what's bugging me. 
It's a matter of responsibility. 

In short, I believe the data duplication should not be the responsibility of the data node.
In my mind, the data node class only knows itself and does not have a global view of all the data nodes, or scenarios. Duplicating data should be the responsibility of classes that have an overview of the situation. Either the managers or new classes dedicated to the data duplication.

Give me some time to propose something else

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -171,9 +179,12 @@ def _upload(self,
...
 
-        self.track_edit(timestamp=datetime.now(),  # type: ignore[attr-defined]
-                        editor_id=editor_id,
-                        comment=comment, **kwargs)
+        self.track_edit(  # type: ignore[attr-defined]
+            timestamp=datetime.now(),
+            editor_id=editor_id,
+            comment=comment,
+            **kwargs,
+        )
```
jrobinAV:
```suggestion
       self.track_edit(timestamp=datetime.now(),  # type: ignore[attr-defined]
                        editor_id=editor_id,
                        comment=comment, **kwargs)
```

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -171,7 +179,7 @@ def _upload(self,
 
         shutil.copy(up_path, self.path)
 
-        self.track_edit(timestamp=datetime.now(),  # type: ignore[attr-defined]
+       self.track_edit(timestamp=datetime.now(),  # type: ignore[attr-defined]
```
jrobinAV:
```suggestion
        self.track_edit(timestamp=datetime.now(),  # type: ignore[attr-defined]
```

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -212,3 +220,23 @@ def _migrate_path(self, storage_type, old_path) -> str:
...
+
+            if os.path.isdir(self.path):
+                shutil.copytree(self.path, new_base_path)
+            else:
+                shutil.copy(self.path, new_base_path)
+
+            if hasattr(self._properties, "_entity_owner"):  # type: ignore[attr-defined]
+                del self._properties._entity_owner  # type: ignore[attr-defined]
+            self._properties[self._PATH_KEY] = new_base_path  # type: ignore[attr-defined]
+
```
toan-quach: I think we can add it back right after setting all the values to the `self._properties` here, the new `_entity_owner` should be the new data node id I suppose

jrobinAV (dismissed):
Self approved

taipy/core/data/_data_duplicator.py
```diff
@@ -0,0 +1,44 @@
...
+        """Duplicate the src data to the data of the destination data node.
+
+        Parameters:
+            dest (DataNode): The destination data node.
+
+        Raises:
+            NotImplementedError: If the data node type is not supported yet.
+            NotImplementedError: If the source and destination data nodes have different storage types.
+        """
+        if isinstance(self.src, _FileDataNodeMixin):
```
trgiangdo, line 40:
```suggestion
        if self.can_duplicate():
```

taipy/core/data/_data_duplicator.py
```diff
@@ -0,0 +1,44 @@
...
+            if self.src.storage_type() != dest.storage_type():
+                raise NotImplementedError("Source and destination data nodes must have the same storage type.")
```
trgiangdo: I don't think we should raise the `NotImplementedError` here, since this doesn't make sense for implementation.

Maybe a custom exception like `InvalidDestinationDataNode`, or the existing `InvalidDataNodeType`

taipy/core/data/_data_duplicator.py
```diff
@@ -0,0 +1,44 @@
...
+            self.src._duplicate_file(dest)
+        else:
+            raise NotImplementedError("Data node type not supported yet.")
```
trgiangdo:
```suggestion
            raise NotImplementedError(f"Data node type '{self.src.storage_type()}' is not supported for duplication yet.")
```

taipy/core/data/_file_datanode_mixin.py
```diff
@@ -212,3 +220,15 @@ def _migrate_path(self, storage_type, old_path) -> str:
         if os.path.exists(old_path):
             shutil.move(old_path, new_path)
         return new_path
+
+    def _duplicate_file(self, dest: DataNode):
+        if os.path.exists(self._path):
+            folder_path, base_name = os.path.split(self._path)
```
trgiangdo, line 226: I don't get this. Look like we have a white space in the path.

Shouldn't we split by the `os.path.sep`?

trgiangdo (commented):
This is much cleaner.

I have a few minor suggestions

taipy/core/data/_data_duplicator.py
jrobinAV: You are right.

taipy/core/data/_file_datanode_mixin.py
jrobinAV: That's what it does. Its a path split, not a string split.

jrobinAV (dismissed):

jrobinAV (dismissed):

taipy/core/data/_file_datanode_mixin.py
trgiangdo: Oh that's right. Perfect then :D

trgiangdo (dismissed):
LGTM

PR #2439, opened 2025-02-05
fix annoying bits

FredLL-Avaiga:
- exception swallowed but event manager
- dict notation for state

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
fix annoying bits

### DIFF
--- a/taipy/gui/state.py
+++ b/taipy/gui/state.py
@@ -149,6 +149,10 @@ class State(SimpleNamespace, metaclass=ABCMeta):
         """
         self._gui.set_favicon(favicon_path, self)
 
+    @abstractmethod
+    def __getitem__(self, key: str):
+        return self
+
 
 class _GuiState(State):
     __gui_attr = "_gui"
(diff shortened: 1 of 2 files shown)


taipy/gui/state.py
```diff
@@ -149,6 +149,10 @@ def set_favicon(self, favicon_path: t.Union[str, Path]):
         """
         self._gui.set_favicon(favicon_path, self)
 
+    @abstractmethod
+    def __getitem__(self, key: str):
+        return self
```
FabienLelaquais: Whether ` __getitem__()` is abstract (then you don't want the default implementation),
or it is not (then you don't want the `@abstractmethod` decorator)
.

FabienLelaquais (dismissed):
👍

taipy/gui/state.py
FredLL-Avaiga: I know but you have to write something and returning self allows the tools to determine that it is returning a State

FredLL-Avaiga: changed

FabienLelaquais (approved):
Perfect!

dinhlongviolin1 (approved):

PR #2440, opened 2025-02-07
Refactor - Allow expanding the TYPE_PATTERN by the _BaseSerializer._SERIALIZABLE_TYPES list

trgiangdo:

- [x] Refactor
- [x] Feature

## Description

This PR creates the `_BaseSerializer._SERIALIZABLE_TYPES` to allow expanding the serializable types when stringify and pythonify the Config.

### DIFF
--- a/taipy/common/config/__init__.py
+++ b/taipy/common/config/__init__.py
@@ -111,6 +111,8 @@ def _inject_section(
     configuration_methods: List[tuple],
     add_to_unconflicted_sections: bool = False,
 ):
+    from ._serializer._base_serializer import _BaseSerializer
+
     Config._register_default(default)
 
     if issubclass(section_clazz, UniqueSection):
@@ -125,6 +127,10 @@ def _inject_section(
 
     for exposed_configuration_method, configuration_method in configuration_methods:
         setattr(Config, exposed_configuration_method, configuration_method)
+
+    for type_to_register in section_clazz._types_to_register():
+        if type_to_register._type_identifier() not in _BaseSerializer._SERIALIZABLE_TYPES:
+            _BaseSerializer._SERIALIZABLE_TYPES.append(type_to_register._type_identifier())
 
 
 @_config_doc_for_method
--- a/taipy/common/config/__init__.py
+++ b/taipy/common/config/__init__.py
@@ -46,7 +46,6 @@ and attributes to configure the Taipy application and retrieve the configuration
 """
 
 import os
-from inspect import signature
 from typing import List
 
 from ._init import Config
@@ -127,10 +126,6 @@ def _inject_section(
 
     for exposed_configuration_method, configuration_method in configuration_methods:
         setattr(Config, exposed_configuration_method, configuration_method)
-
-    for type_to_register in section_clazz._types_to_register():
-        if type_to_register._type_identifier() not in _BaseSerializer._SERIALIZABLE_TYPES:
-            _BaseSerializer._SERIALIZABLE_TYPES.append(type_to_register._type_identifier())
 
 
 @_config_doc_for_method
--- a/taipy/common/config/__init__.py
+++ b/taipy/common/config/__init__.py
@@ -54,7 +54,7 @@ from .section import Section
 from .unique_section import UniqueSection
 
 
-def __write_method_to_doc(configuration_methods):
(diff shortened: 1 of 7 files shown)


taipy/common/config/_serializer/_base_serializer.py
```diff
@@ -29,6 +29,19 @@ class _BaseSerializer(object):
...
+        "bool",
+        "str",
+        "int",
+        "float",
+        "datetime",
+        "timedelta",
+        "function",
+        "class",
+        "SCOPE",
+        "FREQUENCY",
```
jrobinAV: This must be removed and dynamically added from the core package. For instance, it could be 
declared in a `_SUB_TYPES ` variable implemented in the `Section` class, and updated dynamically by the `_inject_section` function (in `common/config/__init__.py`) when registering the Section.

jrobinAV (changes requested):

taipy/common/config/_serializer/_base_serializer.py
```diff
@@ -119,10 +132,7 @@ def _pythonify(cls, val):
         match = re.fullmatch(_TemplateHandler._PATTERN, str(val))
         if not match:
             if isinstance(val, str):
-                TYPE_PATTERN = (
-                    r"^(.+):(\bbool\b|\bstr\b|\bint\b|\bfloat\b|\bdatetime\b||\btimedelta\b|"
-                    r"\bfunction\b|\bclass\b|\bSCOPE\b|\bFREQUENCY\b|\bSECTION\b)?$"
-                )
+                TYPE_PATTERN = r"^(.+):(" + r"\b|\b".join(cls._SERIALIZABLE_TYPES) + r")?$"
```
jrobinAV: It does not look like it generates the same string:
What about:

```suggestion
                TYPE_PATTERN = r"^(.+):(\b" + r"\b|\b".join(cls._SERIALIZABLE_TYPES) + r"\b)?$"
```

trgiangdo: You're right, the "\b" at the beginning and ending were missing

taipy/common/config/__init__.py
```diff
@@ -111,6 +111,8 @@ def _inject_section(
     configuration_methods: List[tuple],
     add_to_unconflicted_sections: bool = False,
 ):
+    from ._serializer._base_serializer import _BaseSerializer
```
trgiangdo, line 113: This import command is inside the method because this is an `__init__.py` file and we don't want to expose the `_BaseSerializer`

jrobinAV (dismissed):

taipy/common/config/__init__.py
```diff
@@ -126,6 +128,10 @@ def _inject_section(
     for exposed_configuration_method, configuration_method in configuration_methods:
         setattr(Config, exposed_configuration_method, configuration_method)
 
+    for type_to_register in section_clazz._types_to_register():
+        if type_to_register._type_identifier() not in _BaseSerializer._SERIALIZABLE_TYPES:
+            _BaseSerializer._SERIALIZABLE_TYPES.append(type_to_register._type_identifier())
```
jrobinAV: IS it necessary? I believe the `Config._register_default(default)` does the job already. no?

trgiangdo: I don't see anything in the `_register_default()` method that does this.

Anyway, removing this would make the serialization tests fail, so I believe it's necessary

jrobinAV: Oh, I understand now. Indeed, the `Config._register_default` calls the `_BaseSerializer._register` that registers the types, but does not update the `_BaseSerializer._SERIALIZABLE_TYPES` variable but the `_registered_types`.

I believe we can update the variable in the `_BaseSerializer._register` method.

tests/common/config/utils/serializable_object_for_test.py
jrobinAV, line 1: I don't see where it is used.

jrobinAV (approved):

tests/common/config/utils/serializable_object_for_test.py
trgiangdo: it is used in the SectionForTest as the `_types_to_register()` to check if registering a section will also include that type.

The reason is because the `Scope` and `Frequency` are automatically registered at conftest.py so we can not test those properly here.

PR #2443, opened 2025-02-11
Chart animation

### LINKED ISSUE #1773: Handle Plotly animation
### Description

The goal would be to handle [Plotly animation](https://plotly.com/python/animations/) inside Taipy or find a workaround for it.

This was asked by a user.

### Technical proposal 1 [DOES NOT WORK]

We seem to cover basic use cases with the addition of an `animation` property that would describe if and how animation would take place.
This property must indicate which dataset columns are animated, and animation properties such as the transition easing and speed.

Use case: I want to change which of the dataset column is represented, animating from the initial one to the next.
Data definition is:
```python
     x = [0, 1, 2, 3]
     y1 = [0, 1, 4, 9]
     y2 = [0, 1, 8, 27]
```
Animated column value: `y_values="y1"`

Chart definition is:
```
      x=x 
      y={y_values}  (so it can be dynamically updated)
      animation={ on="y", duration=1000, easing="linear" }
```
Meaning:
  We are initially showing 'y1' on the 'y' axis.
  If 'y' changes to a new column, animate the chart with the indicated parameters.
Now I set 'y_values' to "y2", the chart can animate from the values in 'y1' to the values in 'y2'.

#### Why it does not work

The problem is `y={y_values}` that presumes that `y` is dynamic, which it is not.
Although this would be the panacea, it is not how Taipy GUI charts work, and it would be a mess to change that.

### Technical proposal 2

Data definition is:
```python
data = {
     "x": [0, 1, 2, 3],
     "y": [0, 1, 4, 9]
}
```

We will add an `animation_data` dynamic property that must be a dataset with only the columns that are animated:
```python
data = {
     "y": [0, 1, 8, 27]
}
```

When the chart component has this property set, it loads it all, and receives all the columns that are impacted.
Because the "y" columns appears in both datasets, the 'from' and 'to' frames can be built from the two column values.
And animation is performed.

When a new `animation_data` is sent (with new data, in the same column(s)), we move the 'to' frame to the 'from' frame and recompute the new 'to' frame from the values in the `animation_data` column(s).

From the user stand point, all that needs to be done is update the `animation_data` content. The animation is carried immediately.
There should also be a `animation_config` (or `animation` for short) dictionary (static) property that can tune the animation parameters such as speed or type.

### Acceptance Criteria

### Code of Conduct

namnguyen20999:

- [ ] Refactor
- [x] Feature (see #1773).
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description

```
from taipy.gui import Gui

# Original data
data = {
    "x": [
        "2024-01-01 00:00:00", "2024-01-01 01:00:00", "2024-01-01 02:00:00",
        "2024-01-01 03:00:00", "2024-01-01 04:00:00", "2024-01-01 05:00:00",
        "2024-01-01 06:00:00", "2024-01-01 07:00:00", "2024-01-01 08:00:00",
        "2024-01-01 09:00:00"
    ],
    "y1": [1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
}

layout = {
    "title": "Animation of Weather Data",
    "xaxis": {"title": "Altitude (m)"},
    "yaxis": {"title": "Value"},
    "updatemenus": [
        {
            "type": "buttons",
            "showactive": False,
            "buttons": [
                {
                    "method": "animate",
                    "args": [
                        ["to"],
                        {
                            "fromcurrent": True,
                            "mode": "immediate",
                            "frame": {"duration": 500, "redraw": True},
                            "transition": {"duration": 2000, "easing": "linear"}
                        }
                    ],
                    "label": "Play"
                }
            ]
        }
    ],
}

animation_data = {
    "y1": [v + 1 for v in data["y1"]],
}


def change_data_inc(state):
    state.animation_data["y1"] = [value + 1 for value in state.animation_data["y1"]]

def change_data_dec(state):
    state.animation_data["y1"] = [value - 1 for value in state.animation_data["y1"]]


# Define the page layout
page = """
<|{data}|chart|mode=lines+markers|x=x|y=y1|layout={layout}|animation_data={animation_data}|>
<|Increase|button|on_action=change_data_inc|>
<|Decrease|button|on_action=change_data_dec|>
"""

if __name__ == "__main__":
    Gui(page).run(title="Animation on", port=8081, use_reloader=True)

```

- Related Issue #
- Closes #

## How to reproduce the issue

_Please replace this line with instructions on how to reproduce the issue or test the feature._

## Other branches or releases that this needs to be backported
_Describe which projects this change will impact and that needs to be backported._

## Checklist
_We encourage you to keep the code coverage percentage at 80% and above._

- [ ] This solution meets the acceptance criteria of the related issue.
- [ ] The related issue checklist is completed.
- [ ] This PR adds unit tests for the developed code. If not, why?
- [ ] End-to-End tests have been added or updated. If not, why?
- [ ] The documentation has been updated, or a dedicated issue created. (If applicable)
- [ ] The release notes have been updated? (If applicable)


frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -155,30 +169,30 @@ const getDecimatorsPayload = (
...
-                      : {
-                            xAxis: getAxis(traces, i, columns, 0),
-                            yAxis: getAxis(traces, i, columns, 1),
-                            zAxis: getAxis(traces, i, columns, 2),
-                            chartMode: modes[i],
-                        }
-              ),
-              relayoutData: relayoutData,
-          }
+            width: plotDiv?.clientWidth,
```
FabienLelaquais: Strange you have a 2-space indentation here and 4 later in the changes....

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -574,8 +618,8 @@ const Chart = (props: ChartProp) => {
                     ret.z = baseZ;
                 }
             }
-            // Hack for treemap charts: create a fallback 'parents' column if needed
-            // This works ONLY because 'parents' is the third named axis
+                // Hack for treemap charts: create a fallback 'parents' column if needed
```
FabienLelaquais: No indent here (and l622)

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -788,9 +861,89 @@ const Chart = (props: ChartProp) => {
...
         },
-        [getRealIndex, dispatch, updateVars, propagate, props.onChange, config.traces.length, module]
+        [getRealIndex, dispatch, updateVars, propagate, props.onChange, config.traces.length, module],
     );
 
+    useEffect(() => {
+        if (!dataPl.length || !animationDataValues) return;
+
+        const toFramesData = dataPl.map((trace) => {
+            const isYaxisAnimated = Object.prototype.hasOwnProperty.call(animationDataValues, (trace as ScatterData).name);
```
FabienLelaquais:
```suggestion
            const isYaxisAnimated = Object.prototype.hasOwn(animationDataValues, (trace as ScatterData).name);
```
`hasOwn()` is 'safer' than `hasOwnProperty()` and I don't understand why you need the `call()`.

taipy/gui/viselements.json
```diff
@@ -502,6 +502,11 @@
                         "default_value": "\"scatter\"",
                         "doc": "Chart type.<br/>See the Plotly <a href=\"https://plotly.com/javascript/reference/\">chart type</a> documentation for more details."
                     },
+                    {
+                        "name": "animation_data",
+                        "type": "dynamic(dict[str, Any])",
+                        "doc": "Column name and its associated data for the animation."
```
```suggestion
                        "doc": "A dictionary representing updated values for the dataset defined in <i>data</i>.<br/>Each key in <i>animation_data</i> must exist in <i>data</i>. Changing this property triggers an animation of the chart, showing the changes between the original (<i>data</i>) and updated (<i>animation_data</i>) values."
```

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -788,9 +861,89 @@ const Chart = (props: ChartProp) => {
...
+                console.error("Error adding frames:", error);
+            }
+        };
+
+        // Debounce or delay to prevent overlapping calls
+        animationTimeout = setTimeout(() => {
+            if (plotRef.current) {
+                addFramesAsync().catch(console.error);
+            }
+        }, 300);
```
FabienLelaquais: I suspect you can make this delay shorter (like 100 ms).

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -788,9 +861,89 @@ const Chart = (props: ChartProp) => {
...
+                await Plotly.react(plotElement, frames.from.data, layout, {});
+                if (plotRef.current && frames?.to?.data && frames.to.traces.length > 0) {
+                    await Plotly.animate(
+                        plotElement,
+                        {
+                            data: frames.to.data,
+                            traces: frames.to.traces,
+                            layout: {},
+                        },
+                        {
```
FabienLelaquais: So that would be the same as in the initialization right?

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -714,19 +762,44 @@ const Chart = (props: ChartProp) => {
...
+            const runAnimation = async () => {
+                if (plotRef.current && frames?.to?.data && frames.to.traces.length > 0) {
+                    await Plotly.animate(
+                        plotRef.current,
+                        {
+                            data: frames.to.data,
+                            traces: frames.to.traces,
+                            layout: layout,
+                        },
+                        {
```
FabienLelaquais: You may want to move this dict to some 'default' constant value. 
Then we could modify it with the setting of the *animation* prop.
At least the default values don't show up in the middle of the code.
(and reuse it later)

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -588,6 +632,10 @@ const Chart = (props: ChartProp) => {
             ret.xaxis = config.xaxis[idx];
             ret.yaxis = config.yaxis[idx];
             ret.hovertext = getValue(datum, config.labels, idx, true);
+            ret.meta = {
+                xAxisName: config.traces[idx][0],
+                yAxisName: config.traces[idx][1],
```
FabienLelaquais, line 662: I'm... not sure how this works in multi-trace, actually.

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -788,9 +861,89 @@ const Chart = (props: ChartProp) => {
                 }
             }
         },
-        [getRealIndex, dispatch, updateVars, propagate, props.onChange, config.traces.length, module]
+        [getRealIndex, dispatch, updateVars, propagate, props.onChange, config.traces.length, module],
     );
 
+    useEffect(() => {
+        if (!dataPl.length || !animationDataValues) return;
```
FabienLelaquais: May be we want to clean the 'frame data' at this point?
One way not to animate any longer...

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -788,9 +861,89 @@ const Chart = (props: ChartProp) => {
...
+
+    useEffect(() => {
+        if (!plotRef.current || !frames.to || !frames.to.data?.length) return;
+
+        const plotElement = plotRef.current as unknown as PlotlyHTMLElement;
+        if (!plotElement || !plotElement.data) return;
+
+        let animationTimeout: NodeJS.Timeout | null = null;
+
+        const addFramesAsync = async () => {
```
FabienLelaquais:
```suggestion
        const animate= async () => {
```
Isn't exactly what this thing does?

frontend/taipy-gui/src/components/Taipy/Chart.tsx
FabienLelaquais: I'm also a bit worried about the use of `ScatterData` here...
If this means not all chart types can use the feature, so be it.
But then we want to ignore the animation data (and document the limitation).

FabienLelaquais (dismissed):
Nice!

taipy/gui/viselements.json
```diff
@@ -502,6 +502,11 @@
                         "default_value": "\"scatter\"",
                         "doc": "Chart type.<br/>See the Plotly <a href=\"https://plotly.com/javascript/reference/\">chart type</a> documentation for more details."
                     },
+                    {
+                        "name": "animation_data",
+                        "type": "dynamic(dict[str, Any])",
```
FredLL-Avaiga: why is this not TaipyData ?

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -60,6 +72,8 @@ interface ChartProp extends TaipyActiveProps, TaipyChangeProps {
     defaultConfig: string;
     config?: string;
     data?: Record<string, TraceValueType>;
+    animationData?: string;
+    defaultAnimationData?: string;
```
FredLL-Avaiga: default shouldn't be needed with TaipyData

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -205,6 +221,13 @@ interface PlotlyDiv extends HTMLDivElement {
     };
 }
 
+interface ExtendedPlotData extends Plotly.PlotData {
```
FredLL-Avaiga: why not import PlotData ?

FredLL-Avaiga (changes requested):
should use TaipyData type

frontend/taipy-gui/src/components/Taipy/Chart.tsx
namnguyen20999: I extended the Plotly.PlotData interface to add the meta property, since it’s not included by default

FredLL-Avaiga: you can import PlotData and extend it, no ?

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -543,9 +550,35 @@ const Chart = (props: ChartProp) => {
         props.figure,
     ]);
 
+    const updateDateVarsName = useMemo(() => {
```
FabienLelaquais: `updateDataVarsName`?

FabienLelaquais (dismissed):
👌

frontend/taipy-gui/src/components/Taipy/Chart.tsx
namnguyen20999: silly mistakes 😂

frontend/taipy-gui/webpack.config.js
FredLL-Avaiga, line 1: no modification I guess ?

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -22,20 +30,27 @@ import { nanoid } from "nanoid";
...
     ModeBarButtonAny,
     PlotDatum,
+    PlotData,
+    PlotlyHTMLElement,
     PlotMarker,
     PlotRelayoutEvent,
     PlotSelectionEvent,
+    ScatterData,
     ScatterLine,
+    AnimationOpts,
```
FredLL-Avaiga: ordering the imports would nice

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -488,12 +550,52 @@ const Chart = (props: ChartProp) => {
         props.figure,
     ]);
 
+    const updateDateVarsName = useMemo(() => {
```
FredLL-Avaiga: There's a function for this ...
 we only need the backend variable name for animationData => use `getUpdateVar(updateVars, "animationData")` from utils.ts.

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -488,12 +550,52 @@ const Chart = (props: ChartProp) => {
...
+                    acc[key.trim()] = value.trim();
+                }
+                return acc;
+            }, {} as Record<string, string>)
+            : {};
+    }, [updateVars]);
+
+    useEffect(() => {
+        if (animationData?.__taipy_refresh) {
+            dispatch(createRequestDataUpdateAction(
```
FredLL-Avaiga: is it not better/easier to use `createRequestChartUpdateAction` ?
should we specify a list of columns ?

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -788,9 +897,72 @@ const Chart = (props: ChartProp) => {
...
+                }
+                return prevFrames;
+            });
+            return;
+        }
+
+        const plotElement = plotRef.current as unknown as PlotlyHTMLElement;
+        if (!plotElement || !plotElement.data) return;
+
+        let animationTimeout: NodeJS.Timeout | null = null;
```
FredLL-Avaiga: bad type NodeJS.Timeout should be number in a browser

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -788,9 +897,72 @@ const Chart = (props: ChartProp) => {
...
+
+        // Debounce or delay to prevent overlapping calls
+        animationTimeout = setTimeout(() => {
```
FredLL-Avaiga: why not `const animationTimeout` ? (ie why 2 lines ?)

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -788,9 +897,72 @@ const Chart = (props: ChartProp) => {
...
+                runAnimation().catch(console.error);
+            }
+        }, 100);
+
+        return () => {
+            if (animationTimeout) {
+                clearTimeout(animationTimeout);
+            }
+        };
+    }, [layout, frames.to, frames.from, runAnimation]);
```
FredLL-Avaiga: are `layout` and `frames.from` really needed here ?

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -308,26 +352,44 @@ const Chart = (props: ChartProp) => {
...
         onClick,
+        animationData,
     } = props;
     const dispatch = useDispatch();
     const [selected, setSelected] = useState<number[][]>([]);
-    const plotRef = useRef<HTMLDivElement>(null);
+    const plotRef = useRef<HTMLDivElement | null>(null);
+    const plotlyRef = useRef<PlotlyObject | null>(null);
     const [dataKeys, setDataKeys] = useState<string[]>([]);
+    const [frames, setFrames] = useState<{ from: Frame; to: Frame }>({
```
FredLL-Avaiga: Would it be a good idea to split this into 2 `useState` ?
toFrame & fromFrame ?

taipy/gui/viselements.json
FredLL-Avaiga: change to TaipyData (or the type for chart.data)

FredLL-Avaiga (changes requested):
A few minor changes

frontend/taipy-gui/webpack.config.js
namnguyen20999: I've removed the css-loader in my local, it shouldn't be there in the next commit

frontend/taipy-gui/src/components/Taipy/Chart.tsx
namnguyen20999: In this case we going to grab all of the column so I don't think we need to specify a list of columns

FredLL-Avaiga: I think we could request the list of columns of the main data, no ?

taipy/gui/viselements.json
```diff
@@ -504,7 +504,7 @@
                     },
                     {
                         "name": "animation_data",
-                        "type": "dynamic(dict[str, Any])",
+                        "type": "chart.data",
```
FabienLelaquais: This should be 'dynamic(Any)': it is used when generating the Page Builder API, and must be a valid Python type.

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -82,6 +82,8 @@ const Metric = (props: MetricProps) => {
     const hover = useDynamicProperty(props.hoverText, props.defaultHoverText, undefined);
     const theme = useTheme();
 
+    console.log(props.title);
```
FabienLelaquais: Remove

frontend/taipy-gui/src/components/Taipy/Metric.tsx
```diff
@@ -191,8 +193,10 @@ const Metric = (props: MetricProps) => {
             layout.template = template;
         }
 
-        if (props.title) {
-            layout.title = props.title;
+        if (typeof props.title !== "object") {
```
FabienLelaquais: How did this situation happen?

FabienLelaquais (dismissed):
👌

frontend/taipy-gui/src/components/Taipy/Metric.tsx
namnguyen20999: They just added a bunch of things inside the title recently: https://plotly.com/javascript/reference/layout/#layout-title

namnguyen20999: Users can still access all these properties by defining their own title object within the layout prop, rather than using our title prop

FabienLelaquais: I see. Thx.

FabienLelaquais (dismissed):
👍

frontend/taipy-gui/src/components/Taipy/Metric.tsx
FredLL-Avaiga: but props.title is typed as string in the backend ?

frontend/taipy-gui/src/components/Taipy/Expandable.spec.tsx
```diff
@@ -24,7 +24,7 @@ describe("Expandable Component", () => {
     it("renders", async () => {
         const { getByText } = render(<Expandable title="foo">bar</Expandable>);
         const elt = getByText("foo");
-        expect(elt.tagName).toBe("DIV");
+        expect(elt.tagName).toBe("SPAN");
```
FredLL-Avaiga: most of the times, the test was passing with DIV
And if it fails because of SPAN, re-run the test and it succeed

frontend/taipy-gui/src/components/Taipy/Metric.tsx
FredLL-Avaiga: I'd rather go for `if (typeof props.title == "string")`
and let the issue on title handle the rest :
With an `else if (props.title)`
and change the backend type

FredLL-Avaiga (dismissed):

frontend/taipy-gui/src/components/Taipy/Expandable.spec.tsx
namnguyen20999: I ran the test locally with DIV, but it never passed

FredLL-Avaiga: well I'd prefer it to pass on the CI than on your mac :-)

namnguyen20999: For the title below I'd go for this way:

```
        if (props.title) {
            layout.title = {
                text: props.title
            }
        }
```

We no longer need to change the backend type. If we follow your approach, we’d have to create a `string_or_dict `type. Instead, why not let users stick with the layout prop if they want to customize their title?

frontend/taipy-gui/src/components/Taipy/Metric.tsx
namnguyen20999: For the title below I'd go for this way:

```
        if (props.title) {
            layout.title = {
                text: props.title
            }
        }
```

We no longer need to change the backend type. If we follow your approach, we’d have to create a `string_or_dict `type. Instead, why not let users stick with the layout prop if they want to customize their title?

FredLL-Avaiga:
demo

```py
from math import ceil, cos

from taipy.gui import Gui

shapes=["Sine", "Square"]
shape = shapes[0]
cos_data = [cos(i/6) for i in range(100)]
square_data = [1 if ceil(i/24) % 2 == 0 else -1 for i in range(100)]
# Original data
data = {
    "x": range(100),
    "y": cos_data,
}

animation_data = None

is_cos= True


def change_data(state):
    state.animation_data = {"y": cos_data if state.shape == shapes[0] else square_data}



# Define the page layout
page = """
<|{shape}|toggle|on_change=change_data|lov={shapes}|>
<|{data}|chart|mode=lines+markers|x=x|y=y|animation_data={animation_data}|>
"""

if __name__ == "__main__":
    Gui(page).run(title="1773 Animation on")

```

frontend/taipy-gui/src/components/Taipy/Chart.tsx
```diff
@@ -789,12 +910,77 @@ const Chart = (props: ChartProp) => {
...
+            if (timer) {
+                clearTimeout(timer);
+            }
+        };
+    }, [runAnimation, toFrame.data?.length]);
+
     return render ? (
         <Tooltip title={hover || ""}>
-            <Box id={id} className={`${className} ${getComponentClassName(props.children)}`} ref={plotRef}>
+            <Box id={"graph"} className={`${className} ${getComponentClassName(props.children)}`} ref={plotRef}>
```
FabienLelaquais: So where `props.id` is not used anymore in for the generated component?

taipy/gui/viselements.json
```diff
@@ -508,6 +508,11 @@
                         "default_value": "\"scatter\"",
                         "doc": "Chart type.<br/>See the Plotly <a href=\"https://plotly.com/javascript/reference/\">chart type</a> documentation for more details."
                     },
+                    {
+                        "name": "animation_data",
+                        "type": "dynamic(Any)",
+                        "doc": "A dictionary representing updated values for the dataset defined in <i>data</i>.<br/>Each key in <i>animation_data</i> must exist in <i>data</i>. Changing this property triggers an animation of the chart, showing the changes between the original (<i>data</i>) and updated (<i>animation_data</i>) values.<br/> This animation functionality currently works only with <i>Scatter</i> traces. If you’re using other trace types, the animation may not behave as expected."
```
```suggestion
                        "doc": "A dictionary representing updated values for the dataset defined in <i>data</i>.<br/>Each key in <i>animation_data</i> must exist in <i>data</i>. Changing this property triggers an animation of the chart, showing the changes between the original (<i>data</i>) and updated (<i>animation_data</i>) values.<br/> This animation functionality currently works only with <i>scatter</i> trace types. If you are using other trace types, the animation may not behave as expected."
```

FabienLelaquais (dismissed):
👌 niiiice!

frontend/taipy-gui/src/components/Taipy/Chart.tsx
FredLL-Avaiga: I missed that one ...

FredLL-Avaiga:
Let the dev, change the base ...

```py
from math import ceil, cos

from taipy.gui import Gui

shapes=["Sine", "Square"]
shape = shapes[0]
cos_data = [cos(i/6) for i in range(100)]
square_data = [1 if ceil(i/24) % 2 == 0 else -1 for i in range(100)]
# Original data
data = {
    "x": range(100),
    "y": cos_data,
}

animation_data = None

is_cos= True


def change_data(state):
    state.data["y"] = cos_data if state.shape == shapes[1] else square_data
    state.refresh("data")
    state.animation_data = {"y": cos_data if state.shape == shapes[0] else square_data}



# Define the page layout
page = """
<|{shape}|toggle|on_change=change_data|lov={shapes}|>
<|{data}|chart|mode=lines+markers|x=x|y=y|animation_data={animation_data}|>
"""

if __name__ == "__main__":
    Gui(page).run(title="1773 Animation on")

```

FabienLelaquais (dismissed):
You lost me in the Effect hook on animationData...
Well done, I'm sure!

FabienLelaquais (approved):
👌

PR #2447, opened 2025-02-13
Extension library elements access through the Page Builder API

FabienLelaquais:
Add the element entry points to the library package module (and not the library module).

So the way to access library elements becomes:
```python
from taipy.gui import Gui

import extension_library

...

Gui.add_library(extension_library.Library())

with tgb.Page() as page:
    extension_library.element_name(...)

if __name__ == "__main__":
    # Run main app
    Gui(page=page).run()

```


dinhlongviolin1 (approved):
Lgtm

PR #2452, opened 2025-02-18
allow front-end synchronization

### LINKED ISSUE #2344: [🐛 BUG] De-synchronized front-ends and  back-end
### What went wrong? 🤔
Front-ends get de-synchronized when re-running an application.

This is more important than it looks like, as it happens frequently and gives users the wrong impression that their application has a bug. 

For example, it happens frequently during an application's development phase, especially with the default behavior of Gui().run(), which opens a new tab without closing the previous tabs.

It also happens when re-deploying a live application, causing the wrong error messages at a very bad time.

### Reproduce the situation

The problem arises when:

- We run an application.
- We access the application on a browser tab.
- We shut down the application without closing the tab.
- We re-run the application (or another application on the same port).
- We use the same old tab that does not correspond to the new back-end application. This may create errors, warnings, or inconsistencies.

### Expected Behavior

De-synchronized front-ends should be aware that the back-end version they are connected to is incorrect. The front-end application should automatically display a message to let the user know he/she should refresh the browser.

### Solution Proposed

What about generating a unique identifier each time we run the server? It could represent a `run_identifier`. Each of the clients (the tabs) could know this unique `run_identifier`. 
The clients could pass the `run_identifier` to the back end every time it communicates with it. On the server side, the client request is properly rejected if the `run_identifier` received from the client does not match its own internal `run_identifier`. The front end should be able to display a message asking the user to refresh the tab.

### Version of Taipy

- 3.X,
- 4.Y,
- develop

### Linked issues
- https://github.com/Avaiga/taipy-enterprise/issues/574

### Acceptance Criteria

- [ ] A unit test reproducing the bug is added.
- [ ] Related issue(s) in taipy-doc are created for documentation and Release Notes are updated.

### Code of Conduct

FredLL-Avaiga:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
In case of server shutdown and restart, the front-end can detect that it is not talking to the same server instance. 
controlled by config.app_id = True

- Closes #2344 

## How to reproduce the issue
- Use any app
- stop the server
- restart the server
- try to update a variable

### DIFF
--- a/frontend/taipy-gui/src/context/taipyReducers.ts
+++ b/frontend/taipy-gui/src/context/taipyReducers.ts
@@ -24,7 +24,7 @@ import { getBaseURL, TIMEZONE_CLIENT } from "../utils";
 import { parseData } from "../utils/dataFormat";
 import { MenuProps } from "../utils/lov";
 import { changeFavicon, getLocalStorageValue, IdMessage, storeClientId } from "./utils";
-import { lightenPayload, sendWsMessage, TAIPY_CLIENT_ID, WsMessage } from "./wsUtils";
+import { lightenPayload, sendWsMessage, TAIPY_APP_ID, TAIPY_CLIENT_ID, WsMessage } from "./wsUtils";
 
 export enum Types {
     SocketConnected = "SOCKET_CONNECTED",
@@ -181,7 +181,7 @@ const getUserTheme = (mode: PaletteMode) => {
                     },
                 },
             },
-        }),
+        })
     );
 };
 
@@ -226,7 +226,7 @@ export const messageToAction = (message: WsMessage) => {
                 (message as unknown as NavigateMessage).to,
                 (message as unknown as NavigateMessage).params,
                 (message as unknown as NavigateMessage).tab,
-                (message as unknown as NavigateMessage).force,
+                (message as unknown as NavigateMessage).force
             );
         } else if (message.type === "ID") {
             return createIdAction((message as unknown as IdMessage).id);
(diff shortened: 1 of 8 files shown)


frontend/taipy-gui/src/context/taipyReducers.ts
```diff
@@ -284,24 +286,46 @@ const initializeBroadcastManagement = (dispatch: Dispatch<TaipyBaseAction>) => {
     }, broadcast_timeout);
 };
 
+// App id
+const checkAppId = (appId: string) => {
+    if (!appId) {
+        return;
+    }
+    appId = `${appId}`;
```
FabienLelaquais, line 294: `appId= appId.toString();` ?

FabienLelaquais (approved):
Nice

FredLL-Avaiga:
> Nice

Are you happy with the config `app_id` boolean default false ?

frontend/taipy-gui/src/context/taipyReducers.ts
FredLL-Avaiga: `${...}` is broader

FabienLelaquais: It sure is. But why should it be? Certainly far slower don't you think?

FredLL-Avaiga: I was more interested in your advice on the naming of the config option (app_id)

dinhlongviolin1 (approved):
LGTM

PR #2454, opened 2025-02-20
local storage can be sent before the client id has been set 

FredLL-Avaiga:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
local storage can be sent before the client id has been set 
(try with private browsing)

### DIFF
--- a/frontend/taipy-gui/package-lock.json
+++ b/frontend/taipy-gui/package-lock.json
@@ -128,22 +128,21 @@
       }
     },
     "node_modules/@babel/core": {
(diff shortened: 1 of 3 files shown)


FabienLelaquais (approved):
👌

dinhlongviolin1 (approved):
Thanks! That is a very good catch!

namnguyen20999 (approved):
LGTM

PR #2457, opened 2025-02-22
Feature - Allow serialize Enum object

trgiangdo:

- Refactor
- Feature

## Description

This PR allows the _BaseSerializer to serialize and deserialize an Enum object.

### DIFF
--- a/taipy/common/config/_serializer/_base_serializer.py
+++ b/taipy/common/config/_serializer/_base_serializer.py
@@ -14,6 +14,7 @@ import re
 import types
 from abc import abstractmethod
 from datetime import datetime, timedelta
+from enum import Enum
 from typing import Any, Dict, Optional
 
 from .._config import _Config
@@ -38,6 +39,7 @@ class _BaseSerializer(object):
         "timedelta",
         "function",
         "class",
+        "enum",
         "SECTION",
     ]
     _section_class = {_GLOBAL_NODE_NAME: GlobalAppConfig}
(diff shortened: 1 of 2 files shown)


taipy/common/config/_serializer/_base_serializer.py
```diff
@@ -155,12 +159,16 @@ def _pythonify(cls, val):
...
+                    elif dynamic_type == "enum":
+                        return _TemplateHandler._to_enum(actual_val)
                     elif dynamic_type == "str":
                         return actual_val
                     else:
                         error_msg = f"Error loading toml configuration at {val}. {dynamic_type} type is not supported."
                         raise LoadingError(error_msg)
             if isinstance(val, dict):
+                if len(val) == 1 and list(val.keys())[0] in cls._registered_types.keys():
+                    return cls._registered_types[list(val.keys())[0]]._pythonify(list(val.values())[0])
```
trgiangdo, line 171: This `if` command handles the case where a dataclass instance is serialized into a dictionary.

For example, UserAttribute("email", "credential_mail", default="example@taipy.org") will be serialized to

```
{ "USER_ATTRIBUTE": {"name": "email", "credentials_name": "credentials_mail", "default": "example@taipy.org"}
```
The key is used to identify the type of the serialized object.

trgiangdo (commented):
Note for reviewer

toan-quach (approved):

toan-quach:
LGTM 😃

taipy/common/config/_serializer/_base_serializer.py
```diff
@@ -155,12 +159,16 @@ def _pythonify(cls, val):
                         return _TemplateHandler._to_function(actual_val)
                     elif dynamic_type == "class":
                         return _TemplateHandler._to_class(actual_val)
+                    elif dynamic_type == "enum":
+                        return _TemplateHandler._to_enum(actual_val)
```
jrobinAV, line 163: Just to be sure I understand: This case does not cover internal enums (SCOPE and FREQUENCY), as they are caught up by:
```
if dynamic_type in cls._registered_types.keys():
    return cls._registered_types[dynamic_type]._pythonify(actual_val)
```
which transforms the string `SCENARIO:SCOPE` into a Scope instance with the SCENARIO value.

Your case transforms the string `Gender.male:enum` into the Gender enum with the male value.

Am I correct?

jrobinAV (approved):

taipy/common/config/_serializer/_base_serializer.py
trgiangdo: yes, all registered types are handled first, then the enum

PR #2507, opened 2025-03-20
Refactor button and input components to use dynamic height based on size prop

### LINKED ISSUE #588: Make more MUI classes available for Buttons

nunoferna:

- [x] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
The sizes have been changed to these:
`small` - minHeight: 32px
`medium`/`default` - minHeight: 48px
`large` - minHeight: 56px

- Related Issue #588 
- Closes #588 

## How to reproduce the issue

Described in the issue.

### DIFF
--- a/frontend/taipy-gui/src/components/Taipy/Button.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Button.tsx
@@ -17,7 +17,7 @@ import MuiButton from "@mui/material/Button";
 import Tooltip from "@mui/material/Tooltip";
 
 import { createSendActionNameAction } from "../../context/taipyReducers";
-import { getCssSize, getSuffixedClassNames, TaipyActiveProps } from "./utils";
+import { getCssSize, getSuffixedClassNames, TaipyActiveProps, getInputButtonHeight } from "./utils";
 import { useClassNames, useDispatch, useDynamicProperty, useModule } from "../../utils/hooks";
 import { stringIcon, Icon, IconAvatar } from "../../utils/icon";
 import { getComponentClassName } from "./TaipyStyle";
@@ -42,7 +42,10 @@ const Button = (props: ButtonProps) => {
     const className = useClassNames(props.libClassName, props.dynamicClassName, props.className);
     const active = useDynamicProperty(props.active, props.defaultActive, true);
     const hover = useDynamicProperty(props.hoverText, props.defaultHoverText, undefined);
-    const buttonSx = useMemo(() => (props.width ? { width: getCssSize(props.width) } : undefined), [props.width]);
+    const buttonSx = useMemo(() => ({
+        minHeight: getInputButtonHeight(props.size),
+        ...(props.width && { width: getCssSize(props.width) }),
+    }), [props.size, props.width]);
 
     const handleClick = useCallback(() => {
         dispatch(createSendActionNameAction(id, module, onAction));
--- a/frontend/taipy-gui/src/components/Taipy/Button.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Button.tsx
@@ -17,7 +17,7 @@ import MuiButton from "@mui/material/Button";
 import Tooltip from "@mui/material/Tooltip";
 
 import { createSendActionNameAction } from "../../context/taipyReducers";
-import { getCssSize, getSuffixedClassNames, TaipyActiveProps, getInputButtonHeight } from "./utils";
+import {getCssSize, getSuffixedClassNames, TaipyActiveProps, expandSx} from "./utils";
 import { useClassNames, useDispatch, useDynamicProperty, useModule } from "../../utils/hooks";
 import { stringIcon, Icon, IconAvatar } from "../../utils/icon";
 import { getComponentClassName } from "./TaipyStyle";
@@ -42,10 +42,11 @@ const Button = (props: ButtonProps) => {
     const className = useClassNames(props.libClassName, props.dynamicClassName, props.className);
     const active = useDynamicProperty(props.active, props.defaultActive, true);
     const hover = useDynamicProperty(props.hoverText, props.defaultHoverText, undefined);
-    const buttonSx = useMemo(() => ({
-        minHeight: getInputButtonHeight(props.size),
-        ...(props.width && { width: getCssSize(props.width) }),
-    }), [props.size, props.width]);
+    const buttonSx = useMemo(() =>
+        expandSx(undefined,
+            props.width ? { width: getCssSize(props.width) } : undefined,
+            size !== "medium" ? { minHeight: "unset" } : undefined)
+    , [props.width, size]);
 
     const handleClick = useCallback(() => {
         dispatch(createSendActionNameAction(id, module, onAction));
--- a/frontend/taipy-gui/src/components/Taipy/Button.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Button.tsx
@@ -17,7 +17,7 @@ import MuiButton from "@mui/material/Button";
 import Tooltip from "@mui/material/Tooltip";
 
 import { createSendActionNameAction } from "../../context/taipyReducers";
-import {getCssSize, getSuffixedClassNames, TaipyActiveProps, expandSx} from "./utils";
+import { getCssSize, getSuffixedClassNames, TaipyActiveProps } from "./utils";
 import { useClassNames, useDispatch, useDynamicProperty, useModule } from "../../utils/hooks";
 import { stringIcon, Icon, IconAvatar } from "../../utils/icon";
 import { getComponentClassName } from "./TaipyStyle";
@@ -42,11 +42,7 @@ const Button = (props: ButtonProps) => {
     const className = useClassNames(props.libClassName, props.dynamicClassName, props.className);
     const active = useDynamicProperty(props.active, props.defaultActive, true);
     const hover = useDynamicProperty(props.hoverText, props.defaultHoverText, undefined);
-    const buttonSx = useMemo(() =>
-        expandSx(undefined,
-            props.width ? { width: getCssSize(props.width) } : undefined,
-            size !== "medium" ? { minHeight: "unset" } : undefined)
-    , [props.width, size]);
+    const buttonSx = useMemo(() => (props.width ? { width: getCssSize(props.width) } : undefined), [props.width]);
 
     const handleClick = useCallback(() => {
         dispatch(createSendActionNameAction(id, module, onAction));
--- a/frontend/taipy-gui/src/components/Taipy/Input.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Input.tsx
@@ -21,7 +21,7 @@ import ArrowDropUpIcon from "@mui/icons-material/ArrowDropUp";
 import ArrowDropDownIcon from "@mui/icons-material/ArrowDropDown";
 
 import { createSendActionNameAction, createSendUpdateAction } from "../../context/taipyReducers";
-import { getCssSize, TaipyInputProps } from "./utils";
+import { getCssSize, TaipyInputProps, getInputButtonHeight } from "./utils";
 import { useClassNames, useDispatch, useDynamicProperty, useModule } from "../../utils/hooks";
 import { getComponentClassName } from "./TaipyStyle";
 
@@ -66,6 +66,7 @@ const Input = (props: TaipyInputProps) => {
         multiline = false,
         actionOnBlur = false,
         linesShown = 5,
+        size = "medium",
     } = props;
 
     const [value, setValue] = useState(defaultValue);
@@ -84,14 +85,14 @@ const Input = (props: TaipyInputProps) => {
     const max = useDynamicProperty(props.max, props.defaultMax, undefined);
 
     const textSx = useMemo(
-        () =>
-            props.width
-                ? {
-                    ...numberSx,
-                    maxWidth: getCssSize(props.width),
-                }
-                : numberSx,
-        [props.width]
+        () => ({
+            ...numberSx,
+            maxWidth: props.width && getCssSize(props.width),
+            "& .MuiInputBase-root": {
+                minHeight: getInputButtonHeight(props.size),
+            },
+        }),
+        [props.width, props.size]
     );
 
     const updateValueWithDelay = useCallback(
@@ -374,6 +375,7 @@ const Input = (props: TaipyInputProps) => {
                     onKeyDown={handleAction}
                     multiline={multiline}
                     minRows={linesShown}
+                    size={size}
                 />
                 {props.children}
             </>
--- a/frontend/taipy-gui/src/components/Taipy/Input.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Input.tsx
@@ -21,7 +21,7 @@ import ArrowDropUpIcon from "@mui/icons-material/ArrowDropUp";
 import ArrowDropDownIcon from "@mui/icons-material/ArrowDropDown";
 
 import { createSendActionNameAction, createSendUpdateAction } from "../../context/taipyReducers";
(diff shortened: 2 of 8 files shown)


nunoferna:
I think I have fixed this issue. The Input element on MUI has two sizes, but I don't know how to add it to the input markdown like this:
`<|input|size=small|>`

image

The issue was not about the input itself but as the input needs to have consistency with the button, I needed to change both.

The previous image was written with:
```
from taipy.gui import Gui

buttons = """
<|input|size=small|>
<|button|size=small|label=Small|>
<|input|>
<|button|label=Medium|>
<|button|size=large|label=Big|>
"""

Gui(buttons).run()
```

As you can see everything looks great, but the markdown of the input size small is not well.

frontend/taipy-gui/src/components/Taipy/Input.tsx
```diff
@@ -84,14 +85,14 @@ const Input = (props: TaipyInputProps) => {
...
-                ? {
-                    ...numberSx,
-                    maxWidth: getCssSize(props.width),
-                }
-                : numberSx,
-        [props.width]
+        () => ({
+            ...numberSx,
+            maxWidth: props.width && getCssSize(props.width),
+            "& .MuiInputBase-root": {
```
NOF03, line 91: This was needed, instead of just minHeight:..., because that was changing only the external div and not the input properly.

frontend/taipy-gui/src/components/Taipy/utils.ts
```diff
@@ -168,3 +169,15 @@ export const expandSx = (sx: SxProps | undefined, ...partials: (SxProps | undefi
...
+    switch (size) {
+        case "small":
+            return "32px";
+        case "large":
+            return "56px";
+        case "medium":
+        default:
+            return "48px";
+    }
+};
```
NOF03, line 183: It is possible to remove this function by changing the [stylekit](https://github.com/Avaiga/taipy/blob/bf929ea26490b094a6b638d49e869a58b4104b5c/taipy/gui/_default_config.py#L39) for:
```
"input_button_height_small": "32px",
"input_button_height_medium": "48px",
"input_button_height_large": "56px",
```    
 and then in the minHeight prop of Button.tsx and Input.tsx, replace the `getInputButtonHeight(props.size)` for `var(--input_button_height_${props.size})`

nunoferna:
I was just watching another issue, noticed that for other inputs (TextField), like in the Chat.tsx they were not having the 48px height. Needed to add again the stylekit `input_button_height` and just changed in the component input properly.

frontend/taipy-gui/src/components/Taipy/utils.ts
nunoferna: The return values can be changed by some variables created in the Stylekit

frontend/taipy-gui/src/components/Taipy/utils.ts
```diff
@@ -168,3 +169,15 @@ export const expandSx = (sx: SxProps | undefined, ...partials: (SxProps | undefi
     }, start);
     return start === end ? sx : end;
 };
+
+export const getInputButtonHeight = (size?: "small" | "medium" | "large") => {
```
FredLL-Avaiga: I don't feel the min-height value should be set in pixel.
I feel it should be unset when we don't want it

frontend/taipy-gui/src/components/Taipy/Button.tsx
```diff
@@ -42,7 +42,10 @@ const Button = (props: ButtonProps) => {
     const className = useClassNames(props.libClassName, props.dynamicClassName, props.className);
     const active = useDynamicProperty(props.active, props.defaultActive, true);
     const hover = useDynamicProperty(props.hoverText, props.defaultHoverText, undefined);
-    const buttonSx = useMemo(() => (props.width ? { width: getCssSize(props.width) } : undefined), [props.width]);
+    const buttonSx = useMemo(() => ({
+        minHeight: getInputButtonHeight(props.size),
+        ...(props.width && { width: getCssSize(props.width) }),
+    }), [props.size, props.width]);
```
FredLL-Avaiga: I would propose to use the utils/expandSx function that allows to not create an object if it is not necessary
`const buttonSx = useMemo(() => expandSx(undefined, props.width ? { width: getCssSize(props.width) } : undefined, size !== "medium" ? {minHeight: "unset"}: undefined), [props.width]);`

frontend/taipy-gui/src/components/Taipy/Input.tsx
```diff
@@ -66,6 +66,7 @@ const Input = (props: TaipyInputProps) => {
         multiline = false,
         actionOnBlur = false,
         linesShown = 5,
+        size = "medium",
```
FredLL-Avaiga, line 69: If you want to support size in input, it has to be defined on the python side too:
[here](https://github.com/Avaiga/taipy/blob/6c236cb8551765658402f6b5143be2c96505f725/taipy/gui/_renderers/factory.py#L333)
and as a property definition [there](https://github.com/Avaiga/taipy/blob/6c236cb8551765658402f6b5143be2c96505f725/taipy/gui/viselements.json#L148)

You can copy the size property definition from button

FredLL-Avaiga (changes requested):
thanks for your work
please, as much as possible,  do not hard-code dimension.
should some tests be added ?

frontend/taipy-gui/src/components/Taipy/utils.ts
nunoferna: I put them static, only because it will have an inconsistent height between the Input size=small and the Button size=small, I guess you would like to maintain that consistency. For the large, as your default was 48px, the large is 42px, it wouldn't make sense.

frontend/taipy-gui/src/components/Taipy/Input.tsx
nunoferna: I will try and add this!

frontend/taipy-gui/src/components/Taipy/utils.ts
nunoferna: I was just thinking, maybe as you had a default value for the medium size, maybe make the small and the big depend on it like for small `minHeight:` `calc(var(--input_button_height) * 0.8)`, and for big * 1.2. I am just doing it because of this var(--input_button_height) you have.

frontend/taipy-gui/src/components/Taipy/Input.tsx
nunoferna: Got it in this [commit](https://github.com/Avaiga/taipy/pull/2507/commits/7a7531ac8d8333f78877f15fc1e2bb8bd5d77560) 

Thanks!

nunoferna:
The current state is this:

image

Code used:
```
from taipy.gui import Gui

buttons = """
<|Insert an input|input|size=small|label=input|>
<|button|size=small|label=Small|>
<|Insert another input|input|label=another input|>
<|button|label=Medium|>
<|button|size=large|label=Big|>
"""

Gui(buttons).run()
```

frontend/taipy-gui/src/components/Taipy/utils.ts
FredLL-Avaiga: the `input-button-height` css variable is used by Mui directly.
- for small, I think we just want to `unset` the `minHeight`
- for medium, do nothing
- for large (button only), I don't really know how Mui sets the size of the element

frontend/taipy-gui/src/components/Taipy/Input.tsx
```diff
@@ -84,14 +85,14 @@ const Input = (props: TaipyInputProps) => {
...
     const textSx = useMemo(
-        () =>
-            props.width
-                ? {
-                    ...numberSx,
-                    maxWidth: getCssSize(props.width),
-                }
-                : numberSx,
-        [props.width]
+        () => ({
```
FredLL-Avaiga: use `expandSx`

FredLL-Avaiga (changes requested):

nunoferna:
Hi @FredLL-Avaiga, I did as you said, but I am still having this issue.

image

I am just seeing this as best option:
```
const getMinHeight = (size) =>
    size === "small"
        ? "calc(var(--input_button_height) * 0.8)"
        : size === "big"
        ? "calc(var(--input_button_height) * 1.2)"
        : undefined;
```
I saw the mui docs for others sizes but I didn't get nothing related to the height, it is only about padding and fontSize and that is working fine.

FredLL-Avaiga:
> Hi @FredLL-Avaiga, I did as you said, but I am still having this issue.
> 
> image
> 
> I am just seeing this as best option:
> 
> ```
> const getMinHeight = (size) =>
>     size === "small"
>         ? "calc(var(--input_button_height) * 0.8)"
>         : size === "big"
>         ? "calc(var(--input_button_height) * 1.2)"
>         : undefined;
> ```
> 
> I saw the mui docs for others sizes but I didn't get nothing related to the height, it is only about padding and fontSize and that is working fine.

I don't understand why you want to set a min height for small size, it's working fine when unset

nunoferna:
> > Hi @FredLL-Avaiga, I did as you said, but I am still having this issue.
> > image
> > I am just seeing this as best option:
> > ```
> > const getMinHeight = (size) =>
> >     size === "small"
> >         ? "calc(var(--input_button_height) * 0.8)"
> >         : size === "big"
> >         ? "calc(var(--input_button_height) * 1.2)"
> >         : undefined;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I saw the mui docs for others sizes but I didn't get nothing related to the height, it is only about padding and fontSize and that is working fine.
> 
> I don't understand why you want to set a min height for small size, it's working fine when unset

Ahh alright! I thought you wanted to keep the same height for both input and button. For the big size of button do I just multiply then?

FredLL-Avaiga:
> Ahh alright! I thought you wanted to keep the same height for both input and button. For the big size of button do I just multiply then?

For large button, it would be interesting to see what's making it larger when the CSS var is not defined
Or what's blocking in taipy compared to standard mui

nunoferna:
> > Ahh alright! I thought you wanted to keep the same height for both input and button. For the big size of button do I just multiply then?
> 
> For large button, it would be interesting to see what's making it larger when the CSS var is not defined Or what's blocking in taipy compared to standard mui

So, I just get it, what makes it different is the padding and the font size:
for large: `font-size: 0.9375rem; padding: 7px 21px;` for medium: `font-size: 0.875rem; padding: 5px 15px;`

The code from the MUI Button size large:
```
{
          props: {
            size: 'large',
            variant: 'outlined',
          },
          style: {
            padding: '7px 21px',
            fontSize: theme.typography.pxToRem(15),
          },
},
```

and this is for the default size:
```
{
          props: { variant: 'outlined' },
          style: {
            padding: '5px 15px',
```

For the large it gives a total height of 42.25px. As you have a minHeight defined for the medium, I think it is needed to add for the large too. If we remove the minHeight from the medium we can see that there is a relation of 1,4 from the padding which is the 7/5=1,4 and the 21/15=1,4.

FredLL-Avaiga:
> > > Ahh alright! I thought you wanted to keep the same height for both input and button. For the big size of button do I just multiply then?
> > 
> > 
> > For large button, it would be interesting to see what's making it larger when the CSS var is not defined Or what's blocking in taipy compared to standard mui
> 
> So, I just get it, what makes it different is the padding and the font size: for large: `font-size: 0.9375rem; padding: 7px 21px;` for medium: `font-size: 0.875rem; padding: 5px 15px;`
> 
> The code from the MUI Button size large:
> 
> ```
> {
>           props: {
>             size: 'large',
>             variant: 'outlined',
>           },
>           style: {
>             padding: '7px 21px',
>             fontSize: theme.typography.pxToRem(15),
>           },
> },
> ```
> 
> and this is for the default size:
> 
> ```
> {
>           props: { variant: 'outlined' },
>           style: {
>             padding: '5px 15px',
> ```
> 
> For the large it gives a total height of 42.25px. As you have a minHeight defined for the medium, I think it is needed to add for the large too. If we remove the minHeight from the medium we can see that there is a relation of 1,4 from the padding which is the 7/5=1,4 and the 21/15=1,4.

tx for digging,
How are those css properties overridden in taipy ?

nunoferna:
> > > > Ahh alright! I thought you wanted to keep the same height for both input and button. For the big size of button do I just multiply then?
> > > 
> > > 
> > > For large button, it would be interesting to see what's making it larger when the CSS var is not defined Or what's blocking in taipy compared to standard mui
> > 
> > 
> > So, I just get it, what makes it different is the padding and the font size: for large: `font-size: 0.9375rem; padding: 7px 21px;` for medium: `font-size: 0.875rem; padding: 5px 15px;`
> > The code from the MUI Button size large:
> > ```
> > {
> >           props: {
> >             size: 'large',
> >             variant: 'outlined',
> >           },
> >           style: {
> >             padding: '7px 21px',
> >             fontSize: theme.typography.pxToRem(15),
> >           },
> > },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and this is for the default size:
> > ```
> > {
> >           props: { variant: 'outlined' },
> >           style: {
> >             padding: '5px 15px',
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > For the large it gives a total height of 42.25px. As you have a minHeight defined for the medium, I think it is needed to add for the large too. If we remove the minHeight from the medium we can see that there is a relation of 1,4 from the padding which is the 7/5=1,4 and the 21/15=1,4.
> 
> tx for digging, How are those css properties overridden in taipy ?

I don't see them being overridden, but I guess if you want to override them we could change in the [stylekit.ts](https://github.com/Avaiga/taipy/blob/67e90d25b5e0d5e1c6b430fe5f53535790d114a2/frontend/taipy-gui/src/themes/stylekit.ts#L116)

nunoferna:
@FredLL-Avaiga I think I get something you might like for this solution. For the large button we can give the line-height the same value as the min-height.

```ts
        MuiButton: {
            styleOverrides: {
                root: {
                    height: "auto",
                    minHeight: window.taipyConfig?.stylekit?.inputButtonHeight,
                    marginBottom: 4,
              -->   "&.MuiButton-sizeLarge": {
                        lineHeight: window.taipyConfig?.stylekit?.inputButtonHeight,
                    },
                },
            },
        },
```

With this the large button is typically the size of the medium but with the padding. It stays like this:

image

In the button instead of using this expression `size !== "medium" ? { minHeight: "unset" } : undefined`, change to this `size === "small" ? { minHeight: "unset" } : undefined`

FredLL-Avaiga:
> @FredLL-Avaiga I think I get something you might like for this solution. For the large button we can give the line-height the same value as the min-height.
> 
> ```ts
>         MuiButton: {
>             styleOverrides: {
>                 root: {
>                     height: "auto",
>                     minHeight: window.taipyConfig?.stylekit?.inputButtonHeight,
>                     marginBottom: 4,
>               -->   "&.MuiButton-sizeLarge": {
>                         lineHeight: window.taipyConfig?.stylekit?.inputButtonHeight,
>                     },
>                 },
>             },
>         },
> ```
> 
> With this the large button is typically the size of the medium but with the padding. It stays like this:
> 
> image
> 
> In the button instead of using this expression `size !== "medium" ? { minHeight: "unset" } : undefined`, change to this `size === "small" ? { minHeight: "unset" } : undefined`

Excellent, I would go a bit further and modify *only* the stylekit.ts file
setting `minHeight` only for `-sizeMedium` class for input and button

FredLL-Avaiga (approved):
Well done, much better like this.
I should have remembered the styleKit.ts part

PR #2509, opened 2025-03-24
Refactor - Remove _Manager._set() api and add _Manager._update()

trgiangdo:

- Refactor
- Feature

### DIFF
--- a/taipy/core/data/_data_manager.py
+++ b/taipy/core/data/_data_manager.py
@@ -87,13 +87,13 @@ class _DataManager(_Manager[DataNode], _VersionMixin):
     def _create(
         cls, data_node_config: DataNodeConfig, owner_id: Optional[str], parent_ids: Optional[Set[str]]
     ) -> DataNode:
-        data_node = cls.__create(data_node_config, owner_id, parent_ids)
+        data_node = cls.__instantiate(data_node_config, owner_id, parent_ids)
         cls._repository._save(data_node)
         Notifier.publish(_make_event(data_node, EventOperation.CREATION))
         return data_node
 
     @classmethod
-    def __create(
+    def __instantiate(
         cls, data_node_config: DataNodeConfig, owner_id: Optional[str], parent_ids: Optional[Set[str]]
     ) -> DataNode:
         try:
--- a/taipy/core/data/_data_manager.py
+++ b/taipy/core/data/_data_manager.py
@@ -66,7 +66,7 @@ class _DataManager(_Manager[DataNode], _VersionMixin):
             dn_configs_and_owner_id, cls._build_filters_with_version(None)
         )
         return {
-            dn_config: data_nodes.get((dn_config, owner_id)) or cls._create_and_set(dn_config, owner_id, None)
+            dn_config: data_nodes.get((dn_config, owner_id)) or cls._create(dn_config, owner_id, None)
             for dn_config, owner_id in dn_configs_and_owner_id
         }
 
@@ -84,11 +84,11 @@ class _DataManager(_Manager[DataNode], _VersionMixin):
         return reason_collection
 
     @classmethod
-    def _create_and_set(
+    def _create(
         cls, data_node_config: DataNodeConfig, owner_id: Optional[str], parent_ids: Optional[Set[str]]
     ) -> DataNode:
         data_node = cls.__create(data_node_config, owner_id, parent_ids)
-        cls._set(data_node)
+        cls._repository._save(data_node)
         Notifier.publish(_make_event(data_node, EventOperation.CREATION))
         return data_node
 
--- a/taipy/core/data/parquet.py
+++ b/taipy/core/data/parquet.py
@@ -187,7 +187,6 @@ class ParquetDataNode(DataNode, _FileDataNodeMixin, _TabularDataNodeMixin):
         # Ensure that the columns are strings, otherwise writing will fail with pandas 1.3.5
         df.columns = df.columns.astype(str)
         df.to_parquet(self._path, **kwargs)
-        self.track_edit(timestamp=datetime.now(), editor_id=editor_id)
 
     def read_with_kwargs(self, **read_kwargs):
         """Read data from this data node.
--- a/taipy/core/exceptions/exceptions.py
+++ b/taipy/core/exceptions/exceptions.py
@@ -28,10 +28,6 @@ class OrchestratorServiceIsAlreadyRunning(Exception):
     """Raised if the Orchestrator service is already running."""
 
 
-class EntityAlreadyExists(Exception):
-    """Raised if it is trying to create an Entity that has already existed."""
-
-
 class NonExistingEntity(Exception):
     """Raised if a requested entity is not known by the Entity Manager."""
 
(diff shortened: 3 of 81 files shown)


taipy/core/task/_task_manager.py
```diff
@@ -52,10 +53,16 @@ def _orchestrator(cls) -> Type[_AbstractOrchestrator]:
...
 
     @classmethod
-    def _set(cls, task: Task) -> None:
-        cls.__save_data_nodes(task.input.values())
-        cls.__save_data_nodes(task.output.values())
-        super()._set(task)
+    def _create(cls, task: Task) -> None:
+        for dn in itertools.chain(task.input.values(), task.output.values()):
+            _DataManagerFactory._build_manager()._repository._save(dn)
+        cls._repository._save(task)
```
trgiangdo, line 59: The `_create()` signature here is different from other `_create()` methods from other entity managers.

Other managers, scenario for example, accept a config as the input to create new entity.

However, `_TaskManager._create()` takes an instance of the Task entity.

This method is used by scenario creation, and in the tests. But we can consider removing it since it is not consistent.

taipy/core/data/_data_manager.py
```diff
@@ -84,11 +84,11 @@ def _can_create(cls, config: Optional[DataNodeConfig] = None) -> ReasonCollectio
         return reason_collection
 
     @classmethod
-    def _create_and_set(
+    def _create(
         cls, data_node_config: DataNodeConfig, owner_id: Optional[str], parent_ids: Optional[Set[str]]
     ) -> DataNode:
         data_node = cls.__create(data_node_config, owner_id, parent_ids)
```
jrobinAV: `__create` could be renamed to `__instantiate` to make it more explicit. What do you think?

taipy/core/data/parquet.py
```diff
@@ -187,7 +187,6 @@ def _write_with_kwargs(self, data: Any, editor_id: Optional[str] = None, **write
         # Ensure that the columns are strings, otherwise writing will fail with pandas 1.3.5
         df.columns = df.columns.astype(str)
         df.to_parquet(self._path, **kwargs)
-        self.track_edit(timestamp=datetime.now(), editor_id=editor_id)
```
jrobinAV, line 190: hum?

taipy/core/exceptions/exceptions.py
```diff
@@ -28,8 +28,31 @@ class OrchestratorServiceIsAlreadyRunning(Exception):
     """Raised if the Orchestrator service is already running."""
 
 
+class EntityAlreadyExists(Exception):
+    """Raised if it is trying to create an Entity that has already existed."""
```
jrobinAV:
```suggestion
    """Raised if trying to create an Entity that has already exists."""
```

taipy/core/exceptions/exceptions.py
```diff
@@ -28,8 +28,31 @@ class OrchestratorServiceIsAlreadyRunning(Exception):
...
+class NonExistingEntity(Exception):
+    """Raised if a requested entity is not known by the Entity Manager."""
+
+    def __init__(self, entity_id: str):
+        self.message = f"Entity {entity_id} does not exist."
+
+
 class CycleAlreadyExists(Exception):
-    """Raised if it is trying to create a Cycle that has already exists."""
+    """Raised if it is trying to create a Cycle that has already existed."""
```
jrobinAV:
```suggestion
    """Raised if trying to create a Cycle that already exists."""
```

taipy/core/exceptions/exceptions.py
```diff
@@ -28,8 +28,31 @@ class OrchestratorServiceIsAlreadyRunning(Exception):
...
+
+
+class DataNodeAlreadyExists(Exception):
+    """Raised if it is trying to create a DataNode that has already existed."""
```
jrobinAV:
```suggestion
    """Raised if trying to create a DataNode that already exists."""
```

taipy/core/exceptions/exceptions.py
```diff
@@ -28,8 +28,31 @@ class OrchestratorServiceIsAlreadyRunning(Exception):
...
+
+
+class TaskAlreadyExists(Exception):
+    """Raised if it is trying to create a Task that has already existed."""
```
jrobinAV:
```suggestion
    """Raised if trying to create a Task that already exists."""
```

taipy/core/exceptions/exceptions.py
```diff
@@ -28,8 +28,31 @@ class OrchestratorServiceIsAlreadyRunning(Exception):
...
+
+
+class ScenarioAlreadyExists(Exception):
+    """Raised if it is trying to create a Scenario that has already existed."""
```
jrobinAV:
```suggestion
    """Raised if trying to create a Scenario that already exists."""
```

taipy/core/task/_task_manager.py
jrobinAV: I believe it makes sense to align all the managers. But I am not sure about the semantics.

what do we expect for the `manager._create()` method? 
- To instantiate an entity in memory from various parameters?
- To save in the repository in-memory entity given as a parameter?
- To instantiate an entity and the nested entities?
- To instantiate an entity and save it in the repository?
- To instantiate an entity and the nested entities, save the entity and the nested ones.
- To save in the repository an in-memory entity and nested in-memory entities given as a parameter?

taipy/core/data/parquet.py
trgiangdo: The `_write_with_kwargs()` method is only called by the `DataNode.write()`, which already calls to the `track_edit()` method.

It seems to be an unnecessary legacy line for me

taipy/core/exceptions/exceptions.py
```diff
@@ -28,6 +28,13 @@ class OrchestratorServiceIsAlreadyRunning(Exception):
...
 
+class NonExistingEntity(Exception):
+    """Raised if a requested entity is not known by the Entity Manager."""
+
+    def __init__(self, entity_id: str):
+        self.message = f"Entity {entity_id} does not exist."
+
+
 class CycleAlreadyExists(Exception):
     """Raised if it is trying to create a Cycle that has already exists."""
```
jrobinAV:
```suggestion
    """Raised if a cycle creation is requested for a cycle that already exists."""
```

taipy/core/data/parquet.py
jrobinAV: OK. SO it is a bug as the edit is created twice. Thanks for the fix.

taipy/core/task/_task_manager.py
trgiangdo: After discussion, the difference is ok because there's no abstract signature for `_Manager._create()`.

The `_JobManager._create()` also receives a Task entity as the input

jrobinAV (approved):

trgiangdo:
Conflict with enterprise is resolved in https://github.com/Avaiga/taipy-enterprise/pull/655

jrobinAV:
Let's merge then.

PR #2570, opened 2025-04-16
update publish.yml

jrobinAV:
- to be backported on develop (and create a requirement.txt file local to the release scripts.)


FabienLelaquais (approved):
👌

PR #2577, opened 2025-04-18
Refactor - Allow passing extra package name to check dependency

trgiangdo:
## Type of PR

- 🛠 Refactor

## Description

The `_check_dependency_is_installed()` function now accept the `extra_taipy_package_name` to be extendable.

Also, the taipy sub-library name can be provided.


.github/workflows/partial-tests.yml
```diff
@@ -17,6 +17,8 @@ jobs:
           mypy_flags:  "--ignore-missing-imports --implicit-optional --disable-error-code attr-defined --no-namespace-packages --exclude (taipy/templates/|tools/|doc/gui/examples/.*/builder.py|taipy/common/config/config.pyi) --follow-imports skip --disable-error-code import-untyped"
 
       - uses: chartboost/ruff-action@v1
+        with:
+          src: './taipy ./tests'
```
jrobinAV: What for?

taipy/core/common/_check_dependencies.py
jrobinAV, line 1: Should it be moved to `taipy.common` instead of `taipy.core.common`?

.github/workflows/partial-tests.yml
trgiangdo: ruff is complaining about a few formatting from `tools` folder.

I don't think we need to use ruff to check for that. What do you think?

jrobinAV: Yep, right.

jrobinAV (dismissed):
I am ok for now to unblock the azure topic.

However, I would like to have a centralized module in `taipy.common` to check dependencies. This should also include the GUI and enterprise packages.
Please create an issue for that. Thx!

jrobinAV (dismissed):

jrobinAV (dismissed):

jrobinAV (approved):

PR #2669, opened 2025-06-02
support boolean value in env file config

### LINKED ISSUE #2453: [🐛 BUG] Unable to load configurations from an env file
### What went wrong? 🤔

Hi, 

So we prefer to keep all our configurations in a env file rather than parameters in the main file, so we found the following from the doc.

[Configuring with a .env file](https://docs.taipy.io/en/latest/userman/advanced_features/configuration/gui-config/)

All parameters can also be set to any values and stored as a list of key-value pairs in a text file for your Gui instance to consume. The name of this file is the one provided as the env_filename parameter to the [Gui constructor](https://docs.taipy.io/en/latest/refmans/reference/pkg_taipy/pkg_gui/Gui/index.html#taipy.gui.Gui.__init__).

If you have such a configuration file, then the value associated with a configuration parameter will override the one provided in the [run()](https://docs.taipy.io/en/latest/refmans/reference/pkg_taipy/pkg_gui/Gui/index.html#taipy.gui.Gui.run) method of your Gui instance.

It sounds like if we provide an env file and specify it with the env_filename parameter it should override anything provided in the run() function. However, for some reason, it didn't work. 

### Expected Behavior

As an example, I created a simple .env file and it only contains one line:
`run_browser=False`
and I specified the env_filename by doing the following:
`Gui(env_filename=".env").run()`

I expected taipy to **NOT** open a browser

### Steps to Reproduce Issue

1.create a .env file and put line `run_browser=False`
2.specify env_filename in your Gui constructor in the main.py file like `Gui(env_filename=".env").run()`
3.execute `taipy run main.py`

### Solution Proposed

I don't have a solution yet but I spent some time digging the code and here's what I found:
- the function that is used to handle this env_filename parameter resides in the gui/config.py file, function signature `def _build_config(self, root_dir, env_filename, kwargs)`
- the actual logic is in the block `if os.path.isfile(env_file_abs_path):`
- maybe I misunderstood but it seems this line should overwrite the value `config[key] = value if config[key] is None else type(config[key])(value)  # type: ignore`
- however, the config already has a default value, for example, run_browser=True by default
- if it can **NOT** override existing configs (because they may have default values), how does it work? Or did I misunderstand the doc?
Thank you.

[PS]
just want to mention that normally boolean type environment variables are not supported so I also tried something like `run_browser=0` but still no luck

### Screenshots

Image

### Additional Context

```bash

```

### Acceptance Criteria

- [ ] A unit test reproducing the bug is added.
- [ ] Related issue(s) in taipy-doc are created for documentation and Release Notes are updated.

### Code of Conduct

FredLL-Avaiga:
resolves #2453

## What type of PR is this? (Check all that apply)

- [ ] 🛠 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] ⚡ Optimization
- [ ] 📝 Documentation Update

## Description
support boolean value in env file config

- Closes #2453 

## How to reproduce the issue

```env
run_browser=False
```

```py
from taipy.gui import Gui

input_value = 0
output_value = None


with tgb.Page() as update_test:
    tgb.selector(
        "{input_value}",
        lov=[0, 1, 2, 3, 4],
        label="select number to add 1",
        dropdown=True,
    )
    tgb.button(
        label="input to output",
        on_action=lambda state: state.assign("output_value", int(state.input_value) + 1),
    )
    tgb.text("Your Number plus 1 is {output_value}")


if __name__ == "__main__":
    gui = Gui(page=update_test, env_filename="demo.env")

    gui.run(
        title="2453 - Env file demo",
    )

```

### DIFF
--- a/taipy/core/data/_file_datanode_mixin.py
+++ b/taipy/core/data/_file_datanode_mixin.py
@@ -14,7 +14,7 @@ import pathlib
 import shutil
 from datetime import datetime
 from os.path import isfile
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Callable, Dict, Optional, cast
 
 from taipy.common.config import Config
 from taipy.common.logger._taipy_logger import _TaipyLogger
@@ -47,7 +47,7 @@ class _FileDataNodeMixin:
     __logger = _TaipyLogger._get_logger()
 
     def __init__(self, properties: Dict) -> None:
-        self._path: str = properties.get(self._PATH_KEY, properties.get(self._DEFAULT_PATH_KEY))
+        self._path: str = cast(str, properties.get(self._PATH_KEY, properties.get(self._DEFAULT_PATH_KEY)))
         self._is_generated: bool = properties.get(self._IS_GENERATED_KEY, self._path is None)
         self._last_edit_date: Optional[datetime] = None
 
@@ -248,5 +248,5 @@ class _FileDataNodeMixin:
             else:
                 shutil.copy(self._path, new_path)
             normalize_path = _normalize_path(new_path)
-            dest._path = normalize_path
+            dest._path = normalize_path # type: ignore[attr-defined]
             dest._properties[self._PATH_KEY] = normalize_path
--- a/taipy/core/data/excel.py
+++ b/taipy/core/data/excel.py
@@ -10,7 +10,7 @@
 # specific language governing permissions and limitations under the License.
 
 from datetime import datetime, timedelta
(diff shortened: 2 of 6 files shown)


FabienLelaquais (dismissed):
👌👍

dinhlongviolin1 (approved):
LGTM

FabienLelaquais (approved):
👍

PR #2757, opened 2025-10-26
Bug/#2756 number min max enforce

### LINKED ISSUE #2756: [🐛 BUG] Number component does not respect min/max for manual input
### 🔍 What went wrong?

The min/max bounds are respected when clicking arrow but not for manual input

### 🔄 Steps to Reproduce

```python

number = 2

def on_change(state: State, var_name: str, var_value):
    print(var_name, type(var_value), var_value)

with tgb.Page() as page:
    tgb.number("{number}", min=1, max=3.0)

Gui(page=page).run()
```

1. Run the code above
2. Prepend the element with `-` to turn it into -2
3. `number` is now -2, even though min should be 1

### 📦 Taipy Version

develop

### 📋 Additional Context (Optional)

```bash

```

### 📜 Code of Conduct

- [x] I have checked the [existing issues](https://github.com/Avaiga/taipy/issues?q=is%3Aissue+) to avoid duplicates.

### ✅ Acceptance Criteria

- [ ] A reproducible unit test is added.
- [ ] Code coverage is **at least 90%**.
- [ ] Relevant **documentation updates** or an issue created in

Agney-gt:
## What type of PR is this? (Check all that apply)

* [ ] 🛠 Refactor
* [ ] ✨ Feature
* [x] 🐛 Bug Fix
* [ ] ⚡ Optimization
* [ ] 📝 Documentation Update

## Description

This PR ensures that the `tgb.number` control in Taipy GUI respects the `min` and `max` bounds **for both manual input and arrow step changes**. Previously, manually typing a value could bypass the `min`/`max` constraints. The fix clamps the value inside the allowed range and updates the state accordingly.

* Related Issue #2756
* Closes #2756

## How to reproduce the issue

1. Run the following code:

```python

number = -2

def on_change(state: State, var_name: str, var_value):
    print(var_name, type(var_value), var_value)

with tgb.Page() as page:
    tgb.number("{number}", min=1, max=3.0)

Gui(page=page).run()
```

2. Manually type a value below 1 or above 3.
3. Observe that before this fix, the number could go below `min` or above `max`. Now it correctly clamps the value.

## Backporting

*This change should be backported to:*

* [ ] 3.0
* [ ] 3.1
* [ ] 4.0
* [x] develop

## Checklist

* [x] ✅ This solution meets the acceptance criteria of the related issue.
* [x] 📝 The related issue checklist is completed.
* [ ] 🧪 This PR does not include **unit tests** for the developed code since functionality requires react state updates. 
* [ ] 🔄 **End-to-End tests** have been added or updated.

  > Not required; behavior verified with existing manual GUI tests.
* [ ] 📚 The **documentation** has not been updated, however a dedicated issue has been created.

  > Updated code comments and type hints.
* [ ] 📌 The **release notes** have noy been updated.

### DIFF
--- a/frontend/taipy-gui/src/components/Taipy/Input.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Input.tsx
@@ -55,6 +55,7 @@ const verticalDivStyle: CSSProperties = {
 const noPaddingYSx = { py: 0 };
 
 const Input = (props: TaipyInputProps) => {
+  
     const {
         type,
         id,
@@ -70,6 +71,7 @@ const Input = (props: TaipyInputProps) => {
     } = props;
 
     const [value, setValue] = useState(defaultValue);
+    
     const dispatch = useDispatch();
     const delayCall = useRef(-1);
     const [actionKeys] = useState(() => getActionKeys(props.actionKeys));
@@ -354,6 +356,7 @@ const Input = (props: TaipyInputProps) => {
     );
 
     useEffect(() => {
+     
         if (props.value !== undefined) {
             setValue(props.value);
         }
@@ -366,7 +369,15 @@ const Input = (props: TaipyInputProps) => {
                     sx={textSx}
                     margin="dense"
                     hiddenLabel
-                    value={value ?? ""}
+                    value={
+                        type === "number"
+                            ? (min !== undefined && Number(value) < min
+                                  ? min
+                                  : max !== undefined && Number(value) > max
+                                  ? max
+                                  : value)
+                            : value ?? ""
+                    }
                     className={`${className} ${getComponentClassName(props.children)}`}
                     type={showPassword && type == "password" ? "text" : type}
                     id={id}
--- a/frontend/taipy-gui/src/components/Taipy/Input.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Input.tsx
@@ -168,7 +168,7 @@ const Input = (props: TaipyInputProps) => {
                     : multiline
                     ? evt.currentTarget.value
                     : evt.currentTarget.value;
-            // Always dispatch on blur within delay
+            // Always dispatch on blur within the delay
             if (delayCall.current > 0 || changeDelay === -1 || delayCall.current === -1) {
                 if (changeDelay > 0) {
                     clearTimeout(delayCall.current);
--- a/frontend/taipy-gui/src/components/Taipy/Input.tsx
+++ b/frontend/taipy-gui/src/components/Taipy/Input.tsx
@@ -19,7 +19,6 @@ import Visibility from "@mui/icons-material/Visibility";
 import VisibilityOff from "@mui/icons-material/VisibilityOff";
 import ArrowDropUpIcon from "@mui/icons-material/ArrowDropUp";
 import ArrowDropDownIcon from "@mui/icons-material/ArrowDropDown";
-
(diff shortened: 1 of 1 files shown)


FabienLelaquais:
Hi @Agney-gt and thank you for your input.

This issue is a bit more tricky than it seems... because input can update the backend synchronously (controlled by the `change_delay` property):

Say you set `min` to 100 and max to 200.
Then you cannot enter '150' any longer because the first '1' will set an 'invalid' value, resetting the input field to '100' (the min value).

Your proposal for `onBlur` is fine to me: we actually send feedback to the backend with a clamped value, that must respect the boundaries.
But changing the render as you do may lead to confusion for the user: the actual value is changed all the time...

Basically: forcing the constraint must be done only when the value is actually validated, and not when interactively entering a new value.
What do you think?

Agney-gt:
Thanks  @FabienLelaquais 

Please have a look at the latest code:
1. Users can freely type in the input field -> Value is validated on blur/step up/ step down and then rendered
2. Value validated between min and max before initial render

FabienLelaquais:
Thanks a lot @Agney-gt for this input.
What I finally did is pick part or you code and injected it into a wider PR (see #2762) whose purpose was to add a new property that enables all values to be forced to integers.

I will now close this issue since it is, at the time of this writing, taken care of in that PR.
Agreed?

Thanks again!

Agney-gt:
I had a look at that #2762

Looks good except for one thing:
https://github.com/Avaiga/taipy/blob/973d09dae847ad84745f997c238a780b98e7dd2c/frontend/taipy-gui/src/components/Taipy/Input.tsx

On line 72, we have  
`actionOnBlur = false,`

If we want the input state value to update correctly if the user edits the value and clicks outside the input box to trigger the blur, then we need to set that value to true since the onBlur flag on line 408 conditionally applies the handleBlur function.

`onBlur={actionOnBlur ? handleBlur : undefined}`

Thanks!

Agney-gt:
It will also require an initialization step otherwise the first render will render -2 instead of min as mentioned [here](https://github.com/Avaiga/taipy/issues/2756) 
@FabienLelaquais 
```
    const initializedDefaultValue = () => {
        const initVal = 
                type === "number"
                    ? (min !== undefined && Number(defaultValue) < min
                    ? min
                    : max !== undefined && Number(defaultValue) > max
                    ? max
                    : defaultValue)
                    : multiline
                    ? defaultValue
                    : defaultValue;
        return initVal;
    }
    const initDefaultValue = initializedDefaultValue();
    const [value, setValue] = useState(initDefaultValue);
```

FabienLelaquais:
@Agney-gt 
> I had a look at that #2762
> 
> Looks good except for one thing:
> ....
> If we want the input state value to update correctly if the user edits the value and clicks outside the input box to trigger the blur, then we need to set that value to true since the onBlur flag on line 408 conditionally applies the handleBlur function.

That is very true,  but... that would then be a breaking change. So I purposely didn't integrate that.

and
> "It will also require an initialization step"

That is also very true, but I decided not to do it: that would be the developer's responsibility, just like min and max not applied when sending the value back to the backend, because of this synchronous update...

Agney-gt:
Allright cool, have a good weekend!

FabienLelaquais:
Closing as a kind-of duplicate to #2762.
The PR does not fix everything (as explained there), but some...

